PR #14, opened 2024-05-31
style(chunk): Delete unnecessary white space

Collection50:
### ASIS

```tsx
 for (let index = 0; index < chunkLength; index++) {
    const start = index * size;
    const end = start + size;
    
    result[index] = arr.slice( start, end);
  }
```

### TOBE

```tsx
 for (let index = 0; index < chunkLength; index++) {
    const start = index * size;
    const end = start + size;
    
    result[index] = arr.slice(start, end); // (*)
  }
```


raon0211 (approved):
Thanks, @Collection50!

PR #16, opened 2024-05-31
fix(package.json): Provide correct types for moduleResolution: node, node10, and node16

manudeli:
I left comments to share intentions


raon0211:
This is actually intended behavior. (exports will be overriden when it is packed into a tarball) Closing this pull request for the reason.

manudeli:
> This is actually intended behavior. (exports will be overriden when it is packed into a tarball) Closing this pull request for the reason.
@raon0211 Actually, because of this point I made this Pull Request. I want to check why we need to overriding with publishConfig.

For testing or developing this package correctly, if we should use this package in another workspace in this repo or any other places, In my opinion, we should use built JavaScript files, not TypeScript files, **to use same files with library users**. because TypeScript file can be transpiled differently depending on the user's environment. so I want to ensure that library maintainer and library users use same files, and I believe that is an important point in the process of developing a library.

raon0211:
In my opinion, when this package is published, library users only see the built JavaScript file, not the TypeScript code. Therefore, all library users interact with the JavaScript code, not the TypeScript.

I believe the issue arises only when we use the es-toolkit library within this workspace. However, the TypeScript files will be transpiled consistently, and they cannot be transpiled differently across user environments, since we control all the compile options within this workspace.

Additionally, using TypeScript directly greatly simplifies the development process. For instance, we don't have to build JavaScript every time we change the source code. Personally, I think this is the direction the JavaScript ecosystem is heading. We have the [types as comments proposal](https://tc39.es/proposal-type-annotations/) in TC39 (as a language spec), and some JavaScript runtimes like Deno natively support TypeScript.

What do you think?

manudeli:
> Personally, I think this is the direction the JavaScript ecosystem is heading. We have the [types as comments proposal](https://tc39.es/proposal-type-annotations/) in TC39 (as a language spec), and some JavaScript runtimes like Deno natively support TypeScript.

Thanks, I understood why you want to override exports field with publishConfig. I reverted it in [3281a7a](https://github.com/toss/es-toolkit/pull/16/commits/3281a7a9e0d5ba337d19386637561b997dde8343)

but we should fix these errors https://arethetypeswrong.github.io/?p=es-toolkit%401.0.2.

image

types field is required in my opinion to support TypeScript users.

package.json
```diff
@@ -25,31 +25,38 @@
     "exports": {
       ".": {
         "import": "./esm/index.mjs",
-        "require": "./dist/index.js"
+        "require": "./dist/index.js",
+        "types": "./dist/index.d.ts"
```
manudeli, line 29: So I updated it

.gitignore
```diff
@@ -15,4 +15,6 @@ coverage
 dist
 esm
 .junit
-out
\ No newline at end of file
+out
+*.d.ts
+*.tgz
```
manudeli, line 20:
```suggestion
*.tgz

```

.scripts/postbuild.sh
```diff
@@ -0,0 +1,6 @@
+echo "export * from './dist/array';" > array.d.ts
+echo "export * from './dist/function';" > function.d.ts
+echo "export * from './dist/math';" > math.d.ts
+echo "export * from './dist/object';" > object.d.ts
+echo "export * from './dist/predicate';" > predicate.d.ts
+echo "export * from './dist/promise';" > promise.d.ts
```
manudeli, line 6:
```suggestion
echo "export * from './dist/promise';" > promise.d.ts

```

raon0211:
Thanks for your recommendation for arethetypeswrong.io! I nearly fixed our type imports as the following image. 

image

However it is notoriously difficult to fix the "masquerading as CJS" part; I guess this should be fixed with either we integrate [tshy](https://github.com/isaacs/tshy) (which is not PnP ready) or we go ESM only (after the release of Node.js 22).

manudeli:
> Thanks for your contribution!

Wait a moment!

package.json
```diff
@@ -18,36 +18,46 @@
...
   ],
   "publishConfig": {
     "access": "public",
+    "main": "./dist/index.js",
+    "types": "./dist/index.d.ts",
     "exports": {
       ".": {
+        "types": "./dist/index.d.ts",
         "import": "./esm/index.mjs",
         "require": "./dist/index.js"
```
manudeli, line 32: > However it is notoriously difficult to fix the "masquerading as CJS" part; I guess this should be fixed with either we integrate [tshy](https://github.com/isaacs/tshy) (which is not PnP ready) or we go ESM only (after the release of Node.js 22).

In my opnion, this may fix our errors.

```suggestion
        "import": {
          "types": "./dist/index.d.ts",
          "default": "./esm/index.mjs"
        },
        "require": {
          "types": "./dist/index.d.ts",
          "default": "./dist/index.js"
        }
```

manudeli (commented):
@raon0211 Let's fix attw errors..!

package.json
raon0211: Actually it doesn't, because in order to be a "ESM" declaration, 

- The type's extension should be `.mts`
- And for each type declaration, we have to specify extensions in its type.

```diff
-export * from './array';
+export * from './array/index.mjs';
```

However this is notoriously difficult in current TypeScript environments unless we introduce some good libraries like tshy.

manudeli: then How about using tsup? I think tsup support this quite easily!
https://github.com/suspensive/react/blob/main/packages/react/package.json#L28-L36

manudeli: Next Pull Request for this is also good for me. If you agree tsup way, I'll make new Pull Request for this!

raon0211: Actually tsup does not support our usecase either.

In order to be tree-shaken well, it is necessary for us to use code splitting. However I find that tsup (and its internal implemenation, esbuild) does not support code splitting properly as rollup's `preserveModules`.

raon0211: Maybe we might try just using tsup's declaration generation...

raon0211:
Will merge this pull request, and will handle integration of tsup as a seperate one. Thanks for your contribution!

po4tion:
Hello @raon0211 !
I have a question. Why do we need this file([.scripts/postbuild.sh](https://github.com/toss/es-toolkit/blob/main/.scripts/postbuild.sh))?

PR #17, opened 2024-05-31
chore(.vscode): Turn on format on save as default

manudeli:
I want to reduce these kind of Pull Request(https://github.com/toss/es-toolkit/pull/14) by this change.

After this change, we force feature "format on save" in contributors' vscode

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -4,7 +4,14 @@
     "**/.pnp.*": true
   },
   "eslint.nodePath": ".yarn/sdks",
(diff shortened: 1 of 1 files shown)


raon0211 (approved):
Thanks, this will help, but I think we should check our linter settings if we could (this would affect only people who use vscode)

manudeli:
> Thanks, this will help, but I think we should check our linter settings if we could (this would affect only people who use vscode)

That's right, Let's make it as soon as possible 👍

PR #22, opened 2024-06-04
chore(*): Apply eslint rule

minsoo-web:
I saw some code that was throwing eslint errors, so I applied lint.

I also changed it to **warn instead of error** because I think there will be cases in the future where the `any` type will be used intermittently.

Before
image

After
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,6 +92,7 @@
   "scripts": {
     "prepack": "yarn build",
     "build": "rm -rf dist esm && tsc -p tsconfig.json --declaration --emitDeclarationOnly --declarationDir dist && rollup -c rollup.config.js && ./.scripts/postbuild.sh",
-    "test": "vitest run --coverage --typecheck"
+    "test": "vitest run --coverage --typecheck",
+    "lint": "eslint ./src --ext .ts"
   }
 }
--- a/src/predicate/isNil.ts
+++ b/src/predicate/isNil.ts
@@ -18,5 +18,5 @@
  * const result3 = isNil(value3); // false
  */
 export function isNil(x: unknown): x is null | undefined {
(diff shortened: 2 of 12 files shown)


src/predicate/isNil.ts
```diff
@@ -18,5 +18,5 @@
  * const result3 = isNil(value3); // false
  */
 export function isNil(x: unknown): x is null | undefined {
-  return x == null || x == undefined;
+  return x == null || x === undefined;
```
raon0211, line 21:
```suggestion
  return x === null || x === undefined;
```

src/predicate/isNotNil.ts
```diff
@@ -14,5 +14,5 @@
  * // result will be [1, 3]
  */
 export function isNotNil<T>(x: T | null | undefined): x is T {
-  return x != null && x != undefined;
+  return x != null && x !== undefined;
```
raon0211, line 17:
```suggestion
  return x !== null && x !== undefined;
```

raon0211 (dismissed):
Thanks!

package.json
```diff
@@ -125,7 +125,8 @@
   "sideEffects": false,
   "scripts": {
     "prepack": "yarn build",
-    "build": "tsup && ./.scripts/postbuild.sh",
-    "test": "vitest run --coverage --typecheck"
+    "build": "rm -rf dist esm && tsc -p tsconfig.json --declaration --emitDeclarationOnly --declarationDir dist && rollup -c rollup.config.js && ./.scripts/postbuild.sh",
```
raon0211, line 128:
```suggestion
    "build": "tsup && ./.scripts/postbuild.sh",
```

raon0211 (approved):

PR #24, opened 2024-06-04
style(*): Setup prettier and apply formatting

minsoo-web:
I think there is no `.prettierrc`, so I added it and applied formatting from the existing files.

The '.pretierrc' file was taken from the [slash](https://github.com/toss/slash/blob/main/.prettierrc) project.


raon0211:
Thanks!

PR #27, opened 2024-06-07
perf(difference): Optimize `difference`

po4tion:
The es-toolkit's `difference` function becomes slower compared to lodash's `difference` as the array size increases.
Therefore, I have modified the `difference` function in es-toolkit so that it performs faster even as the array size increases.

I have reduced the time complexity from O(n*m) to O(n+m).

### AS-IS benchmark(es-toolkit vs lodash)
firstArr's size : 10
secondArr's size : 5
image

firstArr's size : 500
secondArr's size : 250
image

### TO-BE benchmark(es-toolkit vs lodash)
firstArr's size : 10
secondArr's size : 5
image

firstArr's size : 500
secondArr's size : 250
image

### Benchmark(old difference vs new difference)
firstArr's size : 10
secondArr's size : 5
image

firstArr's size : 500
secondArr's size : 250
image

firstArr's size : 10,000
secondArr's size : 5,000
image

### DIFF
--- a/src/array/difference.ts
+++ b/src/array/difference.ts
@@ -15,13 +15,13 @@
  * in the second array.
  *
  * @example
- *  * const array1 = [1, 2, 3, 4, 5];
+ * const array1 = [1, 2, 3, 4, 5];
  * const array2 = [2, 4];
  * const result = difference(array1, array2);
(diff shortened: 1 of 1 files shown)


raon0211 (approved):

PR #28, opened 2024-06-07
fix(differenceBy): Make differenceBy fast

po4tion:
The es-toolkit's `differenceBy` function becomes slower compared to lodash's `differenceBy` as the array size increases.
Therefore, I have modified the `differenceBy` function in es-toolkit so that it performs faster even as the array size increases.

I have reduced the time complexity from O(n*m) to O(n+m).

### AS-IS benchmark(es-toolkit vs lodash)
firstArr's size : 3
secondArr's size : 1
mapper: Math.floor
image

firstArr's size : 500
secondArr's size : 250
mapper: Math.floor
image

### TO-BE benchmark(es-toolkit vs lodash)
firstArr's size : 3
secondArr's size : 1
mapper: Math.floor
image

firstArr's size : 500
secondArr's size : 250
mapper: Math.floor
image

### Benchmark(old differenceBy vs new differenceBy)
firstArr's size : 3
secondArr's size : 1
mapper: Math.floor
image

firstArr's size : 500
secondArr's size : 250
mapper: Math.floor
image

firstArr's size : 10,000
secondArr's size : 5,000
mapper: Math.floor
image

### DIFF
--- a/src/array/differenceBy.ts
+++ b/src/array/differenceBy.ts
@@ -24,7 +24,7 @@
  * // result will be [{ id: 1 }, { id: 3 }] since the elements with id 2 are in both arrays and are excluded from the result.
  */
 export function differenceBy<T, U>(firstArr: T[], secondArr: T[], mapper: (value: T) => U): T[] {
-  const mappedSecondSet = new Set(secondArr.map(mapper));
+  const mappedSecondSet = new Set(secondArr.map(item => mapper(item)));
 
   return firstArr.filter(item => {
     return !mappedSecondSet.has(mapper(item));
--- a/src/array/differenceBy.ts
+++ b/src/array/differenceBy.ts
@@ -17,16 +17,16 @@
  * mapped identity in the second array.
  *
  * @example
- *  * const array1 = [{ id: 1 }, { id: 2 }, { id: 3 }];
+ * const array1 = [{ id: 1 }, { id: 2 }, { id: 3 }];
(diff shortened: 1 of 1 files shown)


src/array/differenceBy.ts
```diff
@@ -17,16 +17,16 @@
...
- *  * const array1 = [{ id: 1 }, { id: 2 }, { id: 3 }];
+ * const array1 = [{ id: 1 }, { id: 2 }, { id: 3 }];
  * const array2 = [{ id: 2 }, { id: 4 }];
  * const mapper = item => item.id;
  * const result = differenceBy(array1, array2, mapper);
  * // result will be [{ id: 1 }, { id: 3 }] since the elements with id 2 are in both arrays and are excluded from the result.
  */
 export function differenceBy<T, U>(firstArr: T[], secondArr: T[], mapper: (value: T) => U): T[] {
-  const mappedSecondArr = secondArr.map(item => mapper(item));
+  const mappedSecondSet = new Set(secondArr.map(mapper));
```
raon0211, line 27:
```suggestion
  const mappedSecondSet = new Set(secondArr.map(item => mapper(item)));
```

raon0211: Changed the implementation so that `mapper` doesn't get other arguments from `Array#map`, e.g. `index`.

raon0211 (approved):

PR #29, opened 2024-06-07
fix(shuffle): Ensure shuffle function does not modify original array

kangju2000:
### [AS-IS]
The previous implementation of the shuffle function modified the original array, which is unintended behavior.

### [TO-BE]
the function now duplicates the array internally before shuffling. This ensures the original array remains unaffected.

## Changes
- Modify the shuffle function to copy the array before shuffling
- Add a test case to validate the preservation of the original array

### DIFF
--- a/src/array/shuffle.ts
+++ b/src/array/shuffle.ts
@@ -12,14 +12,15 @@
  * // shuffledArray will be a new array with elements of array in random order, e.g., [3, 1, 4, 5, 2]
  */
 export function shuffle<T>(arr: T[]): T[] {
+  const copiedArr = arr.slice();
+
   /**
    * https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle#The_modern_algorithm
    */
-  for (let i = arr.length - 1; i >= 1; i--) {
+  for (let i = copiedArr.length - 1; i >= 1; i--) {
     const j = Math.floor(Math.random() * (i + 1));
-
-    [arr[i], arr[j]] = [arr[j], arr[i]];
+    [copiedArr[i], copiedArr[j]] = [copiedArr[j], copiedArr[i]];
   }
 
-  return arr;
+  return copiedArr;
 }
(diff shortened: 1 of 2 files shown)


src/array/shuffle.ts
```diff
@@ -12,14 +12,15 @@
  * // shuffledArray will be a new array with elements of array in random order, e.g., [3, 1, 4, 5, 2]
  */
 export function shuffle<T>(arr: T[]): T[] {
+  const copiedArr = arr.slice();
```
raon0211, line 15:
```suggestion
  const result = arr.slice();
```

src/array/shuffle.ts
```diff
@@ -12,14 +12,15 @@
...
+
   /**
    * https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle#The_modern_algorithm
    */
-  for (let i = arr.length - 1; i >= 1; i--) {
+  for (let i = copiedArr.length - 1; i >= 1; i--) {
```
raon0211, line 20:
```suggestion
  for (let i = result.length - 1; i >= 1; i--) {
```

src/array/shuffle.ts
```diff
@@ -12,14 +12,15 @@
...
     const j = Math.floor(Math.random() * (i + 1));
-
-    [arr[i], arr[j]] = [arr[j], arr[i]];
+    [copiedArr[i], copiedArr[j]] = [copiedArr[j], copiedArr[i]];
```
raon0211, line 22:
```suggestion
    [result[i], result[j]] = [result[j], result[i]];
```

src/array/shuffle.ts
```diff
@@ -12,14 +12,15 @@
...
   }
 
-  return arr;
+  return copiedArr;
```
raon0211, line 25:
```suggestion
  return result;
```

raon0211 (approved):
This was unintended behavior. Thanks for pointing this out!

PR #31, opened 2024-06-10
feat: Support passing arguments to callback function in once function

guesung:
Hi :)
Previously, the `once` function could only handle callback functions without arguments.
With this update, the `once` function now properly handles callback function that require arguments.

### Changes
- Updated once function's type definition to accommodate callbacks that can recieve arguments
  ```ts
  // once.ts
  export function once<F extends (...args: any[]) => any>(func: F): F {
     ...
  }
  ```

### DIFF
--- a/docs/ko/reference/function/once.md
+++ b/docs/ko/reference/function/once.md
@@ -5,11 +5,11 @@
 
 ## 인터페이스
 
-function once<F extends () => any>(func: F): F;
+function once<F extends (...args: any[]) => any>(func: F): F;
 
 ### 파라미터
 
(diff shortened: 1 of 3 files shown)


raon0211:
Hi @guesung, thanks for your contribution!

While this is a possible usecase, I believe the behavior of `once` is not well-defined when we allow the function to get arguments. For example, consider this example:

```tsx
function identity(num: number): number {
  return num;
}

const func = once(identity);

// Returns 1
func(1);
// Returns 1
func(2);
```

I believe this is not what we intended, and it leads to more errors in code. Therefore, I wish we only support functions which take no arguments in `once`. What do you think?

raon0211:
Closing the pull request for inactivity, feel free to open it again :)

PR #32, opened 2024-06-11
fix: add the type readonly array to the partition

ssi02014:
Hello 👋 Here's a suggestion.
Currently, it does not work correctly for arrays with `const assertion` applied. 

I think it would be nice to add a `readonly T[]` type, what do you think?

### AS-IS

스크린샷 2024-06-11 오후 10 34 51

```ts
function partition<unknown>(
  arr: unknown[],
  isInTruthy: (value: unknown) => boolean
): [truthy: unknown[], falsy: unknown[]];
```

스크린샷 2024-06-11 오후 10 35 30

```ts
function partition<unknown>(
  arr: unknown[],
  isInTruthy: (value: unknown) => boolean
): [truthy: unknown[], falsy: unknown[]];
```

### TO-BE
```ts
const arr = [1, 2, 3, 4, 5] as const;

const [evnes, odds] = partition(arr, (item) => item % 2 === 0);
// const evens: (1 | 2 | 3 | 4 | 5)[]
// const odds: (1 | 2 | 3 | 4 | 5)[]
```

```ts
const arr = [
  { name: "mj", age: 27 },
  { name: "gromit", age: 30 },
] as const;

const [ob, yb] = partition(arr, (item) => item.age >= 30);
/*
  const ob: ({
    readonly name: "mj";
    readonly age: 27;
  } | {
    readonly name: "gromit";
    readonly age: 30;
  })[]

 const yb ({
    readonly name: "mj";
    readonly age: 27;
  } | {
    readonly name: "gromit";
    readonly age: 30;
  })[]
*/
```


ssi02014:
I think this `readonly T[]` type should be added to other array utility functions as well.
I can work on it incrementally if it makes sense and is needed. 🤗

raon0211:
Hi @ssi02014, thanks for your contribution! As you mentioned, our library does not currently support `readonly` arrays. Menawhile, I think using `readonly` types in TypeScript is a major use case, and it affects our main interface. Because of our [design principles](https://github.com/toss/es-toolkit/blob/main/.github/CONTRIBUTING.md#12-simplicity) which value simplicity, I wonder this would be beneficial for our users.

ssi02014:
@raon0211 Thank you for your feedback 🙏

In my case, I wanted to narrow down the types as much as possible in my development, so I was disappointed to see a type error when using a readonly array.

>Because of our [design principles](https://github.com/toss/es-toolkit/blob/main/.github/CONTRIBUTING.md#12-simplicity) which value simplicity, I think this would not meet our design principle.

I don't think adding a `readonly` type breaks the principle of simplicity, but this may be a matter of perspective.

I think it's worth considering whether it's more appropriate to keep the existing principle and preclude users from using readonly, or to add the readonly type but change the function interface.

raon0211:
How about this? Since `T[]` is assignable to `readonly T[]`, we can change all our array utilities to accept only `readonly T[]`.

For example, the signature of `partition` would become:

```typescript
function partition<T>(arr: readonly T[], isInTruthy: (value: T) => boolean): [truthy: T[], falsy: T[]];

However, I'd like our documentation to state that `partition` accepts `T[]` as its argument. This is because many of our users might get confused if they see that `partition` only accepts readonly arrays.

ssi02014:
@raon0211 
Yes. I like your suggestion as well, as it fulfills my initial purpose of "readonly arrays are compatible".
I'll follow your lead on providing documentation.🤗

raon0211:
I think for the sake of simplicity, the docs and implementation could go a little bit out of sync.

ssi02014:
@raon0211 I've reverted the document, annotations I was working on, thank you!

raon0211 (approved):
Thanks!

PR #35, opened 2024-06-11
feat(pick, omit): Restrict the type of arguments in object utils

### LINKED ISSUE #25: Improving Type Definitions in Object Utils
## Summary

I believe that when handling functions in Object Utils, it is necessary to define the allowed types more clearly. Currently, the code accepts all types, which negatively impacts the developer experience (DX), particularly with Array.

If Object Utils aims to handle cases related to `object`, I believe it should also properly handle Array. Otherwise, for utilities dealing with object forms, like `Record<string, any>`, it would be better to apply Type Narrowing.

## 1️⃣ Accepts All Types

```ts
export function omit<T, K extends keyof T>(obj: T, keys: K[]): Omit<T, K>
```

The generic type `T` allows all types, leading to various cases.

```ts
expect(omit(1, ["toFixed"])).toEqual({}); // is pass
```

As shown in the code above, cases that are not used in real environments are allowed. If this utility is meant to handle objects, I think the type should be restricted to objects only.

```diff
- export function omit<T, K extends keyof T>(obj: T, keys: K[]): Omit<T, K>
+ export function omit<T extends Record<string, any>, K extends keyof T>(obj: T, keys: K[]): Omit<T, K>
```

## 2️⃣ Array is Also Allowed

When looking into Object Utils, it is evident that array also work. This is because in JavaScript, arrays are a primitive type of `object`. I'm curious if Object Utils is designed to work sufficiently with arrays.

Below is an example related to the `pick` utility. This issue is likely present in other Object Utils as well.

```ts
// pick
describe("pick", () => {
  // ... other tests
  it("should pick elements from an array", () => {
    const array = [1, 2, 3];

    expect(pick(array, [1, 2])).toEqual({ "1": 2, "2": 3 }); // is pass
  });
});
```

If this utility is to allow Array(object), I believe the type inference or code needs improvement. For instance, it degrades the developer experience (DX) in the `pickBy` utility example below.

```ts
// pickBy
it("Array Test", () => {
  const array = [1, 2, 3];
  const shouldPick = (value: number, key: number) => value === 2;
  const result = pickBy(array, shouldPick); // shouldPick type error
  expect(result).toEqual({ "1": 2 });
});
```

The `value` in `shouldPick` causes a type error. The `T[keyof T]` for arrays is extremely complex.

image

Specifying all types for the value makes it too numerous and less readable. Therefore, `shouldPick` has to be inline (callback function) because TypeScript automatically infers the type.

juno7803:
It fixes #25 ("Accepts All Types" section).

In the original `pick` and `omit` functions, non-object values could be passed as the first argument. I have added code to restrict this to the `object` type. (`Record<string, any>`)


raon0211 (approved):
This seems necessary. Thanks for pointing this out!

PR #39, opened 2024-06-13
docs: Fix broken link from README.md and README-ko_kr.md

ho991217:
Current README files include links like 
```markdown
English | [한국어](./README-ko_kr.md)
```
And these are not accessible outside github (such as [npm registry](https://www.npmjs.com/package/es-toolkit)). It is because those links are declared as a relative URLs.
So I changed them into absolute URLs:
```markdown
English | [한국어](https://github.com/toss/es-toolkit/blob/main/README-ko_kr.md)
```


raon0211 (approved):
Thanks!

PR #40, opened 2024-06-13
perf(uniq): Improve performance of `uniq` function

ho991217:
### What have been changed
Previous code used `for ... of ...` statement to check whether an item included in result array. I've changed it to return unique values with `Set`.

### Improvement
1. Time Complexity
   - `for ... of ...` is `O(N)`. `Array.prototype.includes` does linear search, so it is `O(N)` considering the worst case scenario. So the code before had time complexity of `O(N^2)`.
   - `new Set()`is `O(N)`, `Array.from()` is also `O(N)`, so total time complexity becomes `O(N)`.
2. Actual Calculation Time
   - I've tested both functions in this condition:
      - total item count: `100,000`
      - duplicate item count: `25,000`
      - unique item count: `75,000`
   - and the results were:
      - Execution time of `existing code`: 542.303459 ms
      - Execution time of `updated code`: 3.931583 ms
      - screanshot 2024-06-13 16 12 20

### Possible drawbacks
1. Compatibility
   - `Array.prototype.includes` have out with ES7, on the other hand `Set` have done with ES6. So updated method also supports older version without polyfill.
2. Keeping original order
   - `Set` also keeps its insertion order.
3. Memory usage
   - `uniq` function creates `Set` object every time when function is called. There is chance that slight increase in memory usage. But I guess it does not make huge difference from making `result` array every time.

### Added test cases

I've added test cases related to `uniq` function updates which assures its original functions.

Considering performance is primary factor of `es-toolkit` lib, this pr should be fair enough.
Let me know if other drawback does this pr have.


src/array/uniq.ts
```diff
@@ -13,15 +13,5 @@
...
-      continue;
-    }
-
-    result.push(item);
-  }
-
-  return result;
-}
+  return Array.from(new Set(arr));
+}
```
raon0211, line 17:
```suggestion
}

```

raon0211 (approved):
Thanks! It was originally implemented by using `Set`, but if the array is small, using arrays were more performant. 

However I guess this will not make a big difference with small arrays, and will make a huge difference with big arrays. Thanks for your contribution again!

PR #42, opened 2024-06-13
test(omit) : Enhance omit function tests

sossost:
**Description:**
Add additional test cases for the omit function to ensure comprehensive coverage

**Changes:**
1. Add test case to check if all keys are omitted:
- Verifies that the function returns an empty object when all keys are omitted.

```
it('should return an empty object if all keys are omitted', () => {
  const obj = { a: 1, b: 2, c: 3 };
  const result = omit(obj, ['a', 'b', 'c']);
  expect(result).toEqual({});
});
```

2. Add test case to check if no keys are omitted:
- Verifies that the function returns the original object when no keys are omitted.
```
it('should return the same object if no keys are omitted', () => {
  const obj = { a: 1, b: 2, c: 3 };
  const result = omit(obj, []);
  expect(result).toEqual({ a: 1, b: 2, c: 3 });
});
```

3. Add test case to check that the original object is not mutated:
- Ensures that the function does not modify the original object and returns a new object instead.
```
it('should not affect the original object', () => {
  const obj = { a: 1, b: 2, c: 3 };
  const result = omit(obj, ['b']);
  expect(result).toEqual({ a: 1, c: 3 });
  expect(obj).toEqual({ a: 1, b: 2, c: 3 });
});
```

### DIFF
--- a/src/object/omit.spec.ts
+++ b/src/object/omit.spec.ts
@@ -7,4 +7,23 @@ describe('omit', () => {
 
     expect(omit(object, ['foo', 'bar'])).toEqual({ baz: 3 });
   });
+
+  it('should return an empty object if all keys are omitted', () => {
+    const obj = { a: 1, b: 2, c: 3 };
+    const result = omit(obj, ['a', 'b', 'c']);
+    expect(result).toEqual({});
+  });
+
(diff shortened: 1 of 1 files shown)


raon0211 (approved):
Thanks!

PR #44, opened 2024-06-13
perf(intersectionBy): optimize time complexity

WooWan:
This pull request optimizes the time complexity of the intersectionBy function from O(n*m) to O(n + m). The previous implementation used an array and the includes method, leading to quadratic time complexity due to repeated linear searches. The optimized implementation leverages a Set for constant-time lookups, thus improving the overall performance significantly.

* Benchmark1: small size of array
array 1 size: 10
array 2 size: 5
스크린샷 2024-06-13 오후 7 13 53

* Benchmark2: large size of array
array 1 size: 10000
array 2 size: 5000
image


raon0211 (approved):
Thanks!

Issue #37, opened 2024-06-13
Support `AbortSignal` to `debounce` & `delay` for improved cancellation


jchoi-prestolabs:
The [`debounce`](https://es-toolkit.slash.page/reference/function/debounce.html) function already supports cancellation through its return value, but its usability would improve if it also supported cancellation via [`AbortSignal`](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal), which is [implemented in all major web browsers and JavaScript runtimes](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal#browser_compatibility).

It would also be great to add a cancel feature to the [`delay`](https://es-toolkit.slash.page/reference/promise/delay.html) function, and using `AbortSignal` here as well would be a good idea.

jchoi-prestolabs:
Sure, I have no plan to work for this now.

raon0211:
It would help if we could make `delay` throw `new Error('The operation was aborted')` if it is aborted :) (Like in [undici](https://github.com/nodejs/undici/blob/main/lib/core/errors.js#L78) of Node.js)

raon0211:
This issue was resolved by #45 and #52. Thanks for your contribution, @Hanna922 !

PR #45, opened 2024-06-13
feat(debounce): Support AbortSignal to debounce for improved cancellation

Hanna922:
Hi👋🏻 I have reviewed issue #37 and thought it was a good feature, so I have added it.

> The [debounce](https://es-toolkit.slash.page/reference/function/debounce.html) function already supports cancellation through its return value, but its usability would improve if it also supported cancellation via [AbortSignal](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal), which is [implemented in all major web browsers and JavaScript runtimes](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal#browser_compatibility).

If this change is acceptable, I'll add docs and extend this functionality to the `delay` function.

### Before

```ts
function debounce<F extends (...args: any[]) => void>(func: F, debounceMs: number): F & { cancel: () => void }
```

### After

```ts
function debounce<F extends (...args: any[]) => void>(
  func: F,
  debounceMs: number,
  abortSignal?: AbortSignal
): F & { cancel: () => void }
```


raon0211:
Hello @Hanna922, Thanks for your contribution!

src/function/debounce.ts
```diff
@@ -17,25 +18,57 @@
...
+ * }, 1000, signal);
+ *
+ * // Will cancel the debounced function call
+ * controller.abort();
  */
-export function debounce<F extends (...args: any[]) => void>(func: F, debounceMs: number): F & { cancel: () => void } {
+export function debounce<F extends (...args: any[]) => void>(
+  func: F,
+  debounceMs: number,
+  abortSignal?: AbortSignal
```
raon0211, line 35: To receive signals, what about we follow the [fetch API's option](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#aborting_a_fetch)?

```tsx
interface DebounceOptions {
  signal?: AbortSignal;
}

export function debounce<F extends (...args: any[]) => void>(
  func: F,
  debounceMs: number,
  { signal }: DebounceOptions = {}
) { /* ... */ }
```

src/function/debounce.ts
```diff
@@ -17,25 +18,57 @@
...
   debounced.cancel = function () {
-    if (timeoutId != null) {
+    if (timeoutId !== null) {
       clearTimeout(timeoutId);
+      timeoutId = null;
+    }
+
+    if (abortSignal) {
+      abortSignal.removeEventListener('abort', onAbort);
     }
```
raon0211, line 66:
```suggestion
    abortSignal?.removeEventListener('abort', onAbort);
```

What about using optional chaining here?

src/function/debounce.ts
```diff
@@ -17,25 +18,57 @@
...
   };
 
+  if (abortSignal) {
+    abortSignal.addEventListener('abort', onAbort);
+  }
```
raon0211, line 71:
```suggestion
  abortSignal?.addEventListener('abort', onAbort);
```

Here too

Hanna922: It was reflected in '8cf4772'

src/function/debounce.ts
Hanna922: It was reflected in [b6ed097](https://github.com/toss/es-toolkit/pull/45/commits/b6ed097b8c2d62b77b42e736ccb5a57d08f63c0a)

Hanna922:
@raon0211 I have applied the comments and added documentation! Thank you 🙇🏻‍♀️

Would it be good to apply these changes to throttle as well?

raon0211 (approved):
Awesome work!

PR #47, opened 2024-06-13
perf(intersection): optimize `intersection`

WooWan:
Optimized the intersection function by using a Set for improved lookup efficiency.

* Size of array: 10000

Screenshot 2024-06-14 at 12 04 11 AM


raon0211 (dismissed):
Thanks!

src/array/intersection.ts
```diff
@@ -15,8 +15,10 @@
  * const result = intersection(array1, array2);
  * // result will be [3, 4, 5] since these elements are in both arrays.
  */
-export function intersection<T>(firstArr: readonly T[], secondArr: readonly T[]): T[] {
+export function intersection<T>(firstArr: readonly T[], secondArr: T[]): T[] {
```
raon0211, line 18:
```suggestion
export function intersection<T>(firstArr: readonly T[], secondArr: readonly T[]): T[] {
```

raon0211 (approved):

PR #49, opened 2024-06-14
fix(dropWhile): `dropWhile` returns the whole array when every element does not match the predicate

changwoolab:
Thank you for making this lib!
I found an edge case for `dropWhile` function and I fixed it.

### DIFF
--- a/src/array/dropWhile.ts
+++ b/src/array/dropWhile.ts
@@ -17,5 +17,9 @@
  */
 export function dropWhile<T>(arr: readonly T[], canContinueDropping: (item: T) => boolean): T[] {
   const dropEndIndex = arr.findIndex(item => !canContinueDropping(item));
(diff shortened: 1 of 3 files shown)


src/array/dropWhile.ts
```diff
@@ -17,5 +17,5 @@
  */
 export function dropWhile<T>(arr: readonly T[], canContinueDropping: (item: T) => boolean): T[] {
   const dropEndIndex = arr.findIndex(item => !canContinueDropping(item));
-  return arr.slice(dropEndIndex);
+  return dropEndIndex === -1 ? [] : arr.slice(dropEndIndex);
```
raon0211, line 20:
```suggestion
  if (dropEndIndex === -1) {
    return [];
  }
  
  return arr.slice(dropEndIndex);
```

raon0211 (approved):
Thanks for fixing our bug!

PR #52, opened 2024-06-14
feat(delay): Support AbortSignal to delay for improved cancellation

Hanna922:
Hi👋🏻 I have added the functionality to support cancellation via [AbortSignal](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal) to the `delay`, following the `debounce`.

While working on this, I had a question: Would it be better to create and use a common `Error` file, Like in [undici](https://github.com/nodejs/undici/blob/main/lib/core/errors.js#L78)?

Thank you.

### Before

```ts
function delay(ms: number): Promise<void>
```

### After

```ts
function delay(ms: number, { signal }: DelayOptions = {}): Promise<void>
```

### DIFF
--- a/src/function/debounce.ts
+++ b/src/function/debounce.ts
@@ -67,11 +67,9 @@ export function debounce<F extends (...args: any[]) => void>(
       clearTimeout(timeoutId);
       timeoutId = null;
     }
-
-    signal?.removeEventListener('abort', onAbort);
   };
 
-  signal?.addEventListener('abort', onAbort);
+  signal?.addEventListener('abort', onAbort, { once: true });
 
   return debounced;
 }
--- a/docs/ko/reference/promise/delay.md
+++ b/docs/ko/reference/promise/delay.md
@@ -45,7 +45,7 @@ async function foo() {
   try {
     await delay(1000, { signal });
   } catch (error) {
-    console.log(error); // 'AbortError' 로깅
+    console.log(error); // 'The operation was aborted' 로깅
   }
 }
 ```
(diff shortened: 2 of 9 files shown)


src/function/debounce.ts
```diff
@@ -67,11 +67,9 @@ export function debounce<F extends (...args: any[]) => void>(
       clearTimeout(timeoutId);
       timeoutId = null;
     }
-
-    signal?.removeEventListener('abort', onAbort);
   };
 
-  signal?.addEventListener('abort', onAbort);
+  signal?.addEventListener('abort', onAbort, { once: true });
```
raon0211, line 72: 👍

raon0211:
As you mentioned, I think making a separate error file would be helpful, since the error's `name` property has to be `AbortError` when aborted. (This is a spec of fetch, but it would be better if we follow the spec also in `delay`.)

```tsx
// error/AbortError.ts
class AbortError extends Error {
  constructor (message: string = 'The operation was aborted') {
    super(message);
    this.name = 'AbortError';
  }
}

// function/delay.ts
import { AbortError } from '../../error/AbortError';

/* ... */
return reject(new AbortError());
```

Hanna922:
@raon0211 I applied a separate error file. [c4bbfc8](https://github.com/toss/es-toolkit/pull/52/commits/c4bbfc89b490caa0bf1b698bea3aba02550e51b8)
Thank you for your help 🙇🏻‍♀️

raon0211 (approved):

raon0211:
https://github.com/toss/es-toolkit/issues/37

PR #53, opened 2024-06-14
feat(random): Adds a function that generates a random floating-point number between the minimum and maximum.

### LINKED ISSUE #51: Support for `random`
What if you could support a `random` function that returns a random number for a specific range? I think it would be useful in many different situations.
I would like to work on this if it is possible.

sossost:
close #51

I have added a function to generate a random floating point number between the minimum and maximum values input.

Below is a screenshot of the benchmark.
image

If the min and max are the same, it's returning the min value, just like lodash, but I don't know which is better, throwing an exception or not.

I'll also be working on the randomInt function based on feedback from this PR.


raon0211:
Hello, thanks for your contribution! This is a great pull request.

Meanwhile, I think the result of `random(5, 0)` is hard to tell. Why not we throw an error, like `throw new Error("maximum must be greater than minimum.")`?

```typescript
if (minimum > maximum) {
  throw new Error("maximum must be greater than minimum.");
}
```

sossost:
Good opeanion, it's more consistent and clear than the old way. Let's make an edit in that direction.

raon0211 (approved):

sossost:
Thanks! I'll build on that and add a randomInt function as well.

PR #54, opened 2024-06-14
chore(debounce): using type of cancel function 

parkhojeong:
how about introducing cancel function type ?

### Before 
cancel type is inlined
```ts
cancel: () => void 
...
cancel: () => void 
```

### After
cancel type is declared
```ts
type Cancel = () => void;
...

cancel: Cancel
```

### DIFF
--- a/src/function/debounce.ts
+++ b/src/function/debounce.ts
@@ -2,6 +2,8 @@ interface DebounceOptions {
   signal?: AbortSignal;
 }
 
+type Cancel = () => void;
+
 /**
  * Creates a debounced function that delays invoking the provided function until after `debounceMs` milliseconds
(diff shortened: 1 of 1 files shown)


raon0211:
Hello, thanks for your pull request :)

Unfortunately, this might be a matter of style, and it is a slight difference, but I think inlined type is more readable here.

Looking forward to your next contribution!

parkhojeong:
> Hello, thanks for your pull request :)
> 
> Unfortunately, this might be a matter of style, and it is a slight difference, but I think inlined type is more readable here.
> 
> Looking forward to your next contribution!

i agree that it is a matter of style. thanks for your opinion!

PR #60, opened 2024-06-15
fix(pickBy, omitBy): Enforce stricter argument types in `shouldPick` and `shouldOmit`

### LINKED ISSUE #58: Support for typed `Object.entries`
The `omitBy` and `pickBy` function have implemented like this:
```ts

for (const [key, value] of Object.entries(obj)) {
  if (shouldOmit(value, key)) { // or if(!shouldPick(value, key)) {
    continue;
  }

  (result as any)[key] = value;
}
```
The problem is, typescript does not know if the `key` from `Object.entries(obj)` is part of the `obj`'s key. So `result` had to be asserted as `any` even though it was declared at the top, which is less _type-safe_. Also, the `value` is inferred as `any` type, too.

Adding type assertion to `Object.entries(obj)` might be helpful. And changing type of `key` from `string` to `keyof T`. I want to know whether you intentionally set type of `key` to `string` instead of `keyof T` or not.

If you want to utilize typed `Object.entries` more, please consider adding `toPairs` function which implemented in [lodash](https://lodash.com/docs/4.17.15#toPairs).

Let me know what you think!

ho991217:
Resolves #58.

### Changed type of `key` from `string` to `keyof T` in `shouldPick` and `shouldOmit`.
Previously typescript should assume all `key`s as just `string`.
    스크린샷 2024-06-15 17 44 42
Now typescript knows specific types of `key`s
스크린샷 2024-06-15 17 44 17
스크린샷 2024-06-15 17 44 56

And edited existing document as to made changes.
Also, I found that type of `value` had set to `any`, unlike real implementation. So I've fixed that too.

closes #58

### DIFF
--- a/docs/ko/reference/object/omitBy.md
+++ b/docs/ko/reference/object/omitBy.md
@@ -7,13 +7,13 @@
 ## 인터페이스
 
 ```typescript
-function omitBy<T extends Record<string, any>>(obj: T, shouldOmit: (value: any, key: string) => boolean): Partial<T>;
+function omitBy<T extends Record<string, any>>(obj: T, shouldOmit: (value: T[keyof T], key: keyof T) => boolean): Partial<T>;
 ```
 
 ### 파라미터
 
 - `obj` (`T`): 프로퍼티를 생략할 객체예요.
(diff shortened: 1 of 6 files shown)


raon0211 (approved):
Thanks!

Issue #57, opened 2024-06-15
Current bench implementation does not ensure that `es-toolkit` is always faster than `lodash`.


changwoolab:
I think implementation for bench should cover various length of array to ensure that it is always faster than `lodash`

### Example

With current implementation, `intersection` is faster than lodash.

<img width="801" alt="Screenshot 2024-06-15 at 2 49 14 PM" src="https://github.com/toss/es-toolkit/assets/64240134/f5b028fe-86f9-4e7f-adf3-30c466f3f90b">

But after I just changed the length to 10000, `intersection` is slower than lodash
```ts
const array1 = Array.from({ length: 10000 }, () => Math.floor(Math.random() * 1000));
const array2 = Array.from({ length: 10000 }, () => Math.floor(Math.random() * 1000));

describe('intersection', () => {
  bench('es-toolkit', () => {
    intersectionToolkit(array1, array2);
  });

  bench('lodash', () => {
    intersectionLodash(array1, array2);
  });
});
```

<img width="787" alt="Screenshot 2024-06-15 at 2 48 27 PM" src="https://github.com/toss/es-toolkit/assets/64240134/8851aa83-6302-4019-827f-e5c702559225">

raon0211:
Thanks for pointing this out :)

As you mentioned, there were some functions that are less performant than lodash when the array becomes large. We aim to optimize this by reducing time complexity.

When it comes to intersection and intersectionBy, it was actually improved in the following pull requests:

- [intersection](https://github.com/toss/es-toolkit/pull/47)
- [intersectionBy](https://github.com/toss/es-toolkit/pull/44)

Could you pull the latest version and check whether our implementation got faster?

raon0211:
Seems that our implementation is still slow, even after optimizing our implementation -- we should add a test case and ensure that our implementation is faster in most cases.

raon0211:
It's kind of strange in my environment that the result changes every time I re-run the test. 

Maybe the array is randomly created?

<img width="680" alt="image" src="https://github.com/toss/es-toolkit/assets/3102175/843656b0-5a7c-4853-9f65-71749878d8cf">

raon0211:
Our aim is to be faster, or at least as fast as lodash most of the time.

changwoolab:
@raon0211 
## Strange symptoms

It was tested on `v1.3.0` which includes PRs you mentioned.
[...]

PR #62, opened 2024-06-15
perf(intersection): Improved performance for `intersection` function when the size of array is large

changwoolab:
This PR is related to this comment https://github.com/toss/es-toolkit/issues/57#issuecomment-2169311834

It seems that using pure object rather than Map object is more performant when dealing with lots of I/O.
So, I made the `intersection` function to use pure object when it is dealing with the array of size bigger than 120.
For convenience, I made `FastMap` class to abstract the behavior of my job. 
During this job, I referred to `MapCache` in `lodash` (https://github.com/lodash/lodash/blob/main/src/.internal/MapCache.ts)

I tested lengths for all combinations that I could make with the set { 10, 100, 1000, 10000, 100000 }.
My new implementation was more performant than lodash except for `{ 100, 100000 }`, `{10, 100000}`, `{100000, 10}`.

### Before (Length 10000)
Screenshot 2024-06-15 at 10 02 28 PM

### After (Length 10000)
Screenshot 2024-06-15 at 10 02 04 PM

## NOTE
I'm not sure this is the case for all environment. So PLEASE RUN BENCHMARK ON YOUR ENVIRONMENT when you review this PR.

### DIFF
--- a/src/array/intersection.spec.ts
+++ b/src/array/intersection.spec.ts
@@ -1,5 +1,6 @@
 import { describe, expect, it } from 'vitest';
 import { intersection } from './intersection';
+import { randomInt } from 'crypto';
 
 describe('intersection', () => {
   it('should return the intersection of two arrays', () => {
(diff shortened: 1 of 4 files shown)


changwoolab:
Closes this PR because `Map` is more performant at most times.
https://github.com/toss/es-toolkit/issues/57#issuecomment-2170218186

PR #63, opened 2024-06-15
feat: add isEmpty function, check for empty values

dragonsuperf:
Implemented a new `isEmpty` function to check if a value is empty. This function works with various data types like strings, arrays, objects, maps, and sets. 

#### The following values are considered empty:

- null
- undefined
- false
- 0
- NaN
- "" (empty string)
- empty arrays
- empty objects (objects with no enumerable own-properties)
- empty Map and Set
- objects with a valid length property that is 0 or negative

Benchmarks demonstrating the performance of this function are included in the attached images below.
image

### DIFF
--- a/benchmarks/isEmpty.bench.ts
+++ b/benchmarks/isEmpty.bench.ts
@@ -1,0 +1,19 @@
+import { bench, describe } from 'vitest';
+import { isEmpty as isEmptyToolkit } from 'es-toolkit';
+import { isEmpty as isEmptyLodash } from 'lodash';
+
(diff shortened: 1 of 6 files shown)


raon0211:
Hello, thanks for your contribution :)

Unfortunately, we currently do not plan to support `isEmpty` in lodash. We believe it is better to use language primitives like `x.length === 0` or `x == null`, `Number.isNaN(x)`, etc. These methods are more intuitive and likely more performant. With the addition of TypeScript, checking for emptiness has become much easier.

We are closing the pull request for now but look forward to your next contribution!

PR #64, opened 2024-06-15
feat(maxBy): Add maxBy function that select element that have max value by given condition in array

### LINKED ISSUE #55: Support for `maxBy`
Support `max` function in two ways would be helpful.

1. if input param's type is `number[]`: same as lodash's `max`
2. if input param's type is `any[]`: same as lodash's `maxBy`

l2hyunwoo:
close #55 

I added a function that select element that have max value by given condition in array.

스크린샷 2024-06-15 오후 10 55 08
스크린샷 2024-06-15 오후 10 55 21

- If input array is not empty, return first element of array that have maximum value that satisfies given condition
- If input array is empty, return undefined

### DIFF
--- a/src/math/maxBy.ts
+++ b/src/math/maxBy.ts
@@ -16,7 +16,7 @@
  *  const result = maxBy(people, person => person.age);
  *  // result will be { name: 'Nunu', age: 30 }
  */
-export function maxBy<T>(elements: T[], selector: (element: T) => number): T | undefined {
+export function maxBy<T>(elements: T[], selector: (element: T) => number): T {
   if (elements.length === 0) {
     return undefined;
   }
--- a/src/math/maxBy.ts
+++ b/src/math/maxBy.ts
@@ -23,8 +23,7 @@ export function maxBy<T>(elements: T[], selector: (element: T) => number): T | u
   let maxElement = elements[0];
   let max = selector(maxElement);
 
-  for (let i = 1; i < elements.length; i++) {
-    const element = elements[i];
+  for (const element of elements) {
     const value = selector(element);
 
     if (value > max) {
--- a/src/math/maxBy.ts
+++ b/src/math/maxBy.ts
@@ -17,10 +17,6 @@
  *  // result will be { name: 'Nunu', age: 30 }
  */
 export function maxBy<T>(elements: T[], selector: (element: T) => number): T {
-  if (elements.length === 0) {
-    return undefined;
-  }
-
   let maxElement = elements[0];
   let max = -Infinity;
   for (const element of elements) {
--- a/src/math/maxBy.ts
+++ b/src/math/maxBy.ts
@@ -19,6 +19,7 @@
 export function maxBy<T>(elements: T[], selector: (element: T) => number): T {
   let maxElement = elements[0];
   let max = -Infinity;
+
   for (const element of elements) {
     const value = selector(element);
     if (value > max) {
--- a/src/math/maxBy.ts
+++ b/src/math/maxBy.ts
@@ -1,0 +1,34 @@
+/**
+ * Returns the element of the specified array that has the maximum value, according to the specified selector.
+ *
+ * This function takes an array of elements and a selector function, and returns the element with the maximum value according to the selector.
+ *
+ * @param {T[]} elements - An array of elements to be compared.
+ * @param {(element: T) => number} selector - A function that takes an element and returns a number.
+ * @returns {T} The element with the maximum value according to the selector.
+ *
+ * @example
+ * const people = [
+ *  { name: 'Mark', age: 25 },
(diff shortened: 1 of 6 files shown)


src/math/maxBy.ts
```diff
@@ -0,0 +1,36 @@
...
+ * @example
+ * const people = [
+ *  { name: 'Mark', age: 25 },
+ *  { name: 'Nunu', age: 30 },
+ *  { name: 'Overmars', age: 20 }
+ *  ];
+ *  const result = maxBy(people, person => person.age);
+ *  // result will be { name: 'Nunu', age: 30 }
+ */
+export function maxBy<T>(elements: T[], selector: (element: T) => number): T | undefined {
```
raon0211, line 19:
```suggestion
export function maxBy<T>(elements: T[], selector: (element: T) => number): T {
```
I believe returning `undefined` is stricter and less error-prone.

However, following TypeScript's convention that `arr[0]` is not nullable even if `arr` is empty (unless [noUncheckedIndexedAccess](https://www.typescriptlang.org/tsconfig/#noUncheckedIndexedAccess) is on) is also important for consistency -- that's why we chose `sample` to not return nullable even if it can return `undefined`.

For the sake of consistency, I believe we should not return `undefined` here.

What do you think?

src/math/maxBy.ts
```diff
@@ -0,0 +1,36 @@
...
+  for (const element of elements) {
+    const value = selector(element);
+
+    if (value > max) {
+      max = value;
+      maxElement = element;
+    }
+  }
+
+  return maxElement;
```
raon0211, line 35:
```suggestion
  let maxElement = elements[0];
  let max = -Infinity;

  for (const element of elements) {
    const value = selector(element);

    if (value > max) {
      max = value;
      maxElement = element;
    }
  }

  return maxElement;
```

src/math/maxBy.ts
l2hyunwoo: > However, following TypeScript's convention that arr[0] is not nullable even if arr is empty

I tested this on TS Playground. And the result is below.

```typescript
const array: number[] = [];
console.log(array[0]);

// [LOG]: undefined 
```

I think this log says it's null(or undefined) when point first element of empty array.

스크린샷 2024-06-17 오후 2 32 28

--------------------------

I think usage without undefined makes compilcated usecase. Let's see below example code.

## When we use undefined

```ts
  const maxElement = maxBy<number>([]) ?? 0;
```

## When we use Exception

```ts
let maxElement;
try {
  maxElement = maxBy<number>([]);
} catch (error) {
  // TODO: Logic when error occurred
}

const result = maxElement;
```

I think above one is more simple. What do you think?

--------------------------

But for the sake of consistency in this library, I also think it should not return undefined. Thanks for your advice.

src/math/maxBy.ts
l2hyunwoo: It's more readable I think. I'll fix it to this.

src/math/maxBy.ts
raon0211: image

Oh, what I thought was that here `item` is not nullable (in static types) even if `array` is empty.

raon0211: So I thought dealing with empty arrays is what users of `es-toolkit` have to handle... 
(We do not guarantee that the array is not empty using types)

src/math/maxBy.ts
l2hyunwoo: It changed on [this commit](https://github.com/toss/es-toolkit/pull/64/commits/ef4093107619021e63fa4f6eade496b42f4d89c5)

src/math/maxBy.ts
l2hyunwoo: I see... It seems that usual convention in typescript. It's so hard to admit, but now I'm understand what you wrote. I'll change to non-nil type! Thanks for check it.

l2hyunwoo: After modification, all test cases pass well 👍🏻

src/math/maxBy.ts
```diff
@@ -0,0 +1,31 @@
...
+ *  { name: 'Nunu', age: 30 },
+ *  { name: 'Overmars', age: 20 }
+ *  ];
+ *  const result = maxBy(people, person => person.age);
+ *  // result will be { name: 'Nunu', age: 30 }
+ */
+export function maxBy<T>(elements: T[], selector: (element: T) => number): T {
+  let maxElement = elements[0];
+  let max = -Infinity;
+  for (const element of elements) {
```
raon0211, line 22:
```suggestion

  for (const element of elements) {
```

raon0211 (approved):

raon0211:
Thanks, @l2hyunwoo !

PR #67, opened 2024-06-16
docs(isUndefined): Polish document styles and fix incorrect comments

jgjgill:
### Description

I fix documentation and comments related to `isUndefined`.

#### Applying inline code markdown

I applied inline code markdown for `undefined`, `true`, and `false`.
I think this will improve readability.

#### Fix comments

I changed `null[]` to `undefined[]`.

스크린샷 2024-06-16 오후 4 56 47

**AS-IS**

```ts
    // Here the type of result should be `null[]`.
    expect(result).toStrictEqual([undefined]);
```

**TO-BE**

```ts
    // Here the type of result should be `undefined[]`.
    expect(result).toStrictEqual([undefined]);
```

### DIFF
--- a/docs/ko/reference/predicate/isUndefined.md
+++ b/docs/ko/reference/predicate/isUndefined.md
@@ -1,6 +1,6 @@
 # isUndefined
 
-주어진 값이 undefined인지 확인해요.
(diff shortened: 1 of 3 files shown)


raon0211 (approved):
Thanks!

PR #71, opened 2024-06-17
feat(minBy): Add minBy function that select element that have min value by given condition in array

### LINKED ISSUE #65: Support for `minBy`
Support minBy function in two ways would be helpful. (similar as [`maxBy`](https://github.com/toss/es-toolkit/issues/55))

Compares the values of selected properties of the elements in the array when the array is not empty and returns the element with the minimum value. If empty, returns undefined.

l2hyunwoo:
close #65 

I added a function that select element that have min value by given condition in array.

스크린샷 2024-06-17 오후 5 14 46
스크린샷 2024-06-17 오후 5 15 01

- If input array is not empty, return first element of array that have minimum value that satisfies given condition
- If input array is empty, return undefined

### DIFF
--- a/src/math/minBy.ts
+++ b/src/math/minBy.ts
@@ -1,0 +1,27 @@
+/**
+ * Returns the minimum value in an array of numbers.
+ *
+ * This function takes an array of elements and a selector function, and returns the element with the minimum value according to the selector.
+ *
+ * If the array is empty, the function returns `undefined`.
+ *
+ * @param elements An array of elements.
+ * @param selector A function that selects a number from an element.
+ */
+export function minBy<T>(elements: T[], selector: (element: T) => number): T | undefined {
+  if (elements.length === 0) {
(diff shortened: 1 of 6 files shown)


src/math/minBy.ts
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * If the array is empty, the function returns `undefined`.
+ *
+ * @param elements An array of elements.
+ * @param selector A function that selects a number from an element.
+ */
+export function minBy<T>(elements: T[], selector: (element: T) => number): T | undefined {
+  if (elements.length === 0) {
+    return undefined;
+  }
```
raon0211, line 14:
```suggestion
export function minBy<T>(elements: T[], selector: (element: T) => number): T {
```

raon0211 (approved):
Thanks, @l2hyunwoo

PR #76, opened 2024-06-18
style(*): apply formatting

mass2527:
While reviewing the codebase, I noticed that some files weren't formatted with Prettier, so I ran the formatting script.


raon0211 (approved):
Thanks!

Issue #46, opened 2024-06-13
Support for `range`


raon0211:
Support for`range` as in [lodash](https://lodash.com/docs/4.17.15#range) would be helpful.

WooWan:
If you don't plan to work on it, is it okay if I give it a try?

raon0211:
Of course!

PR #77, opened 2024-06-18
feat(range): Add range function

WooWan:
Closes: #46 

This pull request adds a utility range function which behaves range from [lodash](https://lodash.com/docs/4.17.15#range). to generate an array of numbers from a start value to an end value with a specified step. If step is not provided, it defaults to 1 for an ascending range and -1 for a descending range.

It shows a performance that is similar to or slightly faster than lodash's range.

Screenshot 2024-06-18 at 11 00 14 PM


src/array/range.spec.ts
```diff
@@ -0,0 +1,28 @@
...
+    expect(range(0, 20, 5)).toEqual([0, 5, 10, 15]);
+  });
+
+  it('handles custom negative step', () => {
+    expect(range(0, -4, -1)).toEqual([0, -1, -2, -3]);
+  });
+
+  it('handles zero step', () => {
+    expect(range(1, 4, 0)).toEqual([1, 1, 1]);
+  });
```
raon0211, line 23: Actually, I steps like `zero` should not permitted in `range` -- why not we throw an error here?

```tsx
throw new Error(`step must not be zero.`);
```

For example, Python's range function throws an error when `step` is zero.

image

raon0211: I think we should check if `step` is an integer (using `Number.isInteger`) and not zero..

src/array/range.ts
```diff
@@ -0,0 +1,48 @@
...
+
+export function range(start: number, end?: number, step?: number): number[] {
+  if (end === undefined) {
+    end = start;
+    start = 0;
+  }
+
+  if (step === undefined) {
+    step = start <= end ? 1 : -1;
+  }
```
raon0211, line 38: I think we should return an empty array if `end` is smaller than `start`.

For example, python returns an empty array when `end` is smaller than `start`.

image

src/array/range.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  const length = Math.max(Math.ceil((end - start) / (step || 1)), 0);
+  const result = new Array(length);
+
+  for (let i = 0; i < length; i++) {
+    result[i] = start + i * step;
+  }
+
+  return result;
```
raon0211, line 47:
```suggestion
  const result = [];

  for (let i = start; i < end; i += step) {
    result.push(start);
  }
  
  return result;
```

I guess implementation like this might be more readable, what do you think?

src/array/range.spec.ts
WooWan: [lodash](https://lodash.com/docs/4.17.15#range) supports the step=0 case, and I think supporting the step=0 case would allow lodash users to migrate libraries to es-toolkit without breaking changes. what do you think? 
I guess it's a matter of choice, if you think it's better to follow the python implementation, I'll change it.

src/array/range.ts
WooWan: lodash also supports the end is smaller than start case, as I mention previous comment, it would help lodash user to migrate to es-toolkit.

src/array/range.ts
WooWan: I agree it's more readable, but each result.push(i) involves dynamic array resizing, which can be costly. Preallocating a fixed array size demonstrates better performance. I tested this yesterday, and the benchmarks showed it was at least twice as fast, if I remember correctly. 
IMHO, given that we're aiming for a high-performance utility library, it makes sense to prioritize performance considerations a bit more.

WooWan: This is the result of the benchmark I measured again:
Preallocating a fixed array size shows 2~3 times faster than `result.push` in for loop. Let me know what do you think!

Array size is 100:
Screenshot 2024-06-19 at 11 15 27 PM

Array size is 10000:
Screenshot 2024-06-19 at 11 15 45 PM

src/array/range.spec.ts
raon0211: As you mentioned, this will help people transition from lodash. However, I believe that prioritizing a better interface and consistent behavior is more important. 

It seems the behavior when the step is zero isn't well-defined, so I think we should choose not to support that use case.

src/array/range.ts
raon0211: I see why we have this behavior here. However, to keep our implementation simple, I prefer not to support this use case. 

It might make it harder for people using lodash to migrate, but I don't think this is a major use case.

src/array/range.ts
raon0211: If we have this performance difference, I guess we should allocate a fixed-size array in advance.

WooWan:
I was busy last week, Thanks for your thorough and kind review!

PR #78, opened 2024-06-19
fix(clamp): replace null check with typeof for undefined check

BO-LIKE-CHICKEN:
### Description
Replaced the null check with `typeof` check for undefined in the `clamp` function. This change improves type safety and code readability.

### Changes
- Updated clamp function to use `typeof bound2 === 'undefined'` instead of `bound2 == null`.

#### as-is
origin_bench

#### to-be
changed_bench

### References
While considering why it was initially written as `== null`, I came across this [PR](https://github.com/toss/slash/pull/396). However, I believe this case is slightly different, which is why I submitted the PR.

If it was for the same reason, please close this PR.

Thank you!

### DIFF
--- a/src/math/clamp.ts
+++ b/src/math/clamp.ts
@@ -18,7 +18,7 @@
 export function clamp(value: number, maximum: number): number;
 export function clamp(value: number, minimum: number, maximum: number): number;
 export function clamp(value: number, bound1: number, bound2?: number): number {
-  if (bound2 == null) {
+  if (typeof bound2 === 'undefined') {
     return Math.min(value, bound1);
   }
 


raon0211:
I'm not quite sure why `typeof` checks improve type safety and code readability in this context. It seems like `== null` checks are shorter and simpler to use. Can you explain the advantage of `typeof`? 

As far as I know, `typeof` checks are useful if we are not sure if variables are defined, but `bound2` is always defined here.

BO-LIKE-CHICKEN:
First, thank you for your comment. 🙏🏻

I also thought about your feedback throughout the morning. I believe this matter largely falls into the realm of personal preference. However, I prefer the typeof check for the following reasons.

## Clear Expression (main reason)
typeof bound2 === 'undefined' is the most well-known way to check for undefined. While it is indeed useful when variables are not defined, I find it to be the clearest way to check whether an optional value like bound2 has been provided or not.

## Preventing Potential Issues (minor reason)
Although bound2 == null wouldn't behave differently in this function, there is a concern that in a JavaScript environment, other developers using this function might pass null as the third argument, leading to unintended behavior. This might not affect the function's operation, but it still worries me.

I now agree that this does not significantly enhance type safety. However, I still believe that typeof bound2 === 'undefined' more clearly expresses the situation where a value has not been received.

I understand that this can vary based on the library's purpose, existing codebase, or personal preference. So, I'll leave it to your judgment after sharing my opinion. 🙇🏻‍♂️

Thank you! 🤗

PR #79, opened 2024-06-19
fix(round): Throw an exception when `precision` is not an integer

westofsky:
## Description
When rounding, the type of the second parameter, `precision`, is Number, which can also accept float values. Therefore, I added exception handling for cases where it is not an Integer.

## Changes
When attempting to use the round function with inputs like round(1.2345, 3.1), the result was 1.2343860767615331, which does not behave as desired for a rounding function. To fix this, I added exception handling with the following code.
```
if (!Number.isInteger(precision)) {
  throw new Error('Precision must be an integer.');
}
```
This ensures that the `precision` parameter must be an integer, preventing incorrect rounding behavior.


raon0211 (approved):
Thanks!

PR #81, opened 2024-06-20
fix(random): correct JSDoc

mass2527:
I have revised the explanation to correct inaccuracies. I have also included additional details to improve developers' understanding.


raon0211 (approved):
Thanks!

PR #83, opened 2024-06-26
fix(difference): Add missing readonly to parameter

mass2527:
I guess that when supporting read-only arrays, [the `readonly` keyword was mistakenly omitted](https://github.com/toss/es-toolkit/commit/e595e5e017e1f2cb138b1ad3d708635efc5e289e#diff-c6fd3af3ba9f40b9d78d4fbcdfaced66ca7ab40cdf9568ce602b006f045bb9ffL23-R23) from the `difference` parameter.


raon0211 (approved):
Thanks!

PR #93, opened 2024-06-29
feat(keyBy): Add keyBy

### LINKED ISSUE #84: Support for `keyBy`
Support for [keyBy](https://lodash.com/docs/4.17.15#keyBy) as in lodash is needed.

mass2527:
This PR resolves #84.

## Question and Consideration

For consistency, I referenced the `groupBy` function since it bears similarity to `keyBy`. This is why I used almost the same signature for `keyBy` as for `groupBy`, except for the return type:

```typescript
function groupBy<T, K extends string>(arr: readonly T[], getKeyFromItem: (item: T) => K): Record<K, T[]>;
function keyBy<T, K extends string>(arr: readonly T[], getKeyFromItem: (item: T) => K): Record<K, T>;
```

While referencing, I became curious about why type `K` is restricted to `string` and not `PropertyKey`. This restriction requires developers to coerce keys of types such as `number` into `string`.
```typescript
const array = [
      { score: 1, name: 'John' },
      { score: 2, name: 'Jane' },
      { score: 1, name: 'Joe' },
    ];

const result = groupBy(array, item => item.score.toString());
const result2 = keyBy(array, item => item.score.toString());
```

I believe it would enhance usability if developers using the `keyBy` function did not have to handle type conversions differently based on the key type.

Is there a specific reason that I haven't considered?

## Benchmark

스크린샷 2024-06-29 오후 11 36 29


docs/ko/reference/array/keyBy.md
```diff
@@ -0,0 +1,38 @@
+# keyBy
+
+주어진 키 생성 함수로 배열의 각 요소를 매핑해요.
+
+이 함수는 파라미터로 배열과 각 요소에서 키를 생성하는 함수를 받아요.
+키는 함수에서 생성된 키이고, 값은 해당 키를 생성한 요소인 객체를 반환해요.
+만약 동일한 키를 생성하는 요소가 여러 개 있다면, 그 중 마지막 요소가 값으로 사용돼요.
+
+## Signature
```
jgjgill, line 9: Hi, The table of contents is not translated. 🧐
`인터페이스`, `파라미터`, `반환 값`, `예시`

mass2527: Thank you for reviewing my pull request!

docs/ko/reference/array/keyBy.md
```diff
@@ -0,0 +1,38 @@
+# keyBy
+
+주어진 키 생성 함수로 배열의 각 요소를 매핑해요.
```
raon0211, line 3:
```suggestion
배열의 요소를 쉽게 찾을 수 있도록, 키-값 쌍을 이루는 객체로 바꿔요.
```

raon0211:
Thanks for your contribution! Since this is working as expected, I will merge this pull request.

Meanwhile, I agree that all keys might not be numbers. I guess we should accept numbers and symbols as keys. Could you open a seperate pull request for that?

mass2527:
> Thanks for your contribution! Since this is working as expected, I will merge this pull request.
> 
> Meanwhile, I agree that all keys might not be numbers. I guess we should accept numbers and symbols as keys. Could you open a seperate pull request for that?

Yes, I am going to make a PR for this.

PR #101, opened 2024-06-30
feat(sampleSize): add sampleSize

### LINKED ISSUE #87: Support for `sampleSize`
Support for [sampleSize](https://lodash.com/docs/4.17.15#sampleSize) is needed.

jgjgill:
close #87 

## Add error check

I added the error checking code from the code discussed in the issue.

```ts
if (size > array.length) {
  throw new Error('Size must be less than or equal to the length of array.');
}
```

If the user enter a size greater than the length of the array (e.g. `sampleSize([1, 2, 3], 4)`) in the absence of error checking code, the following error message is displayed.

> Error: Invalid input: The maximum value must be greater than the minimum value.

스크린샷 2024-06-30 오후 6 31 31

 I think that error log is less relevant to `sampleSize`.
Feel free to tell me if you think it's unnecessary. 🙇‍♂️

### DIFF
--- a/benchmarks/sampleSize.bench.ts
+++ b/benchmarks/sampleSize.bench.ts
@@ -1,0 +1,15 @@
+import { bench, describe } from 'vitest';
+import { sampleSize as sampleSizeToolkit } from 'es-toolkit';
+import { sampleSize as sampleSizeLodash } from 'lodash';
+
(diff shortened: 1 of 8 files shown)


raon0211 (approved):

PR #102, opened 2024-06-30
docs: update installation section

minsoo-web:
I update installation section with vitepress api

- combine npm, yarn, pnpm command with code-group API

preview

image


raon0211 (approved):
Thanks!

Issue #94, opened 2024-06-30
Support for `fill`


raon0211:
We need to support [fill](https://lodash.com/docs/4.17.15#fill).

hansolbangul:
I've given this a try:

```typescript
export function fill<T>(array: any[], value: T): T[];
export function fill<T, P>(array: Array<T | P>, value: P, start?: number, end?: number): Array<T | P>;
export function fill<T, P>(array: Array<T | P>, value: P, start = 0, end?: number): Array<T | P> {
  if (array == null) {
    return [];
  }

  const length = array.length;
  if (!length) {
    return array;
  }

  const startIdx = start < 0 ? Math.max(length + start, 0) : Math.min(start, length);
  const endIdx = end === undefined ? length : end < 0 ? Math.max(length + end, 0) : Math.min(end, length);

  for (let i = startIdx; i < endIdx; i++) {
    array[i] = value;
  }

  return array;
}

```

If the above code looks good, may I proceed with enhancing the documentation and tests, and then submit a pull request?

raon0211:
```tsx
export function fill<T>(array: unknown[], value: T): T[];
export function fill<T, P>(array: Array<T>, value: P, start: number): Array<T | P>;
export function fill<T, P>(array: Array<T>, value: P, start: number, end: number): Array<T | P>;
export function fill<T, P>(array: Array<T>, value: P, start = 0, end = array.length): Array<T | P> {
  start = Math.max(start, 0);
  end = Math.min(end, array.length);

  for (let i = start; i < end; i++) {
    array[i] = value;
  }

  return array;
}
```

What about implementation like this? I guess it is more simpler.

hansolbangul:
> ```tsx
> export function fill<T>(array: unknown[], value: T): T[];
> export function fill<T, P>(array: Array<T>, value: P, start: number): Array<T | P>;
> export function fill<T, P>(array: Array<T>, value: P, start: number, end: number): Array<T | P>;
> export function fill<T, P>(array: Array<T>, value: P, start = 0, end = array.length): Array<T | P> {
>   start = Math.max(start, 0);
>   end = Math.min(end, array.length);
> 
>   for (let i = start; i < end; i++) {
>     array[i] = value;
>   }
> 
>   return array;
> }
> ```
> 
> What about implementation like this? I guess it is more simpler.

Your code is excellent. I've learned something new!

PR #109, opened 2024-07-02
feat(fill): Add fill function

hansolbangul:
close (#94)

### Description
This PR introduces a new fill function that populates an array with a specified value over a given range. 
- Filling the entire array with a single value.
- Filling the array with a specific value from a start index to an end index.

### Examples

```typescript
const array = [1, 2, 3];
const result = fill(array, 'a');
// Result: ['a', 'a', 'a']
```

```typescript
const result = fill(Array(3), 2);
// Result: [2, 2, 2]
```

```typescript
const array = [1, 2, 3, 4, 5];
const result = fill(array, 'b', 1, 4);
// Result: [1, 'b', 'b', 'b', 5]
```

### Bench
image

### DIFF
--- a/src/array/fill.ts
+++ b/src/array/fill.ts
@@ -23,18 +23,14 @@
  * // => [4, '*', '*', 10]
  */
 
-export function fill<T>(array: any[], value: T): T[];
-export function fill<T, P>(array: Array<T | P>, value: P, start?: number, end?: number): Array<T | P>;
-export function fill<T, P>(array: Array<T | P>, value: P, start = 0, end?: number): Array<T | P> {
-  const length = array.length;
-  if (!length) {
-    return array;
-  }
+export function fill<T>(array: unknown[], value: T): T[];
+export function fill<T, P>(array: T[], value: P, start: number): Array<T | P>;
+export function fill<T, P>(array: T[], value: P, start: number, end: number): Array<T | P>;
+export function fill<T, P>(array: T[], value: P, start = 0, end = array.length): Array<T | P> {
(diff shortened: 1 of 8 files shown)


src/array/fill.ts
```diff
@@ -0,0 +1,38 @@
...
+ * // => [2, 2, 2]
+ *
+ * const result = fill([4, 6, 8, 10], '*', 1, 3);
+ * // => [4, '*', '*', 10]
+ */
+
+export function fill<T>(array: unknown[], value: T): T[];
+export function fill<T, P>(array: T[], value: P, start: number): Array<T | P>;
+export function fill<T, P>(array: T[], value: P, start: number, end: number): Array<T | P>;
+export function fill<T, P>(array: T[], value: P, start = 0, end = array.length): Array<T | P> {
```
hautest, line 29: Many array functions use `arr` as the parameter name instead of `array`. What are your thoughts on following this convention?

hansolbangul: It sounds good. I think it would be better for consistency with other code. :D

raon0211:
image

Seems that we have some error here.

raon0211:
I guess we have to change the function's signature to accept `Array<T | P>` :eyes:

hansolbangul:
> I guess we have to change the function's signature to accept `Array<T | P>` 👀

Based on your suggestion, I updated the function's signature to accept Array<T | P>, and it resolved the error. Thank you for the helpful hint!

raon0211 (approved):
Thanks!

PR #112, opened 2024-07-02
style(*): Add eslint-plugin-jsdoc for improved JSDoc validation and update codebase accordingly

### LINKED ISSUE #105: Proposal: Adding ESLint Plugin for JSDoc
I think it would be great if we add the ESLint plugin for JSDoc.

By using [`eslint-plugin-jsdoc`](https://github.com/gajus/eslint-plugin-jsdoc), we can easily find mistakes with minimal effort. For example, if a template is missing, it will notify us about that.

Screenshot 2024-07-01 at 9:44:47 PM

In addition to this, it will notify us if we don't use @param or @returns, and it will also let us know if the parameter names don't match the ones used in the function.

I would like to know your opinion on this.

mass2527:
close #105 

### Description

- Added [`eslint-plugin-jsdoc`](https://github.com/gajus/eslint-plugin-jsdoc):
   - Disabled `jsdoc/tag-lines` rule, which enforces line breaks between tags (considering moving this responsibility to Prettier due to restricted customization options - I will research this).
   - Disabled `jsdoc/no-defaults` rule, allowing the use of default tags with `@param`.

- Resolved most issues flagged by `eslint-plugin-jsdoc`, except for overloading functions.

### Suggestion

While considering improvements to JSDoc for overloaded functions, I propose providing distinct JSDoc entries for each overload. Currently, functions like `range` display comprehensive information in a single entry, potentially overwhelming developers.

스크린샷 2024-07-02 오후 11 06 17

However, if we write separate JSDoc entries for each overload, we can provide only the necessary information for developers. Although maintainers would need to write more documentation for each function, library users will save time understanding its usage.

스크린샷 2024-07-02 오후 11 14 24

스크린샷 2024-07-02 오후 11 14 51

This is just my idea. feel free to share your opinion.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -14,8 +14,8 @@ module.exports = {
     ecmaFeatures: { jsx: true },
   },
 
-  extends: ['eslint:recommended', 'plugin:@typescript-eslint/recommended', 'prettier'],
-  plugins: ['@typescript-eslint'],
+  extends: ['eslint:recommended', 'plugin:@typescript-eslint/recommended', 'prettier', 'plugin:jsdoc/recommended'],
+  plugins: ['@typescript-eslint', 'jsdoc'],
   rules: {
     'no-implicit-coercion': 'error',
     'no-warning-comments': [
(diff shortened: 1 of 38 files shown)


raon0211 (approved):
Looks great to me. As you mentioned, it would be better to have separate jsdoc comments for each overloaded function.

mass2527:
I will create a pull request for this.

Issue #107, opened 2024-07-02
Support for `forEachRight`


haejunejung:
Hello, 
If you need [loadsh's forEachRight](https://lodash.com/docs/4.17.15#forEachRight) function, Can I work it?

If it is okay to proceed, lodash's forEachRight supports both Array and Object, but in [#91], I would like to ask if it is only necessary to implement array in the array category.

raon0211:
Sure! I think we might just implement it for arrays.

haejunejung:
I've implemented a forEachRight function:

```ts
export function forEachRight<T>(arr: T[] = [], callbackfn: (value: T, index: number, arr: T[]) => void): void {
  for (let i = arr.length - 1; i >= 0; i--) {
    const element = arr[i];
    callbackfn(element, i, arr);
  }
}
```

If this implementation looks good, may I proceed with enhancing the documentation and tests, and then submit a pull request?

haejunejung:
Oh, I think the code you suggested is better too. 
Thanks for your help 👍

PR #119, opened 2024-07-03
feat(forEachRight): add forEachRight

haejunejung:
closes: #107 

I have implemented a `forEachRight` function to address the requirements outlined in issue #107.

image


docs/reference/array/forEachRight.md
```diff
@@ -0,0 +1,38 @@
...
+```ts
+import { forEachRight } from 'es-toolkit/forEachRight';
+
+const array = [1, 2, 3];
+const result: number[] = [];
+
+// Use the forEachRight function to iterate through the array and add each element to the result array.
+forEachRight(array,  (value) => {
+    result.push(value);
+})
```
raon0211, line 35:
```suggestion
// Use the forEachRight function to iterate through the array and add each element to the result array.
forEachRight(array,  (value) => {
  result.push(value);
});
```

raon0211 (dismissed):

Issue #116, opened 2024-07-02
Support for `invert`


wondonghwi:
hello!
If you need [lodash's invert](https://lodash.com/docs/4.17.15#invert) function,

Can I work on it?

I've given this a try:
```typescript
type InvertObject = { [key: string]: string };

export function invert<T extends Record<string, any>>(obj: T): InvertObject {
  const result: InvertObject = {};

  for (const key of Object.keys(obj)) {
    const value = obj[key];
    result[value] = key;
  }

  return result;
}
```
The reason for using the any type in this code is to ensure flexibility. The invert function is designed to work with objects that can have values of any type.

If the above code looks good, may I proceed with enhancing the documentation and tests, and then submit a pull request?

raon0211:
The implementation looks good to me, except we might slightly improve the type here.

```tsx
export function invert<K extends string | number | symbol, V extends string | number | symbol>(obj: Record<K, V>): { [key in V]: K } {
  const result = {} as { [key in V]: K };

  for (const key of Object.keys(obj)) {
    const value = obj[key];
    result[value] = key;
  }

  return result;
}
```

wondonghwi:
@raon0211 

Thank you, I will work on it as you advised and write the test code and document and upload the pull-request.

Thanks to you, I learned something new. Thank you!

PR #125, opened 2024-07-03
feat(invert): add invert

wondonghwi:
close: #116 

I added an invert function to the es-toolkit library. The invert function takes an object as input and returns a new object where the keys and values are swapped. This functionality is available in lodash but was not present in es-toolkit.

The invert function is particularly useful when you need to reverse the mapping of an object. For example, given an object with key-value pairs, the invert function will produce a new object where the original values become the keys and the original keys become the values.

스크린샷 2024-07-04 오전 1 45 37

For reference, you can see the lodash invert function [here.](https://lodash.com/docs/4.17.15#invert)

### DIFF
--- a/benchmarks/invert.bench.ts
+++ b/benchmarks/invert.bench.ts
@@ -1,0 +1,18 @@
+import { bench, describe } from 'vitest';
+import { invert as invertByLodash } from 'lodash';
+import { invert as invertByToolkit } from 'es-toolkit';
+
+const object: { [key: string]: string } = {};
+for (let i = 0; i < 10000; i++) {
+  object[`key${i}`] = `value${i}`;
+}
+
+describe('invert function benchmark', () => {
+  bench('Lodash invert', () => {
+    invertByLodash(object);
+  });
+
+  bench('Custom invert', () => {
+    invertByToolkit(object);
+  });
+});
--- a/benchmarks/invert.bench.ts
+++ b/benchmarks/invert.bench.ts
@@ -8,11 +8,11 @@ for (let i = 0; i < 10000; i++) {
 }
 
 describe('invert function benchmark', () => {
-  bench('Lodash invert', () => {
-    invertByLodash(object);
+  bench('es-toolkit/invert', () => {
+    invertByToolkit(object);
   });
 
-  bench('Custom invert', () => {
-    invertByToolkit(object);
+  bench('lodash/invert', () => {
+    invertByLodash(object);
   });
 });
--- a/src/object/invert.ts
+++ b/src/object/invert.ts
@@ -1,0 +1,31 @@
+/**
+ * Inverts the keys and values of an object. The keys of the input object become the values of the output object and vice versa.
+ *
+ * This function takes an object and creates a new object by inverting its keys and values. If the input object has duplicate values,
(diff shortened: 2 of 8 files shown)


benchmarks/invert.bench.ts
```diff
@@ -0,0 +1,18 @@
...
+for (let i = 0; i < 10000; i++) {
+  object[`key${i}`] = `value${i}`;
+}
+
+describe('invert function benchmark', () => {
+  bench('Lodash invert', () => {
+    invertByLodash(object);
+  });
+
+  bench('Custom invert', () => {
```
hautest, line 15: **custom** is good, but what do you think about doing it with `es-toolkit/invert` ?

end-jarry: @hautest 
Thank you for the good comments, I've made the corrections more clearly!

src/object/invert.ts
```diff
@@ -0,0 +1,31 @@
...
+ * @returns {{ [key in V]: K }} - A new object with keys and values inverted
+ *
+ * @example
+ * invert({ a: 1, b: 2, c: 3 }); // { 1: 'a', 2: 'b', 3: 'c' }
+ * invert({ 1: 'a', 2: 'b', 3: 'c' }); // { a: '1', b: '2', c: '3' }
+ * invert({ a: 1, 2: 'b', c: 3, 4: 'd' }); // { 1: 'a', b: '2', 3: 'c', d: '4' }
+ * invert({ a: Symbol('sym1'), b: Symbol('sym2') }); // { [Symbol('sym1')]: 'a', [Symbol('sym2')]: 'b' }
+ */
+
+export function invert<K extends string | number | symbol, V extends string | number | symbol>(
```
jgjgill, line 20: I think you can use `PropertyKey`.
They're the same, but I think it's more readable.

```suggestion
export function invert<K extends PropertyKey, V extends PropertyKey>(
```

end-jarry: @jgjgill 
I've made the modifications. Thank you for the good comments. I think it's more readable.

wondonghwi: @jgjgill 
I've made the modifications. Thank you for the good comments. I think it's more readable

benchmarks/invert.bench.ts
wondonghwi: @hautest 
Thank you for the good comments, I've made the corrections more clearly!

src/object/invert.ts
```diff
@@ -0,0 +1,31 @@
...
+ * invert({ a: 1, 2: 'b', c: 3, 4: 'd' }); // { 1: 'a', b: '2', 3: 'c', d: '4' }
+ * invert({ a: Symbol('sym1'), b: Symbol('sym2') }); // { [Symbol('sym1')]: 'a', [Symbol('sym2')]: 'b' }
+ */
+
+type PropertyKey = string | number | symbol;
+
+export function invert<K extends PropertyKey, V extends PropertyKey>(obj: Record<K, V>): { [key in V]: K } {
+  const result = {} as { [key in V]: K };
+
+  for (const key of Object.keys(obj) as K[]) {
```
raon0211, line 25:
```suggestion
  for (const key in obj) {
```

What about we simplify here using `for ... in` loop?

PR #129, opened 2024-07-06
fix(package.json): build script

po4tion:
When I run the "yarn build" command, I get the following error. I think it's a matter of executive authority. Rather than setting the executive authority separately, I solved it by adding the "sh" command.

image


raon0211:
I guess this is because we don't have shebangs in the first line of postbuild.sh. Why not add it instead, like:

```sh
#!/bin/sh

echo ...
```

po4tion:
@raon0211 I added "shebangs" and it works fine! Thank you!

raon0211 (approved):
Thanks!

Issue #138, opened 2024-07-09
Support for `zipObjectDeep`


moshuying:
Hello, @raon0211

Can I work startCase of https://github.com/toss/es-toolkit/issues/91 ?

raon0211:
Sure!

raon0211:
Note that [zipObjectDeep will be available in the next release of es-toolkit/compat](https://es-toolkit.slash.page/reference/compat/array/zipObjectDeep.html#zipobjectdeep).

PR #150, opened 2024-07-10
feat(zipObjectDeep): add zipObjectDeep

moshuying:
#138

This pull request implements a function zipObjectDeep. This function takes two arrays: one containing arrays of property paths, and the other containing corresponding values.It returns a new object where paths from the first array are used as key paths to set values, with corresponding elements from the second array as values.Paths can be dot-separated strings or arrays of property names.If the `keys` array is longer than the `values` array, the remaining keys will have `undefined` as their values.

Below is a benchmark comparison between our head implementation and Lodash's head.

image

### DIFF
--- a/benchmarks/zipObjectDeep.bench.ts
+++ b/benchmarks/zipObjectDeep.bench.ts
@@ -1,0 +1,13 @@
+import { bench, describe } from 'vitest';
+import { zipObjectDeep as zipObjectDeepToolkit } from 'es-toolkit';
+import { zipObjectDeep as zipObjectDeepLodash } from 'lodash';
+
+describe('zipObjectDeep', () => {
+  bench('es-toolkit/zipObject', () => {
+    zipObjectDeepToolkit(['a.b[0].c', 'a.b[1].d'], [1, 2]);
+  });
+
+  bench('lodash/zipObject', () => {
+    zipObjectDeepLodash(['a.b[0].c', 'a.b[1].d'], [1, 2]);
+  });
+});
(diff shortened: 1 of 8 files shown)


raon0211:

While your implementation looks great, I think it is hard to grasp what the function actually does. This is because there are several responsibilities entangled in this function:

- Create a new result object.
- Iterate through `keys` and `values` arrays.
- Set `result[key] = value` for each pair of keys and values.

If we had lodash's [set](https://lodash.com/docs/4.17.15#set), I guess it would be much simpler to implement:

```javascript
function zipObjectDeep(keys, values) {
  const result = {};

  for (const [key, value] of zip(keys, values)) {
    set(result, key, value);
  }

  return result;
}
```

So I guess we should implement `set` before implementing `zipObjectDeep`. What do you think?

moshuying:
> Hello, thanks for your contribution!
> 
> While your implementation looks great, I think it is hard to grasp what the function actually does. This is because there are several responsibilities entangled in this function:
> 
> * Create a new result object.
> * Iterate through `keys` and `values` arrays.
> * Set `result[key] = value` for each pair of keys and values.
> 
> If we had lodash's [set](https://lodash.com/docs/4.17.15#set), I guess it would be much simpler to implement:
> 
> ```js
> function zipObjectDeep(keys, values) {
>   const result = {};
> 
>   for (const [key, value] of zip(keys, values)) {
>     set(result, key, value);
>   }
> 
>   return result;
> }
> ```
> 
> So I guess we should implement `set` before implementing `zipObjectDeep`. What do you think?

Sure , but where is issue link for `set0`, Can i implementing it ?

raon0211:
I guess we should make a new issue.

raon0211:
Hello, let me merge this and fix the implementation with the new implemented `set` function :)

raon0211 (approved):

moshuying:
tks

> Hello, let me merge this and fix the implementation with the new implemented `set` function :)

PR #155, opened 2024-07-11
feat(clone): add clone

juno7803:
Hello, @raon0211! 😁 I have implemented a clone function with shallow copy. However, I have some questions!

image

### Why `clone` function in es-toolkit targets only plain objects and plain arrays? 🤔
By plain objects, do you mean objects that are pure key-value pairs, excluding Date, Regex, Map, etc.? If so, I'm curious why you requested a clone that targets only plain objects and plain arrays in the issue.

Since this library is intended to target `lodash`, I wondered if it should support other types of data structures as well.

For now, I've added support for other data types, but I can change this based on your response :)

### Implementation of cloneDeep
This is not directly related to clone, but I have a question since it overlaps with the topic of plain objects and plain arrays.

The relatively recent API, [structuredClone](https://developer.mozilla.org/en-US/docs/Web/API/structuredClone), supports deep cloning. It is available in most modern browsers except IE, and there are quite a few libraries written using this API ([remeda](https://remedajs.com/docs/#clone), [atomic-fns](https://atomic-fns.dev/functions/collections.clone.html)).

image

Since es-toolkit is a modern JavaScript library, I think it would be good to use this API. However, if we are only targeting plain objects and plain arrays, I'm wondering if there is a need to implement cloneDeep separately.

On the other hand, structuredClone throws an error when used with unsupported data types, so it might be worthwhile to provide deepClone to handle these cases!

Please read my thoughts and give me your feedback, and I'll adjust the pull request accordingly. Thank you :)

### DIFF
--- a/docs/ko/reference/object/clone.md
+++ b/docs/ko/reference/object/clone.md
@@ -31,6 +31,6 @@ console.log(clonedArr === arr); // false
 
 const obj = { a: 1, b: 'es-toolkit', c: [1, 2, 3] };
 const clonedObj = clone(obj);
-console.log(clonedObj); // { a: 1, b: 'es-toolkit' }
+console.log(clonedObj); // { a: 1, b: 'es-toolkit', c: [1, 2, 3] }
 console.log(clonedObj === obj); // false
 ```
--- a/docs/ko/reference/object/clone.md
+++ b/docs/ko/reference/object/clone.md
@@ -1,0 +1,36 @@
+# clone
+
+주어진 객체의 얇은 복사본을 생성해요.
+
+## Signature
+
+```typescript
+function clone<T>(value: T): T;
+```
+
+### Parameters
+
+- `obj` (`T`): 복사할 객체예요.
+
+### Returns
+
+(`T`): 주어진 객체의 얇은 복사본이에요.
+
+## Examples
+
+```typescript
+const num = 29;
+const clonedNum = clone(num);
+console.log(clonedNum); // 29
+console.log(clonedNum === num); // true
+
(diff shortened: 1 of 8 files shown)


bovas85:
I feel a like-for-like cloneDeep (and clone) would be beneficial as the main use-case for users is to replace lodash functions here

docs/ko/reference/object/clone.md
```diff
@@ -0,0 +1,36 @@
...
+console.log(clonedNum === num); // true
+
+const arr = [1, 2, 3];
+const clonedArr = clone(arr);
+console.log(clonedArr); // [1, 2, 3]
+console.log(clonedArr === arr); // false
+
+const obj = { a: 1, b: 'es-toolkit', c: [1, 2, 3] };
+const clonedObj = clone(obj);
+console.log(clonedObj); // { a: 1, b: 'es-toolkit' }
```
raon0211, line 34:
```suggestion
console.log(clonedObj); // { a: 1, b: 'es-toolkit', c: [1, 2, 3] }
```

raon0211:
I thought supporting plain objects is usually the most straightforward way, but I guess we might support native JavaScript objects, too. I will also add support for `Map`s and `Set`s.

raon0211:
When it comes to `cloneDeep`, I agree that we use `structuredClone` to simplify our codebase!

raon0211 (approved):

juno7803:
@raon0211 I agree with your feedbacks! Based on your feedback, I will also implement cloneDeep. Thank you! 😊

PR #164, opened 2024-07-11
chore(gitignore): add vscode and intellij ide folders to gitignore

bertyhell:
Ignores the folders that are created when using Visual Studio Code or IntelliJ or Webstorm IDE's

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,8 @@
 .DS_Store
 
+.idea/
+.vscode/
+
 .yarn/*
 !.yarn/patches
 !.yarn/plugins
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,6 @@
 .DS_Store
 
 .idea/
-.vscode/
 
 .yarn/*
 !.yarn/patches


.gitignore
```diff
@@ -1,5 +1,8 @@
 .DS_Store
 
+.idea/
+.vscode/
```
manudeli, line 4: Could you explain why we need to remove these directory?

raon0211: I remember that these directories are sometimes auto-generated by the IDE, which can lead to unintended changes in the repository. While we need to block them, there are also specific settings that we include deliberately. I'll review what our `.gitignore` file needs to include shortly.

sean-esper: Yeah if you dont ignore them, they will just keep showing up in random prs when people `git add .`

bertyhell: exactly
Each developer can keep their IDE settings how they like, they don't need to be checked into the repository in my opinion

raon0211, line 4:
```suggestion
```

It appears that IntelliJ makes frequent modifications to the `.idea` directory, whereas the `.vscode` directory sees fewer changes. Furthermore, our `.vscode` directory includes default configurations to accommodate Yarn's PnP (Plug 'n' Play) install strategy. Therefore, I suggest that we only ignore the `.idea` directory.

raon0211 (approved):

raon0211:
Thanks for your pull request!

PR #165, opened 2024-07-11
Translate Simplified Chinese

BlackWhite2000:
I'm not very familiar with .vitepress, so I'm not sure if the .mts configuration is correct. 
I can't guarantee that all the Chinese inside is correct, but at least it reads normally to me.


docs/.vitepress/shared.mts
```diff
@@ -2,6 +2,7 @@ import { createRequire } from 'module';
 import path from 'path';
 import { defineConfig } from 'vitepress';
 import { search as koSearch } from './ko.mts';
+import { search as zh_hansSearch } from './zh_hans.mjs';
```
raon0211, line 5:
```suggestion
import { search as zh_hansSearch } from './zh_hans.mts';
```

raon0211 (approved):

raon0211:
Thanks for your awesome translation!

PR #175, opened 2024-07-13
perf(meanBy): Improve performance of `meanBy` by not using map

changwoolab:
This PR improves the performance of `meanBy` when it computes a large array.

<table>
  <tr>
    <th>before</th>
    <th>after</th>
  </tr>
  <tr>
<td>

before

</td>
<td>

after

</td>
  </tr>
</table>


raon0211:
Hello :) Could we merge all the similar optimizations in this pull request: https://github.com/toss/es-toolkit/pull/179?

PR #176, opened 2024-07-13
fix(snakeCase): Fixed snakeCase to parse the string with intermediate number well

changwoolab:
This PR fixes the string with intermediate number and improves the performance when it parses a long string.

In lodash, `foo2bar` is converted into `foo_2_bar` while es-toolkit doesn't. So I fixed it.
Moreover, I made es-toolkit faster than lodash when it computes a long string.

<table>
  <tr>
    <th>before</th>
    <th>after</th>
  </tr>
  <tr>
<td>

Screenshot 2024-07-13 at 3 22 18 PM

</td>
<td>

Screenshot 2024-07-13 at 3 21 54 PM

</td>
  </tr>
</table>

### DIFF
--- a/benchmarks/snakeCase.bench.ts
+++ b/benchmarks/snakeCase.bench.ts
@@ -3,13 +3,28 @@ import { snakeCase as snakeCaseToolkit } from 'es-toolkit';
 import { snakeCase as snakeCaseLodash } from 'lodash';
 
 describe('snakeCase', () => {
-  bench('es-toolkit/snakeCase', () => {
-    const str = 'camelCase';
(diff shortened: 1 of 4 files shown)


raon0211:
Hello, I added the case like `foo2bar` in this commit. Thanks for your contribution!

https://github.com/toss/es-toolkit/commit/14ec3152589a1907daa3ba5b3d296c1e6bd711af

Meanwhile, could we merge all the performance improvement cases in this pull request (https://github.com/toss/es-toolkit/pull/179) ? I think optimize with identical methods, and I would like to discuss this in one pull request.

PR #179, opened 2024-07-13
perf(lowerCase): improves performance for lowerCase

changwoolab:
This PR improves performance for lowerCase.

<table>
  <tr>
    <th>before</th>
    <th>after</th>
  </tr>
  <tr>
<td>

Screenshot 2024-07-13 at 5 03 46 PM

</td>
<td>

Screenshot 2024-07-13 at 5 04 24 PM

</td>
  </tr>
</table>

### DIFF
--- a/benchmarks/lowerCase.bench.ts
+++ b/benchmarks/lowerCase.bench.ts
@@ -3,13 +3,26 @@ import { lowerCase as lowerCaseToolkit } from 'es-toolkit';
 import { lowerCase as lowerCaseLodash } from 'lodash';
 
 describe('lowerCase', () => {
-  bench('es-toolkit/lowerCase', () => {
-    const str = 'camelCase';
-    lowerCaseToolkit(str);
+  describe('short string', () => {
+    bench('es-toolkit/lowerCase', () => {
+      const str = 'camelCase';
+      lowerCaseToolkit(str);
+    });
+
+    bench('lodash/lowerCase', () => {
+      const str = 'camelCase';
+      lowerCaseLodash(str);
+    });
   });
 
-  bench('lodash/lowerCase', () => {
-    const str = 'camelCase';
-    lowerCaseLodash(str);
+  describe('long string', () => {
+    const LONG_STR = 'camelCase'.repeat(100);
+    bench('es-toolkit/lowerCase', () => {
+      lowerCaseToolkit(LONG_STR);
+    });
+
+    bench('lodash/lowerCase', () => {
+      lowerCaseLodash(LONG_STR);
+    });
   });
 });
--- a/benchmarks/lowerCase.bench.ts
+++ b/benchmarks/lowerCase.bench.ts
@@ -16,7 +16,7 @@ describe('lowerCase', () => {
   });
 
   describe('long string', () => {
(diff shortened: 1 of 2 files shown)


raon0211:
Actually, in my environment (Macbook Pro M1 Max, Node.js v20.13.1), it appears this implementation is slower than the original one, or it does not make a big difference. Could you elaborate on what environment are you testing?

| Before | After |
|---|---|
| image | image |

changwoolab:
@raon0211 
I think sticking with the original implementation is a good idea!
(You don't need to see the details below)

I was testing on (Macbook Pro M1 Max, Node.js v22.2.0) with Rosetta.
It is interesting that as soon as I changed the node version to v22.4.1 with nvm, original one was faster.

But with my Chrome broswer in my Macbook M1, new implementation was faster than the original one.
Screenshot 2024-07-14 at 7 38 33 PM

```ts
const CASE_SPLIT_PATTERN = /[A-Z]{2,}(?=[A-Z][a-z]+[0-9]*|\b)|[A-Z]?[a-z]+[0-9]*|[A-Z]+|[0-9]+/g;
let str = "AsdfAsdfAsdfAsdfAsdfAsdfAsdfAsdf".repeat(100)

console.time('1')
const splitWords = str.match(CASE_SPLIT_PATTERN) || [];
splitWords.map(word => word.toLowerCase()).join(' ');
console.timeEnd('1')

console.time('2')
let result = splitWords[0]?.toLowerCase() ?? '';
for (let i = 1; i < splitWords.length; i++) {
  result += ' ' + splitWords[i].toLowerCase();
}
console.timeEnd('2')
```

I also tested on windows operating system which uses intel chips because I thought that it could be a thing with intel architecture (Rosetta).
It turns out that in my Windows x64 with Intel(R) Core(TM) i3-9100F environment, the code above with new implementation was 2x faster than original one on node.js and 1.2x ~ 1.5x faster on Chrome browser.

But, because it could be the case only for my Windows environment, it needs to be tested more thoroughly on various environments. 
I think this costs too much, so it would be better to stick with the original implementation.

changwoolab:
Oh, it was my mistake. I forgot to apply str.match again 😅
When the length is long, new impl is slower than the original one while it is faster when the length is short or it is with Rosetta.

```ts
const CASE_SPLIT_PATTERN = /[A-Z]{2,}(?=[A-Z][a-z]+[0-9]*|\b)|[A-Z]?[a-z]+[0-9]*|[A-Z]+|[0-9]+/g;
let str = 'AsdfAsdfAsdfAsdfAsdfAsdfAsdfAsdf'.repeat(100);

console.time('1');
const splitWords1 = str.match(CASE_SPLIT_PATTERN) || [];
splitWords1.map((word) => word.toLowerCase()).join(' ');
console.timeEnd('1');

console.time('2');
const splitWords2 = str.match(CASE_SPLIT_PATTERN) || []; // <-- missed
let result = splitWords2[0]?.toLowerCase() ?? '';
for (let i = 1; i < splitWords2.length; i++) {
    result += ' ' + splitWords2[i].toLowerCase();
}
console.timeEnd('2');
```

raon0211:
Thanks for your understanding! Closing the pull request for now.

PR #185, opened 2024-07-13
perf(minBy): optimize minBy

choi2021:
improvement of `minBy` using initialization optimization and converting `for of loop `to `indexed for loop `

**AS-IS benchmark(es-toolkit vs lodash)**

**TO-BE benchmark(es-toolkit vs lodash)**

### DIFF
--- a/src/array/minBy.ts
+++ b/src/array/minBy.ts
@@ -10,15 +10,19 @@
  * minBy([{ a: 1 }, { a: 2 }, { a: 3 }], x => x.a); // Returns: { a: 1 }
  * minBy([], x => x.a); // Returns: undefined
  */
-export function minBy<T>(items: T[], getValue: (element: T) => number): T {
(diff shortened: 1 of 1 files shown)


raon0211:
Hello, thanks for your contribution! While your code improves our code quality by approximately 5%, it seems to introduce several things to our codebase:

- It is more verbose compared to the current implementation.
- It increases our bundle size due to additional `if` statements and access to the `.length` property.

Since the performance difference isn't significant, I think we should stick with the current implementation. What are your thoughts?

choi2021:
Thank you for your feedback. I agree that the performance gains are not substantial enough to offset the increased verbosity and bundle size. Maintaining the current implementation is indeed more beneficial for our codebase's readability and efficiency.

raon0211:
Thanks for your understanding! Always looking forward to your next contribution.

Issue #135, opened 2024-07-08
Support for `curry`


seungrodotlee:
Hello, @raon0211 !
As I mentioned at #133 , I would start work on `curry`!

raon0211:
Sure!

evan-moon:
I was personally thrilled to hear that es-toolkit decided to support the curry function!!

I'm curious about your thoughts on implementing a strict rule for currying, where if a function takes multiple arguments, it transforms them into a composition of single-argument functions. This approach is a bit different from the lodash specification, which I will explain further.

[The curry function in lodash](https://lodash.com/docs/4.17.15#curry) doesn't strictly enforce the rules of currying, and is more like providing partial application rather than true currying.

I think one reason why developers often get confused between currying and partial application is because popular libraries like lodash use these terms incorrectly 😢 

**Therefore, I would like to suggest designing es-toolkit to follow the correct definition of currying strictly!**

---

However, this might make it harder to gain users since it wouldn't be a drop-in replacement for those who are used to lodash.

seungrodotlee:
@evan-moon Thank you for GOOD SUGGESTION! Indeed, the `curry` method provided by existing libraries like `Lodash` has a different form compared to the 'strict type of currying'. However, I believe that many developers, including myself, are familiar with the usage of this 'modified type of currying' provided by these libraries.

In fact, I am now implementing a 'modified type of currying'. However, after reading your suggestions, I think it is important to respect the traditional method to enhance the depth of the library.

**How about this approach?**

1. Implement both curry and partial application.
2. In the documentation, guide developers who want a 'Lodash-style curry' to look for partial application (or another method named differently that implements this type of function).

By implementing it this way, although users might be confused at first, I believe it is a valuable process to correct any misconceptions.

seungrodotlee:
@evan-moon  Based on the suggestions provided and my reply, I have created draft PR #187!

Please check it out and feel free to leave any comments & ideas!

evan-moon:
> @evan-moon Based on the suggestions provided and my reply, I have created draft PR #187!
> 
> Please check it out and feel free to leave any comments & ideas!

Thank you. Let's continue the discussion in that PR! 👍

D-Sketon:
Now curry and curryRight have been fully implemented!

evan-moon:
Thank you @D-Sketon 🙏  Then this issue can be closed now!

PR #187, opened 2024-07-14
feat: add `curry` function

seungrodotlee:
I'm currently working on `curry` function. and draft version is now implemented.

## Benchmark

image

According to the benchmark results, it appears to be **11 to 17x faster** compared to `lodash`.

## Concept

As suggested by @evan-moon in '#135', the `curry` function has been implemented in a way that preserves its purity. However, for developers who find 'Lodash-style curry'(= 'flexible type of curry') familiar and convenient, I have also implemented 'lodash-style curry' under the name `curry.flexible`!

```ts
const fn = (a: number, b: string, c: boolean) => ({ a, b, c });

// each function taking arguments one by one.
curry(fn)(1)('a')(true);

// each funtion taking 1~N arguments.
curry.flexible(fn)(1, 'a')(true);
```

**I would like to hear the opinions of the contributors on this approach.**

Additionally, there are many utility types written during the implementation of curry. It would be great to discuss a structure for storing such internal codes. Currently, I are using the `src/_internal` directory!

image


evan-moon (commented):
Thank you for your contribution! @seungrodotlee 

After reviewing [this document](https://es-toolkit.slash.page/compatibility.html), it seems that the library owner, plans to provide a separate project called `es-toolkit/compat` to allow users to use lodash functions as a drop-in replacement in es-toolkit. (cc. @raon0211)

What do you think about having the official `curry` function in `es-toolkit` follow the original concept, transforming a function with multiple arguments into a composition of single-argument functions, while providing lodash's version of curry (which is essentially partial application) in `es-toolkit/compat`?

Additionally, how about including a note in the jsDoc or reference documentation for the curry function provided in `es-toolkit/compat` that says, "This function offers partial application, not currying. If you want true currying, please migrate to the curry function in `es-toolkit`"?

Also, if we clearly separate the concerns in this way, the `curry` function inside `es-toolkit` will have a much simpler implementation.

seungrodotlee:
@evan-moon Thank you for good opinion! 👍 

In summary, I understand your suggestion as moving the currently implemented `curry.flexible` to `es-toolkit/compat`, leaving only the pure implementation of `curry` in `es-toolkit`. I wonder if my understanding is correct.

Based on the [referenced document](https://es-toolkit.slash.page/compatibility.html), it seems that `es-toolkit/compat` is recommended to be used only during the 'smooth transition' from `lodash` to `es-toolkit`.

Therefore, it seems that the values we actually want to provide should be reflected in `es-toolkit`. Here, we need to decide whether to focus more on purity or convenience. (I am also curious about @raon0211' s opinion on this matter)

As I understand it, the characteristics of `curry` and `partial application` are as follows:

- `curry`: Takes one argument at a time over multiple calls.
- `partial application`: Takes arguments in two stages. After receiving arguments once, the function must be executed the next time it is called.

I thought that there might be a demand for taking multiple arguments over multiple calls. So, I implemented this, i.e., 'lodash-style curry', as `curry.flexible`.

Acutally, I also feel a bit uneasy about calling this implementation `curry`, but personally, I think it would be more convenient to have this implementation under some name!

Is there any other name we could use for this implementation?

evan-moon:
Thanks for your feedback. @seungrodotlee!
I'll put my thoughts together and write them down more clearly.

**I would like to suggest implementing the curry function in both `es-toolkit` and `es-toolkit/compat`.**

Firstly, let’s think about the purposes of `es-toolkit` and `es-toolkit/compat`. I guess the library owner may have created `es-toolkit/compat` because he believes the current implementation in `lodash` is incorrect. Therefore, I think he does not want to include `lodash`'s implementation in `es-toolkit`.

I believe the current implementation of the `curry` function in `lodash` is incorrect.

1. This function provides partial application, not true currying.
2. This is similar to calling a `forEach` function a `map` function just because `map` can perform the same actions as `forEach` if its callback returns nothing. In this sense, `forEach` can be seen as a subset of `map`, and currying can be seen as a subset of partial application. Therefore, the following proposition holds. 

> "all currying is partial application, but not all partial application is currying."

Therefore, it is not correct to name a function that provides the behavior of partial application as curry. I think this logical error is not just a matter of convenience like DX.

However, since many users are familiar with the misconception created by lodash, **I agree with providing a curry function in `es-toolkit/compat` that passes the lodash `curry` function test cases.** (This is only necessary to let users easily switch to es-toolkit using drop-in replacement.)

But I propose that the `curry` function in `es-toolkit` should correctly implement the concept of currying, transforming a function with multiple arguments into a composition of single-argument functions.

(+) Partial application is not just about receiving arguments in two stages. It means fixing some of the arguments of a multi-argument function and returning a function with fewer arguments. Therefore, lodash's curry provides these behaviors!
```ts
const foo = (a, b, c, d) => [a, b, c, d];
const curried = curry(foo);

curried(1, 2)(3, 4); // [1, 2, 3, 4]
curried(1)(2, 3)(4); // [1, 2, 3, 4]
curried(1, 2, 3)(4); // [1, 2, 3, 4]

curried(1)(2)(3)(4); // [1, 2, 3, 4] Only this case is true currying.
```

seungrodotlee:
@evan-moon I greatly appreciate the thorough and insightful explanation you provided!

Now I have a much clearer understanding of the relationship and differences between `curry` and `partial application`.

Given that, the direction you suggested of keeping only the pure implementation of `curry` in `es-toolkit`, and implementing the 'lodash-style `curry`' in `es-toolkit/compat` seems good to me.

Additionally, I think it would also be beneficial to implement `partial application` separately (ex. as a dedicated `partial` method) in es-toolkit, alongside the `curry` implementation.

"On the other hand, I think it would also be better to handle the detailed implementation of these functional programming concepts in a separate library, such as `es-fp`, where we can cover them in a more specialized and thorough manner in the future.

I'm also curious to hear @even-moon and @raon0211's thoughts on this.

```typescript
const fn = (a: number, b: string, c: boolean) => ({ a, b, c });

// each function taking arguments one by one.
curry(fn)(1)('a')(true);

// each funtion taking 1~N arguments.
partial(fn)(1, 'a')(true);
```

raon0211:
Sorry for the late review. This pull request is big, let me review this when I have some time to think.

seungrodotlee:
@raon0211 No worries. Looking forward to a pleasant discussion! 😄

darkyzhou:
Very nice work! Hope the review made a progress

seungrodotlee:
@raon0211 As a `partial` function implemented(Thanks to @D-Sketon), I remove `flexible curry` and refactor types of `curry`. I will work on compat version of curry in another PR as I have time.

kazlabagane (approved):

raon0211 (dismissed):
The implementation looks nice! 

Meanwhile, I think the function's interface might be improved slightly, using function overloading, like:

```typescript
function curry<P, R>(func: (param: P) => R): (param: P) => R;
function curry<P1, P2, R>(func: (param1: P1, param2: P2) => R): (param1: P1) => (param2: P2) => R;
function curry<P1, P2, P3, R>(func: (param1: P1, param2: P2, param3: P3) => R): (param1: P1) => (param2: P2) => (param3: P3) => R;
/* ... */
// We support only functions taking less than 5 arguments
function curry(func: (...args: any[]) => any): (...args: any[]) => any
```

raon0211:
Found one problem for `curry`: `curry` tries to accept an argument even if it is optional. Let us look into the problem..

image

raon0211:
Found out that other libraries like `ramda` or `lodash` does not support optional parameters properly; seems that it is very tricky to handle it in TypeScript...

raon0211 (approved):
Thanks for your work! Sorry for the late merge.

PR #190, opened 2024-07-14
feat(chunk): Add type inference for tuple type to `chunk`

kakasoo:
I'm wondering if it's what you'd expect from this library to help you infer more accurate types. I can often make contributions if these contributions match your expectations.
If you find this PR interesting, please allow me to install an additional library so that I can do tests on the type. Or, set it up to catch compilation errors when operating the test code so that I can do those tests.

image

### DIFF
--- a/src/array/chunk.spec.ts
+++ b/src/array/chunk.spec.ts
@@ -2,6 +2,65 @@ import { describe, expect, it } from 'vitest';
 import { chunk } from './chunk';
 
 describe('chunk', () => {
+  it('if tuple with size 1, it should be divided according to the size.', () => {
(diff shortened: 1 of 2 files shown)


raon0211:
Hello, this actually looks nice. I'm really impressed by the type inference.

Meanwhile, I believe this is not the goal of our `es-toolkit` library for several reasons:

- The core value of `es-toolkit` is performance and implementation simplicity.
- Additionally, adding these type annotations might degrade our developer experience due to the extensive type calculation required.
- Moreover, complex TypeScript types are difficult to write and read, which would reduce the readability and maintainability of our code.

kakasoo:
Unfortunately, I understand what you said.

PR #191, opened 2024-07-14
chore: Refactor the sum function to be more intuitive

endmoseung:
## Why I Refactored This Code

The previous code was based on JavaScript array functions like map and slice. I refactored the code to be more declarative and to **follow coding conventions**. 

I am aware that this code will not work in Node versions earlier than ES6. However, Babel can be used to transpile the code to ensure compatibility.

### DIFF
--- a/src/math/sum.ts
+++ b/src/math/sum.ts
@@ -1,3 +1,4 @@
+const defaultNumber = 0;
 /**
  * Calculates the sum of an array of numbers.
  *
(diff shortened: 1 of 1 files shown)


raon0211:
Hello, thanks for your pull request :)

Meanwhile, we understand that there are a variety of ways to implement the same functionality. In our coding style guideline, we prefer `for` loops over `reduce`. Please refer to our contributing guidelines.

https://github.com/toss/es-toolkit/blob/main/.github/CONTRIBUTING.md#12-simplicity

Please open the pull request again if you have more opinions.

Looking forward to your next contribution.

endmoseung:
Thank you for the quick response. It seems that I missed the section in the documentation you pointed out. 😥
I will make sure to submit a pull request the next time I have the opportunity to contribute!

PR #192, opened 2024-07-14
feat(difference): Add type inference for tupe type to `difference`

kakasoo:
```ts
difference([1, 2, 'a', 'b', true, false] as const, [1, 2, true] as const); // ["a", "b", false]
```

Like this #190,
I would like to support type inference for tuple type.
As I said in the PR before, please let me know if these PRs are not what you expect from this project.


raon0211:
Hello, let's discuss this in this pull request :)

https://github.com/toss/es-toolkit/pull/190#issuecomment-2227272023

Issue #140, opened 2024-07-09
Support for `cloneDeep`


juno7803:
~Can I work [cloneDeep](https://lodash.com/docs/4.17.15#cloneDeep) of https://github.com/toss/es-toolkit/issues/91? 😁~

I'll implement the clone first, and if there are no contributors afterwards, I'll reopen it 🙇

juno7803:
@pcbowers Thank you so much for the great code example and suggestions!
If it's okay with you, how about submitting a pull request yourself to contribute? It would be a huge help to es-toolkit!

If needed, I can provide assistance with documentation, translation work, or test cases. :)

juno7803:
Related with https://github.com/toss/es-toolkit/pull/155, [structedClone](https://developer.mozilla.org/en-US/docs/Web/API/structuredClone) would be better solution for cloneDeep! 
I'll implement it as soon as possible.

bovas85:
@juno7803 structuredClone works fine for most cases, but remember when working with React or other non-supported primitives it will not clone properly and throw

juno7803:
@bovas85 
Yes! That's correct. As you mentioned, [structuredClone](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm#supported_types) throws an error for unsupported types, so I thought it would be good to handle these exceptions and provide support for them. (DOM, function .. extra) related #155 

However, since someone have already created a pull request for #199 to implement cloneDeep first, it seems appropriate to continue the related discussion in that pull request.

raon0211:
This function was recently added to our library. Thanks for opening the issue!

- https://es-toolkit.slash.page/reference/object/cloneDeep.html#clonedeep

PR #199, opened 2024-07-14
feat(cloneDeep): add cloneDeep

de-novo:
## Summary

This PR addresses [Issue #140](https://github.com/toss/es-toolkit/issues/140) by implementing the cloneDeep function to create deep copies of objects, arrays, and other complex data types.

## Benchmark
스크린샷 2024-07-15 오전 8 29 26

## Changes Made
- Functionality: Added cloneDeep to deeply clone objects, arrays, dates, regex, maps, sets, and more.	
- Tests: Added comprehensive test cases to ensure the correctness of the deep cloning functionality.
- Documentation: Updated documentation with examples demonstrating the usage of cloneDeep.
- Benchmarking: Added benchmarks comparing cloneDeep performance with similar functions.

## Considerations
In the current implementation, cloneDeep supports deep cloning of instances, allowing instance methods to be used after cloning. In contrast, the existing library’s clone function does not allow the use of class methods on deeply cloned instances, even though IDEs can infer the methods. This discrepancy can cause significant confusion during development. To address this, the implementation ensures that instances are fully cloned, preserving the functionality of their methods.

For more details, please refer to [PR #155](https://github.com/toss/es-toolkit/pull/155).


samchon:
@de-novo I think it would better to change to `cloneDeep()` function to return `Resolved<T>` type.

Also, it would better to consider more native classes case.

- `Resolved<T>` type: https://github.com/samchon/typia/blob/master/src/Resolved.ts
- Clone function of other library (does not consider `RegExp` type): https://github.com/samchon/typia/blob/master/src/functional/%24clone.ts

bovas85:
note/bug/support issue:
This does not seem to work with Events
image

raon0211:
We discussed that we use `structuredClone` rather than re-implementating deep clones. (See https://github.com/toss/es-toolkit/issues/140)

 Is it more performant to implement `cloneDeep` ourselves?

de-novo:
@samchon Thank you for your valuable feedback. We have applied the changes to the cloneDeep function to return the Resolved<T> type, as well as considered more native class cases. We appreciate your suggestions and have implemented them accordingly!

de-novo:
@bovas85 Thank you for your feedback! I have resolved the current issue related to cloning Event objects. However, the modified code now returns an empty object {} when attempting to clone Web spec classes like Event. While I tried to incorporate other Web API specifications similar to Event, the extensive number of cases within the Web API specifications makes fully supporting all of them practically unfeasible at this time.

Initially, I aimed to clone class instances such that instance => instance. However, due to the inability to access #private members, I have shifted to using the structuredClone specification.

de-novo:
@raon0211 The benchmark results show that the cloneDeep function is approximately twice as fast as Node’s structuredClone.  
스크린샷 2024-07-17 오전 10 50 00

bovas85:
> @bovas85 Thank you for your feedback! I have resolved the current issue related to cloning Event objects. However, the modified code now returns an empty object {} when attempting to clone Web spec classes like Event. While I tried to incorporate other Web API specifications similar to Event, the extensive number of cases within the Web API specifications makes fully supporting all of them practically unfeasible at this time.
> 
> Initially, I aimed to clone class instances such that instance => instance. However, due to the inability to access #private members, I have shifted to using the structuredClone specification.

Sounds good. I'll test tomorrow and let you know if it solves it with the extra cases added

seungrodotlee:
image

Similar to the case in #206, it returns an error in Node v18.

de-novo:
> image
> 
> Similar to the case in #206, it returns an error in Node v18.

@seungrodotlee  Fixed the problem

raon0211 (dismissed):
Let me merge this, and make the type of `cloneDeep` a little simpler.

We might discuss the `Resolved<T>` type in the following pull request: https://github.com/toss/es-toolkit/pull/206.

PR #149, opened 2024-07-10
feat(last):add last 

minchodang:
close #136. 

### description
I have created a function that returns the last element of an array. Below is a performance comparison with Lodash. If you have any questions or comments, please feel free to share.

Thank you.

### benchmark
스크린샷 2024-07-09 오후 4 59 45

PR #202, opened 2024-07-15
docs(last-ko): delete error category in last-ko docs

minchodang:
hello @raon0211. 

I remember that unnecessary error handling was deleted from the last function in the last pr #149. 
The English and Chinese are missing, but only the Korean docs are currently included, so I'll post a correction. 
thank you

- delete unnecessary error section in last-ko docs.


raon0211 (approved):
Thanks!

Issue #134, opened 2024-07-08
Support for `startCase`


Duck-98:
Hello, @raon0211 

Can I work `startCase` of #91 ?

raon0211:
Sure!

PR #203, opened 2024-07-15
feat(startCase): add startCase

Duck-98:
## Summary

#134  

This pull request implements a function `startCase`. This function takes a string as input and converts it so that the first letter of each word is capitalized. The words are separated by spaces, and the rest of the letters in each word are in lowercase.

The function handles various input formats, including:
- CamelCase strings (e.g., `fooBar`)
- Snake_case strings (e.g., `__FOO_BAR__`)
- Strings with different delimiters (e.g., `--foo-bar--`)

Key features:
- The function uses a regular expression to split the input string into words.
- Each word is processed to ensure the first letter is in uppercase and the rest in lowercase.
- Handles edge cases such as strings with only delimiters or mixed numbers and letters.

Below is a benchmark comparison between our `startCase` implementation and Lodash's `startCase`:

| Test Case                 | Our Implementation | Lodash Implementation |
|---------------------------|--------------------|-----------------------|
| '--foo-bar--'             | 'Foo Bar'          | 'Foo Bar'             |
| 'fooBar'                  | 'Foo Bar'          | 'Foo Bar'             |
| '__FOO_BAR__'             | 'FOO BAR'          | 'FOO BAR'             |
| 'XMLHttpRequest'          | 'XML Http Request' | 'XML Http Request'    |
| '_abc_123_def'            | 'Abc 123 Def'      | 'Abc 123 Def'         |
| '__abc__123__def__'       | 'Abc 123 Def'      | 'Abc 123 Def'         |
| '_-_-_-_'                 | ''                 | ''                    |
| '12abc 12ABC'             | '12 Abc 12 ABC'    | '12 Abc 12 ABC'       |

This comparison shows that our `startCase` implementation provides similar results to Lodash's implementation while handling various input cases effectively.

## Benchmark

스크린샷 2024-07-15 오전 11 14 31


src/string/startCase.ts
```diff
@@ -0,0 +1,30 @@
...
+  for (const word of words) {
+    if (result) {
+      result += ' ';
+    }
+    if (word === word.toUpperCase()) {
+      result += word;
+    } else {
+      result += word[0].toUpperCase() + word.slice(1).toLowerCase();
+    }
+  }
```
raon0211, line 28:
```suggestion
  return words.map(word => {
    if (word === word.toUpperCase()) {
      return word;
    } else {
      return word[0].toUpperCase() + word.slice(1).toLowerCase();
    }
  }).join(' ');
```

I thought this implementation is more readable. Is using for loops because they are more performant? If so, how much is it more performant?

Duck-98: image

Using the `for of` loop in the `startCase` function is faster than using the `map` function.
The image above compares `map[startCase]` and `for of[startCase]` using benchmark tests.

raon0211 (approved):
Thanks!

willmanduffy:
Hey @Duck-98 @raon0211 It looks like this handles a string like `HELLO_WORLD` differently than Lodash. Is that intentional?

I can open an issue if not.

CleanShot 2024-07-22 at 18 07 59

Duck-98:
@willmanduffy 
 When comparing the startCase function from es-toolkit and lodash in the lodash playground, do they produce the same result or different results?
 
image

PR #204, opened 2024-07-15
test(chunk,compact): Add property testing using fast-check

dengdengyee:
Implemented [property testing](https://en.wikipedia.org/wiki/Property_testing) using the [fast-check](https://github.com/dubzzz/fast-check) library.
Applied it to the `chunk` and `compact` functions as examples.
I hope to implement property tests for all functions in the future.

### DIFF
--- a/package.json
+++ b/package.json
@@ -137,6 +137,7 @@
     "eslint": "^8.56.0",
     "eslint-config-prettier": "^8.5.0",
     "eslint-plugin-jsdoc": "^48.5.0",
+    "fast-check": "^3.20.0",
     "lodash": "^4.17.21",
     "prettier": "^3.2.5",
     "tsup": "^8.1.0",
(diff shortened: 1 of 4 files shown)


raon0211:
Hello, thanks for your pull request! I understand the benefits of property-based testing, but I think the interface is not so great in JavaScript. It might be quite hard to maintain in our library. 

If there are some better libraries which provide great interface, I am happy to integrate it to our library. If not, I wish we don't merge this into our library. I'm thinking of the following interface:

```tsx
fc.assert(() => {
  const arr = fc.array()
  const size = fc.integer()

  for (const chunked of chunk(arr, size)) {
    expect(chunk.length).lessThanOrEqual(size);
})l
```


Do you know some other libraries that provide better interface What are your tho

raon0211:
Closing due to inactivity, but always we're always welcome when you're ready!

Issue #196, opened 2024-07-14
Suggest to enhance `clone()` function.


samchon:
Current `clone()` function considers only those native classes.

- `Date`
- `Set`
- `Map`
- `RegExp`

However, there are much more native classes in the JavaScript, especially about binary handling. So, I think the `clone()` function should consider them.

- `Uint8Array`
- `Uint8ClampedArray`
- `Uint16Array`
- `BigInt64Array`
- `Int8Array`
- `Int16Array`
- `Int32Array`
- `BigInt64Array`
- `Float32Array`
- `Float64Array`
- `ArrayBuffer`
- `SharedArrayBuffer`
- `DataView`
- `Blob`
- `File`

Also, current `clone()` function is returning the same `T` type with its parameter, but it is not correct. The returned type must be casted, because non-native classes are converted to primitve object type.

To solve this problem, the `clone()` function needs to return `Shallowed<T>` type like below.

```typescript
type Shallowed<T> = Equal<T, ShallowMain<T>> extends true ? T : ShallowMain<T>
type ShallowMain<T> = T extends [never]
  ? never
  : T extends object
  ? T extends 
    | Array<any> | Set<any> | Map<any, any> | Date | RegExp | Date
    | Uint8Array
    | Uint8ClampedArray
    | Uint16Array
    | Uint32Array
    | BigUint64Array
    | Int8Array
    | Int16Array
    | Int32Array
    | BigInt64Array
    | Float32Array
    | Float64Array
    | ArrayBuffer
    | SharedArrayBuffer
    | DataView
    | Blob
    | File
  ? T
  : {
    [P in keyof T]: T[P] extends Function ? never : T[P];
  }
  : T;
type Equal<X, Y> = X extends Y ? (Y extends X ? true : false) : false;
```

raon0211:
As you mentioned, our `clone` function should handle most JavaScript native objects. One thing that comes to my mind is that it will hurt bundle size, but accuracy is more important...

ahaoboy:
Maybe you should use https://developer.mozilla.org/en-US/docs/Web/API/structuredClone

PR #206, opened 2024-07-15
feat(clone): enhance type and logic.

samchon:
Current `clone()` function considers only those native classes.

- `Date`
- `Set`
- `Map`
- `RegExp`

However, there are much more native classes in the JavaScript, especially about binary handling.

So, I think that the `clone()` function should consider them.

- `Uint8Array`
- `Uint8ClampedArray`
- `Uint16Array`
- `BigInt64Array`
- `Int8Array`
- `Int16Array`
- `Int32Array`
- `BigInt64Array`
- `Float32Array`
- `Float64Array`
- `ArrayBuffer`
- `SharedArrayBuffer`
- `DataView`
- `Blob`
- `File`

Also, current `clone()` function is returning the same `T` type with its parameter, but it is not correct.

The returned type must be casted, because non-native classes are converted to primitve object type.

To solve this problem, the `clone()` function needs to return `Shallowed<T>` type like below.

```typescript
type Shallowed<T> = Equal<T, ShallowMain<T>> extends true ? T : ShallowMain<T>
type ShallowMain<T> = T extends [never]
  ? never
  : T extends object
  ? T extends
    | Array<any> | Set<any> | Map<any, any> | Date | RegExp | Date
    | Uint8Array
    | Uint8ClampedArray
    | Uint16Array
    | Uint32Array
    | BigUint64Array
    | Int8Array
    | Int16Array
    | Int32Array
    | BigInt64Array
    | Float32Array
    | Float64Array
    | ArrayBuffer
    | SharedArrayBuffer
    | DataView
    | Blob
    | File
  ? T
  : {
    [P in keyof T]: T[P] extends Function ? never : T[P];
  }
  : T;
type Equal<X, Y> = X extends Y ? (Y extends X ? true : false) : false;
```

- related issue: https://github.com/toss/es-toolkit/issues/196
- previous PR: https://github.com/toss/es-toolkit/pull/155

### DIFF
--- a/src/object/clone.ts
+++ b/src/object/clone.ts
@@ -35,50 +35,53 @@ export function clone<T>(obj: T): Shallowed<T> {
     return obj.slice() as Shallowed<T>;
   }
 
-  if (obj instanceof Date) {
-    return new Date(obj.getTime()) as Shallowed<T>;
-  }
-
-  if (obj instanceof RegExp) {
-    return new RegExp(obj.source, obj.flags) as Shallowed<T>;
-  }
-
-  if (obj instanceof Map) {
-    const result = new Map();
-    for (const [key, value] of obj) {
-      result.set(key, value);
-    }
-    return result as Shallowed<T>;
-  }
-
-  if (obj instanceof Set) {
-    const result = new Set();
-    for (const value of obj) {
-      result.add(value);
-    }
-    return result as Shallowed<T>;
-  }
-
-  // BINARY DATA
-  if (obj instanceof Uint8Array) return new Uint8Array(obj) as Shallowed<T>;
-  if (obj instanceof Uint8ClampedArray) return new Uint8ClampedArray(obj) as Shallowed<T>;
-  if (obj instanceof Uint16Array) return new Uint16Array(obj) as Shallowed<T>;
-  if (obj instanceof Uint32Array) return new Uint32Array(obj) as Shallowed<T>;
-  if (obj instanceof BigUint64Array) return new BigUint64Array(obj) as Shallowed<T>;
-  if (obj instanceof Int8Array) return new Int8Array(obj) as Shallowed<T>;
-  if (obj instanceof Int16Array) return new Int16Array(obj) as Shallowed<T>;
-  if (obj instanceof Int32Array) return new Int32Array(obj) as Shallowed<T>;
-  if (obj instanceof BigInt64Array) return new BigInt64Array(obj) as Shallowed<T>;
-  if (obj instanceof Float32Array) return new Float32Array(obj) as Shallowed<T>;
-  if (obj instanceof Float64Array) return new Float64Array(obj) as Shallowed<T>;
-  if (obj instanceof ArrayBuffer) return obj.slice(0) as Shallowed<T>;
-  if (obj instanceof SharedArrayBuffer) return obj.slice(0) as Shallowed<T>;
-  if (obj instanceof DataView) return new DataView(obj.buffer.slice(0)) as Shallowed<T>;
-  if (obj instanceof File) return new File([obj], obj.name, { type: obj.type }) as Shallowed<T>;
-  if (obj instanceof Blob) return new Blob([obj], { type: obj.type }) as Shallowed<T>;
-
   if (typeof obj === 'object') {
-    return Object.assign({}, obj) as Shallowed<T>;
+    // NATIVE CLASSES
+    if (obj instanceof Date) {
+      return new Date(obj.getTime()) as Shallowed<T>;
+    }
+
+    if (obj instanceof RegExp) {
+      return new RegExp(obj.source, obj.flags) as Shallowed<T>;
+    }
+
+    if (obj instanceof Map) {
+      const result = new Map();
+      for (const [key, value] of obj) {
+        result.set(key, value);
+      }
+      return result as Shallowed<T>;
+    }
+
+    if (obj instanceof Set) {
+      const result = new Set();
+      for (const value of obj) {
+        result.add(value);
+      }
+      return result as Shallowed<T>;
+    }
+
+    // BINARY DATA
+    if (obj instanceof Uint8Array) return new Uint8Array(obj) as Shallowed<T>;
+    if (obj instanceof Uint8ClampedArray) return new Uint8ClampedArray(obj) as Shallowed<T>;
+    if (obj instanceof Uint16Array) return new Uint16Array(obj) as Shallowed<T>;
+    if (obj instanceof Uint32Array) return new Uint32Array(obj) as Shallowed<T>;
+    if (obj instanceof BigUint64Array) return new BigUint64Array(obj) as Shallowed<T>;
+    if (obj instanceof Int8Array) return new Int8Array(obj) as Shallowed<T>;
+    if (obj instanceof Int16Array) return new Int16Array(obj) as Shallowed<T>;
+    if (obj instanceof Int32Array) return new Int32Array(obj) as Shallowed<T>;
+    if (obj instanceof BigInt64Array) return new BigInt64Array(obj) as Shallowed<T>;
+    if (obj instanceof Float32Array) return new Float32Array(obj) as Shallowed<T>;
+    if (obj instanceof Float64Array) return new Float64Array(obj) as Shallowed<T>;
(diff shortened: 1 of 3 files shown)


kakasoo:
It's a very complex type, but that doesn't mean it's overly pedantic. This is because this type is more like solving an error unlike the previous one. In most libraries, there was a problem that the clone type did not properly respond to the class type.

src/object/clone.ts
```diff
@@ -26,48 +26,103 @@
  * console.log(clonedObj); // { a: 1, b: 'es-toolkit', c: [1, 2, 3] }
  * console.log(clonedObj === obj); // false
  */
-export function clone<T>(obj: T): T {
+export function clone<T>(obj: T): Shallowed<T> {
```
raon0211, line 29: I think this type looks not as readable to me. What about making use of TypeScript's function overloading, like:

```tsx
export function clone<T>(obj: T[]): T[];
export function clone(obj: Date): Date;
export function clone(obj: RegExp): RegExp;

/* ... overloading for other native objects ... */

export function clone<T>(obj: T): Shallowed<T>;
```

```tsx
type Shallowed<T> = OmitNever<{
           [P in keyof T]: T[P] extends Function ? never : T[P];
         }>
```

src/object/clone.ts
```diff
@@ -26,48 +26,103 @@
...
+        | Float64Array
+        | ArrayBuffer
+        | SharedArrayBuffer
+        | DataView
+        | Blob
+        | File
+      ? T
+      : OmitNever<{
+          [P in keyof T]: T[P] extends Function ? never : T[P];
+        }>
```
raon0211, line 116: I'm not sure what this type does here. Could you explain what this is doing? 

I thought cloning some objects with methods, like `clone({ foo: () => {} })` will also work. However this type definition does not seem to respect it.

raon0211: image

raon0211: There could be many edge cases, so we might just return `T` from `clone`. After all, it's well-known that shallow cloning class instances doesn't copy methods.

src/object/clone.ts
```diff
@@ -26,48 +26,103 @@
...
+  if (obj instanceof Int16Array) return new Int16Array(obj) as Shallowed<T>;
+  if (obj instanceof Int32Array) return new Int32Array(obj) as Shallowed<T>;
+  if (obj instanceof BigInt64Array) return new BigInt64Array(obj) as Shallowed<T>;
+  if (obj instanceof Float32Array) return new Float32Array(obj) as Shallowed<T>;
+  if (obj instanceof Float64Array) return new Float64Array(obj) as Shallowed<T>;
+  if (obj instanceof ArrayBuffer) return obj.slice(0) as Shallowed<T>;
+  if (obj instanceof SharedArrayBuffer) return obj.slice(0) as Shallowed<T>;
+  if (obj instanceof DataView) return new DataView(obj.buffer.slice(0)) as Shallowed<T>;
+  if (obj instanceof Blob) return new Blob([obj], { type: obj.type }) as Shallowed<T>;
+  if (obj instanceof File) return new File([obj], obj.name, { type: obj.type }) as Shallowed<T>;
```
de-novo, line 78: The following line may not work as intended because File extends Blob: 
```suggestion
  if (obj instanceof File) return new File([obj], obj.name, { type: obj.type }) as Shallowed<T>;
  if (obj instanceof Blob) return new Blob([obj], { type: obj.type }) as Shallowed<T>;
```

de-novo: 스크린샷 2024-07-17 오전 2 02 35

src/object/clone.ts
de-novo: A developer working on the code might know if a value has been shallow copied, but other collaborators may not be aware without inspecting the code (specifically, the function returning the cloned value). Therefore, I believe the type proposed by samchon can help reduce misunderstandings among developers.

Using this type enhances code readability and prevents unnecessary bugs and confusion within the team. By employing the Resolved type, we can avoid the misconception of certain methods existing on plain objects, ensuring that developers understand the correct nature of the cloned value and can use the code more accurately.

de-novo: 스크린샷 2024-07-17 오전 9 01 34
Additionally, as shown in the screenshot, lodash and structuredClone do not support function cloning. It would be clearer for the clone function in es-toolkit to also not support functions

src/object/clone.ts
samchon: I think `Shallowed<T>` type is better than multiple function overloading definitions.

src/object/clone.ts
```diff
@@ -26,48 +26,103 @@
...
+        | DataView
+        | Blob
+        | File
+      ? T
+      : OmitNever<{
+          [P in keyof T]: T[P] extends Function ? never : T[P];
+        }>
+    : T;
+
+type Equal<X, Y> = X extends Y ? (Y extends X ? true : false) : false;
```
kakasoo, line 119: @samchon 

```ts
type Equal<X, Y> = X extends Y ? (Y extends X ? true : false) : false;
type Answer = Equal<1 | 2, 1 | 2>; // It infer boolean. I think it is weird.
```

If it is not wrapped in generic again, the union type branches into each element type to ride the conditional type, so if the two union types are the same, it is inferred to be boolean. For complete reasoning about the union type, the Equal type must be modified.

```ts
export type Equal<X, Y> =
    (<T>() => T extends X ? 1 : 2) extends
    (<T>() => T extends Y ? 1 : 2) ? true : false;

type Answer = Equal<1 | 2, 1 | 2>; // true
```

src/object/clone.ts
samchon: The `Shallowed<T>` type is required as @de-novo mentioned. If the source object type does not have any function property type, it `clone()` function will return the same type of original. Otherwise the source object some methods, the `Shallowed<T>` will remove the method properties.

By the way, if there's an object that defining functional property, I think it would better to change the logic of `clone()` function about the object type like below:

```typescript
return Object.fromEntries(
  Object.entries(obj).filter(([_key, value]) => value !== undefined && typeof value !== "function")
)
```

samchon: @raon0211 In the typ level, functional property explicitly removed, otherwise in the variable level, it still alive as a `never` type. It does not occur compiler level bug because of the characteristics of the `never` type, but seems ugly. @kakasoo We need to study about it.

image

image

src/object/clone.ts
samchon: Good

src/object/clone.ts
raon0211: What will happen to `clone({ foo: () => {} })`? Here, as far as I know, we can correctly clone the `foo` property.

seungrodotlee:
image

Node v18 and earlier versions do not natively support the `File` object, so the `clone` function you implemented results in an error.

samchon:
> image
> Node v18 and earlier versions do not natively support the `File` object, so the `clone` function you implemented results in an error.

Fixed the problem.

src/object/clone.ts
samchon: > I understand that `clone` does not clone class instance's methods.
> 
> I was wondering what will happen to `clone({ foo: () => {} })`; Here, we can correctly clone the `foo` property. However it will be prohibited by types.
> 
> As far as I know, we cannot easily distinguish class methods and function properties, since TypeScript uses structural typing.

To fix the problem, changed the implementation code.

```typescript
return Object.fromEntries(
  Object.entries(obj).filter(([_key, value]) => value !== undefined && typeof value !== "function")
)
```

raon0211: I think changing the implementation is awkward since most people would expect that `foo` will be properly cloned. `lodash` will clone those properties, and in JavaScript function properties are cloned when shallow copying an object. 

```tsx
const object = { foo: () => {} )
const cloned = { ...object };
```

raon0211: In my opinion, unless there is a TypeScript type that simply handles this case, we allow this behavior, because I think cloning a class instance is not a major use case. 

Please refer to our [contributing guide](https://github.com/toss/es-toolkit/blob/main/.github/CONTRIBUTING.md#12-simplicity) where we value simple implementation, and our functions will have complex implementation to suit every usecase.

Issue #91, opened 2024-06-28
⛳ Goal: Almost feature parity with lodash


raon0211:
For easy migration from lodash to es-toolkit, our goal is to achieve almost complete feature parity with lodash.

For details, please refer to our [compatibility documentation](https://es-toolkit.slash.page/compatibility.html).

We intend to migrate most of the functions that lodash offers, with the following exceptions:

- Functions that have specialized implementations for specific types of arrays, like `sortedUniq`.
   - Such functions fall outside the scope of es-toolkit.

If you encounter any lodash features that could complicate migration to es-toolkit, please inform us via comments on the issue.

## 👉 [Contribution guide](https://github.com/toss/es-toolkit/wiki/Contributing-to-%60es%E2%80%90toolkit-compat%60)

dcastil:
Was just checking whether we can replace lodash with es-toolkit and noticed that we're using the lodash `pick` function with deep keys, like

```ts
const pickedObject = pick(object, [
    'firstLevel.secondLevel.thirdLevel',
    'firstLevel.anotherKey.thirdLevel',
    'firstLevel.anotherKey.anotherKey',
])
```

Would it make sense for toolkit-es to provide a similar function? Maybe an API like this could work with type safety on the keys?

```ts
const pickedObject = pickDeep(object, [
    ['firstLevel', 'secondLevel', 'thirdLevel'],
    ['firstLevel', 'anotherKey', 'thirdLevel'],
    ['firstLevel', 'anotherKey', 'anotherKey'],
])
```

arty-name:
I’ve noticed another major lodash feature seemingly not being supported. I very much enjoy how concise lodash makes the following code:
```js
const email = 'test@example.com'
// ...
const user = find(users, { email })

KeKs0r:
Also did a short scan of my codebase and would love to replace lodash, mainly for the bad types of it.
The following seem to be not yet covered:
- `mapValues` & `mapKeys`
- `find` with object as query (as described by @arty-name 
- `map` - which I use to map over an object and return an array. I dont think doing these 2 things in 2 steps would be too bad though
- `max` and `min` - could easily be done by `maxBy` and `minBy`. Maybe worth considering having the identity function as default iteree
- `set` and `has` to check deal with deeply nested properties

Would love to hear your opinions. I don't think those necessarily need to be implemented, but I think extending and making the "not gonna implement" list as well, will be beneficial.

damianobarbati:
[...]

PR #208, opened 2024-07-15
feat(memoize): add memoize

de-novo:
# Summary
This PR addresses the goal outlined in Issue #91 by implementing the memoize function to cache the results of function calls based on the arguments provided. Additionally, function overloads have been added to enhance flexibility and support various use cases.

# Benchmark

스크린샷 2024-07-15 오후 3 04 25

# Changes Made

### Functionality:
- Implemented the memoize function to cache function results.

### Tests:
  - Added comprehensive test cases to ensure the correctness of the memoization functionality with different overloads.
  - Verified the correct functioning of memoize with default cache, custom resolver, and custom cache.
### Documentation:
  - Updated documentation with examples demonstrating the usage of memoize with different combinations of resolvers and caches.


raon0211:
Hello, this looks nice. Meanwhile, it seems that the function's implementation has become complicated to handle function overloading. The key design principle of `es-toolkit` is [simplicity](https://github.com/toss/es-toolkit/blob/main/.github/CONTRIBUTING.md#12-simplicity), to prefer implementation simplicity over handling edge cases.

Why not we get the resolver and cache by objects, and simplify our implementation?

```tsx
import { head } from '../../array/head.ts';

function memoize(func, { resolver = head, cache = new Map() }) {
  const memoized = function(...args) {
    const key = resolver(args);
    
    if (cache.has(key)) {
      return cache.get(key);
    }

    const result = func(...args);
    cache.set(result);
    return result;
  }

  memoized.cache = cache;

  return memoized;
}
```

raon0211:
Meanwhile, I'm not sure we should accept all types of functions in our `memoize`.

Why not accept functions that take a single argument in `es-toolkit` and have users pass unary functions to our `memoize`? If we only accept unary functions, implementing `memoize` becomes straightforward.

```tsx
function memoize(func, { cache = new Map() }) {
  const memoized = function(arg) {    
    if (cache.has(arg)) {
      return cache.get(key);
    }

    const result = func(...args);
    cache.set(result);
    return result;
  }

  memoized.cache = cache;

  return memoized;
}
```

I believe this approach is much simpler and easier to read. We could implement a separate `memoize` for compatibility in `es-toolkit/compat`.

de-novo:
> Meanwhile, I'm not sure we should accept all types of functions in our `memoize`.
> 
> Why not accept functions that take a single argument in `es-toolkit` and have users pass unary functions to our `memoize`? If we only accept unary functions, implementing `memoize` becomes straightforward.
> 
> ```tsx
> interface MemoizeOptions<P, R> {
>   cache?: Cache<P, R>;
> }
> 
> function memoize<P, R>(func: (arg: P) => R, { cache = new Map() }: MemoizeOptions<P, R> = {}) {
>   const memoized = function(arg: P): R {    
>     if (cache.has(arg)) {
>       return cache.get(key)!;
>     }
> 
>     const result = func(arg);
>     cache.set(result);
>     return result;
>   }
> 
>   memoized.cache = cache;
> 
>   return memoized;
> }
> ```
> 
> I believe this approach is much simpler and easier to read. We could implement a separate `memoize` for compatibility in `es-toolkit/compat`.

I think we need to consider the second suggestion a bit more.
```ts
```

raon0211:
I thought using the first argument as the default key for memoization is very implicit. 

So I thought of supporting `memoize` for functions that only get a single argument; this would enable us not consider edge cases and keep the behavior straightforward.

We might support `resolver`s in our compat version of memoize, but I thought supporting functions accepting only one argument might be enough in most cases.

de-novo:
The proposed approach to use the first argument as the default key for memoization is quite implicit. To simplify and avoid edge cases, we could support memoization only for functions that accept a single argument. This would keep the behavior straightforward.

raon0211:
Yes, you are right. We might accept a resolver here.

raon0211 (dismissed):
Hello, this pull request was open for a while, let me merge this and apply what we discussed earlier (making it an unary function and adding a resolver). 

Huge thanks for this contribution!

Issue #168, opened 2024-07-12
Support for `flatMap`


ssi02014:
Implements [lodash's flatMapDepth](https://lodash.com/docs/4.17.15#flatMapDepth) based on `flatten`. 

### Suggested flatMap
```ts
function flatMap<T, U, D extends number = 1>(
  arr: readonly T[],
  iteratee: (item: T) => U,
  depth = 1 as D
): FlatArray<U[], D>[] {
  return flatten(arr.map(iteratee), depth);
}
```

The existing [Array#flatMap](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap) is limited to a `depth of 1`. We can improve this inconvenience, and it's faster. 

<br />

```ts
const arr = [1, [2, 3], 4]

arr.flatMap((item) => [item, item]); // Array#flatMap
// return type: (number | number[])[]
// value: [1, 1, [2, 3], [2, 3], 4, 4]
```

<br />

💡 **Question: I'm wondering about the naming, should I go with `flatMap` like lodash? An alternative is `flattenMap`.** 

<br />

### Example
```ts
const arr = [1, [2, 3], 4]

flatMap(arr, (item) => [item, item]); // depth defulat: 1, It behaves the same as Array#flatMap. 
// return type: (number | number[])[]
// value: [1, 1, [2, 3], [2, 3], 4, 4]

flatMap(arr, (item) => [item, item], 1); // It behaves the same as Array#flatMap. 
// return type: (number | number[])[]
// value: [1, 1, [2, 3], [2, 3], 4, 4]

flatMap(arr, (item) => [item, item], 2);
// return type: number[]
// value: [1, 1, 2, 3, 2, 3, 4, 4]
```

## Note
[Array#flatMap - MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap)

ssi02014:
@raon0211 Can I proceed with the task? If you have any additional thoughts, I'd appreciate it. 🙏

ssi02014:
Furthermore, if we implement `flatMap`, we can extend it to `flatMapDeep`.

raon0211:
In general, I think `flatten` and `flatMap` is most used. So we could go on with `flatMap`.

ssi02014:
@raon0211 Thank you so much for your feedback!
Great, I'll go ahead and work with `flatMap`.

PR #209, opened 2024-07-15
feat(flatMap): add flatMap function

ssi02014:
close: #168 

Add a `flatMap` function based on `flatten`.
This improves on the shortcoming of `flatMap`, which can only be used at depth 1. 

It works the same as `map(iteratee).flat(depth)`, but is faster.

## benchmark
스크린샷 2024-07-15 오후 4 21 24

### DIFF
--- a/src/array/flatMap.ts
+++ b/src/array/flatMap.ts
@@ -1,0 +1,29 @@
+import { flatten } from './flatten';
+
+/**
+ * Maps each element in the array using the iteratee function and flattens the result up to the specified depth.
+ *
+ * @template T - The type of elements within the array.
+ * @template U - The type of elements within the returned array from the iteratee function.
+ * @template D - The depth to which the array should be flattened.
+ * @param {T[]} arr - The array to flatten.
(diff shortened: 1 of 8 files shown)


src/array/flatMap.ts
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * flatMap(arr, (item: number) => [[item, item]], 2);
+ * // [1, 1, 2, 2, 3, 3]
+ */
+export function flatMap<T, U, D extends number>(
+  arr: readonly T[],
+  iteratee: (item: T) => U,
+  depth = 1 as D
+): Array<FlatArray<U[], D>> {
+  return flatten(arr.map(iteratee), depth);
```
raon0211, line 28:
```suggestion
  return flatten(arr.map(item => iteratee(item)), depth);
```

We prefer inline lambda functions because we might accidentally pass `index` as the second argument to `iteratee`.

raon0211 (approved):

Issue #172, opened 2024-07-12
Why does `lowerCase`'s type signature is so general?


KonradLinkowski:
Why does [`lowerCase`](https://es-toolkit.slash.page/reference/string/lowerCase.html)'s type signature looks like that:
```
function lowerCase(str: string): string;
````
When [`capitalize`](https://es-toolkit.slash.page/reference/string/capitalize.html)'s looks like this:
```
function capitalize<T extends string>(str: T): Capitalize<T>;
```

Shouldn't it be:
```
function lowerCase<T extends string>(str: T): Lowercase<T>;
```
?

**Comment by changwoolab** [comment]
LowerCase<T> converts "first character" of string literal type to lowercase.
So it could not be applied to `lowerCase` in es-toolkit.

**Comment by raon0211** [comment]
I guess we should match `capitalize` if possible.

**Comment by raon0211** [comment]
Thanks for the suggestion!

Unfortunately, `lowerCase` in es-toolkit doesn't behave the same way as TypeScript's built-in `Lowercase<T>`. `Lowercase<T>` simply lowercases every character, while `lowerCase` splits the string into words and joins them with spaces:

```ts
type A = Lowercase<'helloWorld'>; // 'helloworld'

lowerCase('helloWorld'); // 'hello world'
```

Since the two are not equivalent, typing `lowerCase` as `Lowercase<T>` would return incorrect types and confuse users.

We could implement an exact template literal type that models the actual runtime behavior, but as our principles in [CONTRIBUTING.md](https://github.com/toss/es-toolkit/blob/main/.github/CONTRIBUTING.md) state, we focus on the major use cases — and returning exact literal string types isn't one of them, while it would make the implementation significantly more complex.

So I'm going to close this issue. Thanks again for taking the time to propose this!

PR #212, opened 2024-07-15
Fix(capitalize): remove template for consistent code, capitalize in string

haejunejung:
Remove template `T` in `capitalize` for consistent code formatting in string directory (snakeCase, kebabCase, etc...)

It suggested in #172 issue.

### DIFF
--- a/docs/ko/reference/string/capitalize.md
+++ b/docs/ko/reference/string/capitalize.md
@@ -5,12 +5,12 @@
 ## 인터페이스
 
 ```typescript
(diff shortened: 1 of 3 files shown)


raon0211:
Actually, I guess this is not intended.

For example, the original implementation correctly infers the type.

image

On the other hand, the new implementation does not infer the type.

image


I guess we should fix the implementation for other functions (`camelCase`, `lowerCase` etc.), not capitalize.

haejunejung:
Oh, I misunderstood. I fix the other functions, not capitalize and I will re-post the PR with the corrected content!

PR #213, opened 2024-07-15
fix(falsey): remove unused `falsey` constant in `src/compat`

haejunejung:
The `falsey` constant was removed as it was not being used anywhere in the codebase. I think that removing unused code helps in maintaining a cleaner and more maintainable codebase. So, I removed it.


raon0211 (approved):
Thanks!

PR #218, opened 2024-07-16
fix(drop): Fixed `drop` and `dropRight` only to allow positive integers

changwoolab:
This PR fixes `drop` and `dropRight` to only allow positive integers.
This makes `drop` and `dropRight` compatible with lodash.


raon0211 (approved):
Looks nice. Thanks for your help!

PR #220, opened 2024-07-16
perf(dropRightWhile): improve performance for dropRightWhile

cwrllab:
This PR improves the performance for dropRightWhile

<table>
  <tr>
    <th>before</th>
    <th>after</th>
  </tr>
  <tr>
<td>

before

</td>
<td>

after

</td>
  </tr>
</table>


raon0211 (approved):
Thanks!

PR #221, opened 2024-07-16
fix(invert): Fixed `invert` not to invert inherited properties

changwoolab:
`lodash's invert` does not invert inherited properties.
This PR fixes the implementation of invert to be compatible with lodash's implementation

You may close this PR if you want inherited properties to be inverted :)

### DIFF
--- a/src/object/invert.ts
+++ b/src/object/invert.ts
@@ -19,9 +19,9 @@
 export function invert<K extends PropertyKey, V extends PropertyKey>(obj: Record<K, V>): { [key in V]: K } {
   const result = {} as { [key in V]: K };
 
-  for (const key in obj) {
+  for (const key of Object.keys(obj)) {
     const value = obj[key as K] as V;
-    result[value] = key;
+    result[value] = key as K;
   }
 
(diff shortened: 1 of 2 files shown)


raon0211 (dismissed):
Thanks!

src/object/invert.ts
```diff
@@ -19,8 +19,10 @@
 export function invert<K extends PropertyKey, V extends PropertyKey>(obj: Record<K, V>): { [key in V]: K } {
   const result = {} as { [key in V]: K };
 
-  for (const key in obj) {
-    const value = obj[key as K] as V;
+  const keys = Object.keys(obj) as K[];
+  for (let i = 0; i < keys.length; i++) {
+    const key = keys[i];
+    const value = obj[key];
```
raon0211, line 25:
```suggestion
  const keys = Object.keys(obj) as K[];
  
  for (let i = 0; i < keys.length; i++) {
    const key = keys[i];
    const value = obj[key];
```

raon0211 (approved):

Issue #159, opened 2024-07-11
Improving fill Function to Support Negative Indices


wondonghwi:
**Description**
Currently, the fill function in lodash and the toFilled function in esToolkit support negative indices. However, the fill function in esToolkit does not support negative indices. Enhancing the fill function to support negative indices will ensure consistency with the fill function in lodash and the toFilled function in esToolkit.

**Proposed Solution**
To maintain consistency, we propose enhancing the fill function in esToolkit to support negative indices, similar to the fill function in lodash and the toFilled function in esToolkit. For more details, please refer to the [discussion here](https://github.com/toss/es-toolkit/pull/154).

raon0211:
Sure!

PR #222, opened 2024-07-16
feat(fill): Improving fill

wondonghwi:
#159 

The existing fill function had a limitation where it could not accept negative indices. To address this, I modified it to support negative indices, similar to lodash's fill and toFilled functions, and improved the TypeScript code. Additionally, I added and updated test cases to ensure consistency and functionality.

Negative Index Support:
- Modified the fill function to accept negative indices, allowing calculations from the end of the array.

TypeScript Improvements
- Clarified function type definitions and improved the interface.

Test Code Additions and Updates:
- Fixed typos in the toFilled function.
- Added test cases to verify the new functionality.
- Ensured consistency by using it in the test descriptions.

 I made efforts to ensure consistency between the toFilled and fill functions.

스크린샷 2024-07-17 오전 12 35 16


docs/ko/reference/array/fill.md
```diff
@@ -5,24 +5,26 @@
...
 
+음수 인덱스를 사용할 수도 있어요. 이 경우 배열의 끝에서부터 인덱스를 계산해요.
+
 ## 인터페이스
 
 ```typescript
-function fill<T>(array: unknown[], value: T): T[];
-function fill<T, P>(array: T[], value: P, start: number): Array<T | P>;
-function fill<T, P>(array: T[], value: P, start: number, end: number): Array<T | P>;
+function fill<T, U>(arr: Array<T | U>, value: U): Array<T | U>;
```
raon0211, line 13: I was wondering why we use two type parameters here; Doesn't the original type signature already infer the the correctly?

```tsx
function fill<T>(array: unknown[], value: T): T[];
```

wondonghwi: Hello, thanks your review!

AS_IS
`fill<T>(array: unknown[], value: T): T[];` 

TO_BE
`function fill<T, U>(arr: Array<T | U>, value: U): Array<T | U>;` 

Flexibility: To separate the element type (T) and the value type (U) of the array, allowing the array to hold elements of different types.

Clarity: To clearly define the element type and the value type of the array, better expressing the intent of the code.
Replace unknown: There was also an intention to replace the unknown type.

However, based on your feedback, since the original simple type signature is sufficient for correct type inference, it seems better to define the type as `fill<T>(array: unknown[], value: T): T[];`

Shall we change the type definition back to the original?

raon0211: Yes, I guess we should change the definition to the original one, since behavior is identical and the original implementation is simple.

end-jarry: I agree with your opinion. I have updated the original definition and revised the document accordingly. Thank you.

Original definition changed, and the document has been revised.

wondonghwi: I agree with your opinion. I have updated the original definition and revised the document accordingly. Thank you.

Original definition changed, and the document has been revised.

raon0211 (approved):

Issue #201, opened 2024-07-15
Support for ```set``` function


de-novo:
In reference to [PR #150](https://github.com/toss/es-toolkit/pull/150), it was suggested that implementing a type-safe set function would be beneficial. The existing lodash set function does not infer return types, requiring as type casting, which may lead to undetected developer errors. Thus, a set function that supports type inference is needed.

<img width="585" alt="스크린샷 2024-07-15 오전 9 24 58" src="https://github.com/user-attachments/assets/341ba13e-0865-4681-9ea6-059c828f7479">

raon0211:
That's right, and I think `lodash` already does this...

raon0211:
Note that [set will be added in the next es-toolkit/compat release](https://es-toolkit.slash.page/reference/compat/object/set.html).

PR #223, opened 2024-07-17
feat(set): add `set` function

de-novo:
# Functionality
- Implemented the set function to set values at specified paths within objects.

# Benchmark
스크린샷 2024-07-17 오후 12 34 17

스크린샷 2024-07-17 오후 12 27 45

스크린샷 2024-07-17 오후 12 28 58

The first benchmark comparison shows the performance when using memoize from PR #208. The second benchmark is the performance of the current code without using memoize.

Since lodash’s set function internally uses memoize, for an accurate performance comparison, PR #208’s memoize should be applied first.

# Tests
- Verified the correct functioning of set with nested objects, arrays, and creating non-existent paths.

# Documentation
- Updated documentation with examples demonstrating the usage of set with different combinations of paths and values.


de-novo:
#201

de-novo:
```ts
const stringToPath = memoize((string: string) => {
  return string
    .replace(/\[|\]/g, match => {
      return match === '[' ? '.' : '';
    })
    .split('.');
});
```

raon0211:

Meanwhile, I have some different thoughts on how we should implement `get` and `set`. I guess we should allow setting on `Map` and `Set` objects, since JavaScript does not prohibit it.

We will first merge this pull request and change the implementation a bit.

raon0211 (approved):

PR #224, opened 2024-07-17
feat(startsWith): add startsWith and endsWith string functions

bertyhell:
similar to
* https://lodash.com/docs/4.17.15#startsWith
* https://lodash.com/docs/4.17.15#endsWith

 ```typescript
 import { startsWith } from 'es-toolkit/string';

 startsWith('fooBar', 'foo') // returns true
 startsWith('fooBar', 'Bar') // returns false
 startsWith('fooBar', 'abcdef') // returns false
 startsWith('fooBar', 'Bar', 3) // returns true
 ```

 ```typescript
 import { endsWith } from 'es-toolkit/string';

 endsWith('fooBar', 'foo') // returns true
 endsWith('fooBar', 'Bar') // returns false
 endsWith('fooBar', 'abcdef') // returns false
 endsWith('fooBar', 'foo', 3) // returns true
 ```

### DIFF
--- a/src/string/endsWith.ts
+++ b/src/string/endsWith.ts
@@ -16,9 +16,5 @@
  * const isPrefix = endsWith('fooBar', 'abc', 5) // returns false
  */
 export const endsWith = (str: string, target: string, position: number = str.length): boolean => {
-  if (position) {
-    return str.slice(0, position).endsWith(target);
-  } else {
-    return str.endsWith(target);
-  }
+  return str.endsWith(target, position);
 };
--- a/src/string/endsWith.ts
+++ b/src/string/endsWith.ts
@@ -1,0 +1,24 @@
+/**
+ * Checks if a string contains another string at the end of the string.
+ *
+ * Checks if one string endsWith another string. Optional position parameter to offset searching before a certain index.
+ *
+ * @param {string} str - The string that might contain the target string.
+ * @param {string} target - The string to search for.
+ * @param {number} position - An optional position from the start to search up to this index
+ * @returns {boolean} - True if the str string ends with the target string.
+ *
+ * @example
+ * const isPrefix = endsWith('fooBar', 'foo') // returns true
+ * const isPrefix = endsWith('fooBar', 'bar') // returns false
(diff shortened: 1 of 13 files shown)


src/string/startsWith.ts
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * @example
+ * const isPrefix = startsWith('fooBar', 'foo') // returns true
+ * const isPrefix = startsWith('fooBar', 'bar') // returns false
+ * const isPrefix = startsWith('fooBar', 'abc') // returns false
+ * const isPrefix = startsWith('fooBar', 'Bar', 2) // returns true
+ * const isPrefix = startsWith('fooBar', 'Bar', 5) // returns false
+ */
+export const startsWith = (str: string, target: string, position: number = 0): boolean => {
+  return str.indexOf(target, position) === position;
```
raon0211, line 19:
```suggestion
  return str.startsWith(target, position);
```
What about using [String#startsWith](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith)?

src/string/endsWith.ts
```diff
@@ -0,0 +1,24 @@
...
+ * const isPrefix = endsWith('fooBar', 'abc') // returns false
+ * const isPrefix = endsWith('fooBar', 'foo', 3) // returns true
+ * const isPrefix = endsWith('fooBar', 'abc', 5) // returns false
+ */
+export const endsWith = (str: string, target: string, position: number = str.length): boolean => {
+  if (position) {
+    return str.slice(0, position).endsWith(target);
+  } else {
+    return str.endsWith(target);
+  }
```
raon0211, line 23: Couldn't we use `str.endsWith(target, position)` here?

bertyhell:
> What about adding these functions in `es-toolkit/compat`, because they are very easy to use in modern JavaScript using `String#startsWith` and `String#endsWith`?

I'm not sure what needs to happen to add these functions to es-toolkit/compat. Where does that code live? do you have an example file for an existing function?

raon0211:
You can implement them in our `./src/compat` directory. Currently we have a couple of functions there :)

PR #228, opened 2024-07-18
ci. fix circleCI Node.js version (20 -> 18)

seungrodotlee:
Based on the [documentation](https://es-toolkit.slash.page/installation.html#:~:text=es%2Dtoolkit%20supports%20Node.js%2018%20or%20later.), updating the CI to ensure compatibility with Node.js 18.

See #227.

Issue #227, opened 2024-07-18
`es-toolkit` supports Node.js 18 or later. But CI running based on Node.js 20


seungrodotlee:
The `es-toolkit` documentation states that [this library supports Node.js 18 and later](https://es-toolkit.slash.page/installation.html#:~:text=es%2Dtoolkit%20supports%20Node.js%2018%20or%20later.). However, since the CI is running on Node.js 20, it fails to catch errors that may occur in Node.js 18. 

**We need to take measures to ensure safe operation on Node.js 18.**

PR #229, opened 2024-07-18
ci. fix circleCI Node.js version (20 -> 18) - fixed

seungrodotlee:
Based on the [documentation](https://es-toolkit.slash.page/installation.html#:~:text=es%2Dtoolkit%20supports%20Node.js%2018%20or%20later.), updating the CI to ensure compatibility with Node.js 18.

See #227.

> Closed last PR(#228) due to a mistake during the merge process.


juno7803:
Then, It would be good to modify the `.nvmrc` file as well 😁
https://github.com/toss/es-toolkit/blob/2168a67045d70ed20cf52da827fb368b264bbcb8/.nvmrc#L1

seungrodotlee:
> Then, It would be good to modify the .nvmrc file as well 😁

@juno7803 Thanks! 👍 👍

raon0211:
Actually, I think that only the `test` job should be run on Node.js 18. (This is because I prefer to use the latest LTS version of Node.js if possible..) What are your thoughts on this?

seungrodotlee:
> Actually, I think that only the test job should be run on Node.js 18. (This is because I prefer to use the latest LTS version of Node.js if possible..) What are your thoughts on this?

I agree with your opinion! I guess running other tasks on Node.js 20 should not have any significant impact on ensuring stability!

seungrodotlee:
@raon0211 Rolled back the Node.js to v20 in the `.nvmrc` file and for tasks except test.

raon0211 (approved):
Thanks!

PR #230, opened 2024-07-18
fix. remove set from object/index

seungrodotlee:
There was an error while moving `set` to compat.
Fix by remove the `set` from `src/object/index.ts`


raon0211:
Oh! Seems that I fixed this in some other commit. Thanks for your pull request.

Issue #215, opened 2024-07-15
Support for 'get' function


scarletborat:
Hi @raon0211. 
Want to implement 'get' function from the list of #91.
Can I do it?

raon0211:
Hello, would you mind if I implement the `get` and `set` functions?

scarletborat:
Hello @raon0211, Sure ).
Is there some task I can help with?

raon0211:
Thanks for helping us! We will organize what should be done and ask for help soon. (Probably this weekend..)

Note that [get will be added to es-toolkit/compat](https://es-toolkit.slash.page/reference/compat/object/get.html#get) in the next release :)

PR #232, opened 2024-07-18
feat(get): add get

piquark6046:
# Summary
This PR addresses https://github.com/toss/es-toolkit/issues/215 and https://github.com/toss/es-toolkit/issues/91 by implementing the `get` function.

# Benchmark
Screenshot from 2024-07-18 17-11-36

### DIFF
--- a/src/object/get.ts
+++ b/src/object/get.ts
@@ -1,0 +1,48 @@
+/**
+ * Get a value from an object by path.
+ * 
+ * This function takes an object and a path, and default value that will be returned if
+ * a returnable value about the given path does not exist.
+ * 
+ * @template O - Type of the object to get value from. 
+ * @template T - Type of the default value to return if the value does not exist.
+ * @param {O} obj - The object to get value from. 
+ * @param {string | string[]} path - The path to get value from. 
+ * @param {T} defaultValue - The default value to return if the value does not exist.
+ * @returns {O[keyof O] | undefined} - The value from the object by the given path.
+ * 
+ * @example
+ * const obj = { a: { b: 3 } };
+ * get(obj, 'a.b', null); // 3
+ */
+
+export function get<O, T>(obj: O, path: string | string[], defaultValue?: T): O[keyof O] | T | undefined {
+  if (!Array.isArray(path)) {
+    let pathArr = path.split('.');
+    path = []
+    for (let i = 0; i < pathArr.length; i++) {
+      if (/^.+\[[0-9]+\]$/.test(pathArr[i])) {
+        path.push(/^.+(?=\[[0-9]+\]$)/.exec(pathArr[i])![0])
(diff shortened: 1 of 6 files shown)


de-novo:
@piquark6046  I would like to suggest an improvement to the types in this PR. The current issue is that when the path depth exceeds one level, the return type is inferred as `unknown`. It would be beneficial if the return value of the `get` function could be clearly inferred.
스크린샷 2024-07-18 오후 7 39 21

Below is a type example that addresses this issue. This is a type I have used personally, and while it may not work perfectly in all situations, it should serve as a useful reference.

```ts
type Split<STR extends string, SEP extends string> = string extends SEP
  ? string[]
  : STR extends `${infer A}${SEP}${infer B}`
    ? [A, ...(B extends '' ? [] : Split<B, SEP>)]
    : SEP extends ''
      ? []
      : [STR];

type ReplaceAll<T extends string, A extends string, B extends string> = T extends `${infer C}${A}${D}`
  ? `${C}${B}${ReplaceAll<D, A, B>}`
  : T;

export type StringPathToArray<T extends string> = Split<ReplaceAll<ReplaceAll<T, '[', '.'>, ']', ''>, '.'>;

export type At<T extends object | any[], K extends Array<string | number>> = K extends [
  infer F extends string | number,
  ...infer R extends Array<string | number>,
]
  ? F extends keyof T
    ? T[F] extends object | any[]
      ? At<T[F], R>
      : T[F]
    : F extends `${infer N extends number}`
      ? N extends keyof T
        ? T[N] extends object | any[]
          ? At<T[N], R]
          : T[N]
        : never
      : never
  : T;

interface Test {
  a: { b: { c: number[] } };
}

type test = At<Test, ['a', 'b', 'c', 0]>; // number
```

raon0211 (approved):
Hey, thanks for your contribution! 

We were in the process of working on the `get` function. Let me combine our version with the one you implemented.

raon0211:
Hello @de-novo , thanks for your suggestion!

Unfortunately, one of our core values of our library is [simple implementation](https://github.com/toss/es-toolkit/blob/main/.github/CONTRIBUTING.md#12-simplicity). I understand that your provided type helps our type safety, but I guess the type is too complicated to merge. We might open a new issue about this.

Let me merge this pull request, and let's discuss about correct types in the other issue if needed.

src/object/get.ts
```diff
@@ -0,0 +1,48 @@
...
+ * @returns {O[keyof O] | undefined} - The value from the object by the given path.
+ * 
+ * @example
+ * const obj = { a: { b: 3 } };
+ * get(obj, 'a.b', null); // 3
+ */
+
+export function get<O, T>(obj: O, path: string | string[], defaultValue?: T): O[keyof O] | T | undefined {
+  if (!Array.isArray(path)) {
+    let pathArr = path.split('.');
```
raon0211, line 21:
```suggestion
    const pathArr = path.split('.');
```

raon0211 (approved):

PR #233, opened 2024-07-18
test(clone): replace Foo class with object in clone benchmark

### LINKED ISSUE #234: Replace Foo class with object in clone benchmark
The current benchmark data used in the clone tests includes a custom `Foo` class, which does not accurately represent typical usage scenarios. This discrepancy can lead to unreliable benchmark results.

To improve the reliability and accuracy of our performance benchmarks, the benchmark data should be updated to use a standard object variable `obj` instead of the custom `Foo` class.

dayongkr:
This pull request updates the benchmark data used in the clone tests to replace the custom `Foo` class with a `obj` variable. 

Changes:
- Removed the `Foo` class from the lodash clone benchmark.
- Replaced the `Foo` class with the standard object `obj` in the lodash clone benchmark.

close #234


raon0211 (approved):
Thanks!

PR #238, opened 2024-07-18
chore(ko-docs-sidebar): ko-object-sidebar get,set link fix

minchodang:
Thank you for your hard work @raon0211, 

While looking at the document by chance, I found that the links to the Korean get and set sections were incorrectly entered. So I edited it and uploaded it.


raon0211 (approved):
Thanks!

Issue #235, opened 2024-07-18
🚚 Move test cases from `lodash` and validate `es-toolkit/compat`


raon0211:
## What is es-toolkit/compat?

`es-toolkit/compat` is a compatibility layer designed to help you transition from `lodash` to `es-toolkit`. It aims for 100% compatibility with lodash.

See our [compatibility docs](https://es-toolkit.slash.page/compatibility.html).

## Moving test cases from lodash

We need your help to move test cases from lodash to es-toolkit. Here's how you can contribute:

1. [Find](https://es-toolkit.slash.page/compatibility.html) a function in the "in review" stage in our compatibility docs.
2. [Migrate the corresponding test code](https://github.com/lodash/lodash/tree/main/test) from lodash.
   - Place the test code in our [compat directory](https://github.com/toss/es-toolkit/tree/main/src/compat). 
   - Adjust the code as needed, e.g., change `lodashStable.map(arr, fn)` to `arr.map(fn)`.
   - For which test cases to move, please refer to our [design principles](https://es-toolkit.slash.page/compatibility.html#design-principles).
   - Example: [The test code for drop](https://github.com/toss/es-toolkit/blob/main/src/compat/array/drop.spec.ts).
3. Run the migrated test code with `vitest ./src/compat/**/*`.
4. Fix any failing test cases.
    - Add compatibility implementations if needed. 
    - Example: [chunk](https://github.com/toss/es-toolkit/blob/main/src/compat/array/chunk.ts).
5. If all tests pass, update our [compatibility docs](https://es-toolkit.slash.page/compatibility.html).

dayongkr:
We can also run the migrated test code with `vitest compat/**/*`.

dayongkr:
At now, we can find the test codes in [here](https://github.com/lodash/lodash/tree/afcd5bc1e8801867c31a17566e0e0edebb083d0e/test)

PowerKiKi:
Should this issue be closed since 100% compat has been achieved by [v1.39.3](https://github.com/toss/es-toolkit/releases/tag/v1.39.3) ? or are some tests still missing ?

dayongkr:
Thanks for letting me know! I missed that!

PR #239, opened 2024-07-18
chore(head,tail): Add lodash tests

KiKoS0:
Add the lodash tests for both `head` and `tail` following this guideline https://github.com/toss/es-toolkit/issues/235:
- `head`: https://github.com/lodash/lodash/blob/6a2cc1dfcf7634fea70d1bc5bd22db453df67b42/test/head.spec.js#L1
- `tail`: https://github.com/lodash/lodash/blob/6a2cc1dfcf7634fea70d1bc5bd22db453df67b42/test/tail.spec.js#L1

### DIFF
--- a/src/array/tail.ts
+++ b/src/array/tail.ts
@@ -60,6 +60,8 @@ export function tail<T, U>(arr: readonly [T, ...U[]]): U[];
  * const result3 = tail(arr3);
  * // result3 will be []
  */
+export function tail<T>(arr: readonly T[]): T[];
+
 export function tail<T>(arr: readonly T[]): T[] {
   const len = arr.length;
   if (len <= 1) {
--- a/src/array/tail.ts
+++ b/src/array/tail.ts
@@ -60,8 +60,6 @@ export function tail<T, U>(arr: readonly [T, ...U[]]): U[];
  * const result3 = tail(arr3);
  * // result3 will be []
  */
-export function tail<T>(arr: readonly T[]): T[];
-
 export function tail<T>(arr: readonly T[]): T[] {
   const len = arr.length;
   if (len <= 1) {
--- a/src/compat/array/head.spec.ts
+++ b/src/compat/array/head.spec.ts
@@ -13,12 +13,6 @@ describe('head', () => {
     expect(head(array)).toBe(1);
   });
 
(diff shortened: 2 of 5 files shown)


src/compat/array/head.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  it('should return the first element', () => {
+    expect(head(array)).toBe(1);
+  });
+
+  // it('should return `undefined` when querying empty arrays', () => {
+  //   arrayProto[0] = 1;
+  //   expect(head([])).toBe(undefined);
+  //   arrayProto.length = 0;
+  // });
```
KiKoS0, line 20: It seems that lodash returns undefined even if the first element of the Array prototype is overwritten so I'm not sure if we should es-toolkit's `head` should support that or have a different compat version of it?

###### Lodash's implementation
```js
function head(array) {
    return array != null && array.length ? array[0] : undefined;
}
```

raon0211: This is a catch! I think we should not handle these cases, since prototype pollution is a problem anyway 😞  Updated our docs.

https://github.com/toss/es-toolkit/commit/eee44287e7fd21e52a154adeb9a9e60729231e18

src/array/tail.ts
```diff
@@ -60,6 +60,8 @@ export function tail<T, U>(arr: readonly [T, ...U[]]): U[];
  * const result3 = tail(arr3);
  * // result3 will be []
  */
+export function tail<T>(arr: readonly T[]): T[];
+
```
raon0211, line 64:
```suggestion
```

src/compat/array/head.spec.ts
raon0211, line 20:
```suggestion
```

src/compat/array/head.spec.ts
```diff
@@ -0,0 +1,30 @@
+import { describe, expect, it } from 'vitest';
+import { head, first } from '../index';
+
+const arrayProto = Array.prototype;
+
```
raon0211, line 5:
```suggestion
```

raon0211 (approved):
Thanks for your great help!

Issue #236, opened 2024-07-18
Support for `pick` in `es-toolkit/compat`


raon0211:
Support for `pick` is desirable. [Reference](https://github.com/toss/es-toolkit/issues/91#issuecomment-2223873187)

- `pick` with object paths like `a.b.c` should be in `es-toolkit/compat`. 
   - It would be easy to implement if we use [get](https://es-toolkit.slash.page/reference/compat/object/get.html#get).
- `pick` with simple object keys should be in the original `es-toolkit`.

seungrodotlee:
Hello, @raon0211! Let me implement this! :)

raon0211:
Thanks! :)

bitttttten:
Hey just a heads up that lodash pick support `const three = pick({ array: [1, 2, 3] }, "array[2]")` and also `const item = pick({ my: { deep: { item: true }}, "my.deep.item")`. Not sure how compatible you want to be although just wanted to make you aware of the array syntax too. Thanks!

seungrodotlee:
@bitttttten Thank you for information! I guess it should be implemented since [`es-toolkit/compat` is designed to provide an identical API to `lodash`](https://es-toolkit.slash.page/compatibility.html#:~:text=This%20module%20is%20designed%20to%20provide%20an%20identical%20API%20to%20lodash)

bitttttten:
It probably is a larger api and larger file size, so I can imagine `pick` and `pickDeep` existing side by side but then indeed the api will diverge from lodash.

jamierumbelow:
Here's a rough version of `pick` that supports nested paths; perhaps it'll be useful:

```
type Primitive = string | number | boolean | null | undefined;

type Path<T> = T extends Primitive
  ? never
  : T extends any[]
  ? never
  : {
      [K in keyof T & string]: K | `${K}.${Path<T[K]>}`;
    }[keyof T & string];

type PickResult<T, K extends Path<T>> = {
  [P in K]: P extends keyof T
    ? T[P]
    : P extends `${infer A}.${infer B}`
      ? A extends keyof T
        ? PickResult<T[A], B & Path<T[A]>>
        : never
      : never;
};

export const pick = <T extends Record<string, any>, K extends Path<T>>(
  obj: T,
  keys: K[],
): PickResult<T, K> => {
  const result = {} as PickResult<T, K>;
  for (const key of keys) {
    result[key] = get(obj, key);
  }
  return result;
};
```

akira28:
[...]

PR #240, opened 2024-07-19
feat(compat): add `pick`(compat ver)

seungrodotlee:
image

Add compat version of `pick`.
Test codes referenced by https://github.com/lodash/lodash/blob/main/test/pick.spec.js
See #236.


raon0211 (approved):
Thanks! I might have to add some docs and fix a little bit of implementation.

raon0211 (approved):
Thanks for your contribution! Didn't notice the pull request because it was down there 😓

PR #243, opened 2024-07-19
chore(ci): add Prettier format check to CI workflow

### LINKED ISSUE #241: Add Prettier Format Check to CI Workflow
## Background:
Currently, our CI workflow does not include a code format check. Adding this step would help maintain code consistency and reduce unnecessary discussions during code reviews.

## Proposed Solution:
I propose to add a prettier format check to our CI workflow. This would ensure that all pull requests are automatically validated for code style before being merged.

dayongkr:
This pull request adds a prettier format check to our CI workflow, ensuring that only modified {js, ts, tsx} files are checked against the formatting rules. The comparison is made with the origin/main branch.

close #241 

> Additionally, there are currently many unformatted files, so we will also need a pull request to format these files using the yarn format command.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -31,7 +31,7 @@ jobs:
       - setup
       - run:
           name: Lint
-          command: yarn eslint -c .eslintrc.js $(git diff --name-only --diff-filter=ACMRUXB origin/main | grep -E "(.js$|.ts$|.tsx$)")
+          command: yarn eslint -c .eslintrc.js $(git diff --name-only --diff-filter=ACMRUXB origin/main | grep -E "(.js$|.ts$)")
   formatcheck:
     docker:
       - image: cimg/node:20.12
@@ -40,7 +40,7 @@ jobs:
       - setup
       - run:
           name: Format Check
-          command: yarn prettier --check $(git diff --name-only --diff-filter=ACMRUXB origin/main | grep -E "(.js$|.ts$|.tsx$)")
+          command: yarn prettier -c $(git diff --name-only --diff-filter=ACMRUXB origin/main | grep -E "(.js$|.ts$)")
   typecheck:
     docker:
       - image: cimg/node:20.12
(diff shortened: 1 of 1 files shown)


dayongkr:
And I also make lint and formatcheck job  not to detact tsx files

.circleci/config.yml
```diff
@@ -31,7 +31,16 @@ jobs:
...
+          command: yarn eslint -c .eslintrc.js $(git diff --name-only --diff-filter=ACMRUXB origin/main | grep -E "(.js$|.ts$)")
+  formatcheck:
+    docker:
+      - image: cimg/node:20.12
+    steps:
+      - checkout
+      - setup
+      - run:
+          name: Format Check
+          command: yarn prettier -c $(git diff --name-only --diff-filter=ACMRUXB origin/main | grep -E "(.js$|.ts$)")
```
raon0211, line 43:
```suggestion
          command: yarn eslint -c .eslintrc.js $(git diff --name-only --diff-filter=ACMRUXB origin/main | grep -E "(.js$|.ts$)")
      - run:
          name: Prettier
          command: yarn prettier -c $(git diff --name-only --diff-filter=ACMRUXB origin/main | grep -E "(.js$|.ts$)")
```

What about we merge the `lint` and `format` like this? 

I guess it does not have to be run in separate containers.

dayongkr: I agree that!

dayongkr:
It has a issue when we dont have changed file then prettier throw a error. So I am finding a solution now.

dayongkr:
> It has a issue when we dont have changed file then prettier throw a error. So I am finding a solution now.

Done

raon0211:
image

Prettier seems to fail. Could you check why?

dayongkr:
> image
> Prettier seems to fail. Could you check why?

I change the logic(using `xargc -r`) for simplify but that command makes fails because `xargc -r` return exit with 1 status code when modified files doesn't exist.

so I rollback that logic.

raon0211 (approved):

PR #249, opened 2024-07-19
feat(clone): add copying a prototype

### LINKED ISSUE #247: clone has to copy prototype for being same Instance
Screenshot 2024-07-19 at 5 43 02 PM

Currently, we don't copy a prototype. So It clone instance like `ArrayBuffer` as just object instance.

dayongkr:
## Description

For copying its prototype, I add a `Object.create(Object.getPrototypeOf(obj))` instead of just `{}`.

This trick is common trick, you can see this code in [`shallowCopy` function of `immer.js`](https://github.com/immerjs/immer/blob/e2d222bd4fb26abded04075c936290715e9ee335/src/utils/common.ts#L181).

This makes it slightly slower than previous code, but more accurate.

### Previous benchmark

Screenshot 2024-07-19 at 5 48 19 PM

#### Don't clone as a same instance

Screenshot 2024-07-19 at 6 21 50 PM

#### Can't copy prototype's method.

Screenshot 2024-07-19 at 6 21 53 PM

### After benchmark

Screenshot 2024-07-19 at 6 27 42 PM

close #247


src/object/clone.ts
```diff
@@ -59,15 +59,14 @@ export function clone<T>(obj: T): T {
     return result as T;
   }
 
-  if (typeof obj === "object") {
-    return Object.assign({}, obj) as T;
+  if (typeof obj === 'object') {
+    return Object.assign(Object.create(Object.getPrototypeOf(obj)), obj);
```
raon0211, line 63:
```suggestion
    const prototype = Object.getPrototypeOf(obj);
    const result = Object.create(prototype)
    return Object.assign(result, obj);
```

raon0211 (approved):
Seems necessary. Thanks for your contribution!

PR #252, opened 2024-07-19
feat (take, compat) add `take` (compat version)

choi2021:
image

I implemented take in the compatibility version. 

The test codes are referenced from [lodash take.spec.js](https://github.com/lodash/lodash/blob/main/test/take.spec.js). 

I skipped `the should treat falsey 'n' values, except 'undefined', as '0'` test case because of type compatibility issues.


raon0211 (approved):
Thanks for your great help!

PR #253, opened 2024-07-19
test(flattenObject): change using constant date in test case

choi2021:
I have made a PR regarding the [take(compat) function](https://github.com/toss/es-toolkit/pull/252), but it failed due to issues in the flattenObject test case. 

The failure was caused by different date objects being used for comparison in the test cases. 

To fix this, I have modified the test cases to use a consistent date object for comparisons, ensuring that the date objects being compared are identical.

<table>
 <tr>
  <th>AS-IS </th>
</tr>
<tr>
  <td>

</td>
</td>
</tr>
 <tr>
  <th>TO-BE</th>
</tr>
<tr>
  <td> 
스크린샷 2024-07-15 오전 12 40 52
</td>
</tr>
</table>

### DIFF
--- a/src/object/flattenObject.spec.ts
+++ b/src/object/flattenObject.spec.ts
@@ -13,7 +13,7 @@ describe('flattenObject', function () {
       'a.b': 'yay'
     })
 
-    const date= new Date();
+    const date = new Date();
 
     const result2 = flattenObject({
       a: {
--- a/src/object/flattenObject.spec.ts
+++ b/src/object/flattenObject.spec.ts
@@ -39,7 +39,7 @@ describe('flattenObject', function () {
   })
 
   it('flattens multiple keys', () => {
-    const date= new Date();
+    const date = new Date();
 
     const result = flattenObject({
       a: {
--- a/src/object/flattenObject.spec.ts
+++ b/src/object/flattenObject.spec.ts
@@ -1,4 +1,4 @@
-import { describe, expect, it, test } from "vitest"
+import { describe, expect, it } from "vitest"
 import { flattenObject } from "./flattenObject";
 
 describe('flattenObject', function () {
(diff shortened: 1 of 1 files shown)


src/object/flattenObject.spec.ts
```diff
@@ -13,6 +13,8 @@ describe('flattenObject', function () {
       'a.b': 'yay'
     })
 
+    const date= new Date();
```
fvsch, line 16: Formatting: missing a space before the `=` sign.

src/object/flattenObject.spec.ts
```diff
@@ -32,11 +34,13 @@ describe('flattenObject', function () {
       'a.b.boolean': true,
       'a.b.null': null,
       'a.b.undefined': undefined,
-      'a.b.date': new Date(),
+      'a.b.date': date,
     })
   })
 
   it('flattens multiple keys', () => {
+    const date= new Date();
```
fvsch, line 42: Formatting: same issue here.

fvsch (approved):
This looks good to me.

There are a couple code formatting issues. This could be fixed by running Prettier on the whole file (`yarn prettier --write src/object/flattenObject.spec.ts`), but that would create a larger diff because the file has mixed formatting currently.

PS: I approved this PR even though I’m not a maintainer. I’m not sure what this project's policy is for review by non-maintainers. Is it fine to approve, or better to just comment?

src/object/flattenObject.spec.ts
raon0211:
```suggestion
    const date = new Date();
```

src/object/flattenObject.spec.ts
raon0211:
```suggestion
    const date = new Date();
```

raon0211 (approved):
Thanks!

Issue #254, opened 2024-07-19
Testing the dist output?


fvsch:
### The problem

While working on https://github.com/toss/es-toolkit/issues/216 to generate a cleaner `dist` folder (currently trying with Rollup instead of `tsup`), I’m having trouble making sure I don't regress the current build output.

The existing test suites seem to be pulling code from `src` directly:

- Benchmark tests tend to import from `'es-toolkit'` or `'es-toolkit/{category}'`, and it looks like Vitest is using the `package.json#exports` field to resolve that to TypeScript files in `src`:

```jsonc
{
  // …
  "exports": {
    ".": "./src/index.ts",
    "./array": "./src/array/index.ts",
    "./compat": "./src/compat/index.ts",
    "./function": "./src/function/index.ts",
    "./math": "./src/math/index.ts",
    "./object": "./src/object/index.ts",
    "./predicate": "./src/predicate/index.ts",
    "./promise": "./src/promise/index.ts",
    "./string": "./src/string/index.ts",
    "./package.json": "./package.json"
  },
  // …
}
```

- Unit tests are located alongside the TypeScript sources and tend to have two types of imports:
    - Relative, without an extension: `import { functionName } from './functionName';`
    - Relative, with an extension: `import { functionName } from './functionName.ts';`

As a result, we can't use those tests to verify that the build output in `dist` is correct and doesn't have issues like missing functions or files, flagrant runtime errors when evaluating a module, or more subtle bugs introduced by transpiling.

### Possible solutions

My original idea was to reuse existing unit tests (perhaps not the benchmarks) and forcing Vitest to resolve imports to the ESM or CJS builds. This can be done in `vitest.config.mts` with something like:

```ts
import path from 'node:path';
import { defineConfig } from 'vitest/config';

function libAliases() {
  const index = `index.${process.env.TEST_LIB_EXT || 'ts'}`;
  const baseDir = path.join(import.meta.dirname, index.endsWith('.ts') ? 'src' : 'dist');
  return [
    {
      find: /^es-toolkit$/,
      replacement: path.join(baseDir, 'index.js'),
    },
    {
      find: /^es-toolkit\/([a-z]+)$/,
      replacement: path.join(baseDir, '$1', 'index.js'),
    },
  ];
}

export default defineConfig({
  resolve: {
    alias: libAliases(),
  },
  // …
});
```

[...]

PR #255, opened 2024-07-19
chore: use rollup instead of tsup for a cleaner dist

### LINKED ISSUE #216: Dist contains bare imports of empty chunks in ESM output
In the `dist` folder produced by `yarn build`, a number of `.mjs` files contain bare imports for chunks which:

1. have some exports, but those exported names are not used by the bare imports.
2. Or contain no code at all (only a reference to an empty source map), for example:
   - https://cdn.jsdelivr.net/npm/es-toolkit@1.10.1/dist/chunk-QKYYOLJ6.mjs
   - https://cdn.jsdelivr.net/npm/es-toolkit@1.10.1/dist/chunk-NKLIE2OI.mjs
   - https://cdn.jsdelivr.net/npm/es-toolkit@1.10.1/dist/chunk-ARZ5BDR6.mjs

The [`dist/index.mjs`](https://cdn.jsdelivr.net/npm/es-toolkit@1.10.1/dist/index.mjs) module contains 11 bare imports, looking like this:

```js
import "./chunk-NKLIE2OI.mjs";
import "./chunk-QKYYOLJ6.mjs";
// …
```

And there are 93 other `dist/**/*.mjs` files containing 148 bare imports. For instance, [`dist/string/camelCase.mjs`](https://cdn.jsdelivr.net/npm/es-toolkit@1.10.1/dist/string/camelCase.mjs) looks like this:

```js
import {
  camelCase
} from "../chunk-Y5V7UAZZ.mjs";
import "../chunk-U56ZOM7L.mjs";
import "../chunk-WVQ7YVRK.mjs";
import "../chunk-2MM5EJJX.mjs";
export {
  camelCase
};
//# sourceMappingURL=camelCase.mjs.map
```

I wonder if those bare imports for empty files or unused exports affect the performance of the package. In Node it would add unnecessary disk i/o, and in the browser it would add unnecessary HTTP requests (e.g. if loading a function from `es-toolkit` from a CDN).

I suspect this was introduced by the move to `tsup` in https://github.com/toss/es-toolkit/pull/21, and I can see the issue in the published package all the way back to `1.1.0`.

It looks like there are a few issues in the https://github.com/egoist/tsup repository that might match this problem, but they haven't seen any activity. It looks like it might be a problem with multiple entries. Here are a couple issues:
https://github.com/egoist/tsup/issues/730
https://github.com/egoist/tsup/issues/841

fvsch:
Fixes #216.

This PR removes `tsup` and adds an equivalent Rollup configuration, in order to improve the structure and verbosity of the `dist` output.

### ESM build

- Presence of many chunks at `dist/chunk-{hash}.mjs`, often one per `es-toolkit` function, leading to hard-to-browse `dist` folder and a deeper-than-necessary module graph (`dist/index.mjs -> dist/array/index.mjs -> dist/array/unionWith.mjs -> dist/chunk-OG2RLT5U.mjs`).
- Presence of many bare imports (possibly used as a way to hoist transitive imports and cause the JS engine to import code earlier, but may not be ideal for re-bundling those sources, in addition to being noisy).
- Presence of many empty chunks (and bare imports of empty chunks).
- CJS build produces `dist/browser.mjs` (which duplicates `umd/browser.global.js` and is not exposed in `package.json#exports`).

### CJS build

- `dist/index.js` is a full bundle of the whole toolkit, instead of importing from the rest of the CJS output.
- `dist/{category}/index.js` are bundles for those categories.
- `dist/{category}/{functionName}.js` are also produced (along with their sourcemaps), but those files are not used
- CJS build produces `dist/{category}/{functionName}.js` files and their sourcemaps, but those files are never used (because `package.json#exports` points to the `dist/{category}/index.js` files, which are bundles and don't require `dist/{category}/{functionName}.js`).
- CJS build produces `dist/browser.js` (which duplicates `umd/browser.global.js` and is not exposed in `package.json#exports`).

## Proposed `dist` layout

```
dist/
  index.js         # CJS root entrypoint, requires all {category}/index.js
  index.js.map
  index.d.ts       # Declarations for CJS
  index.mjs        # ESM root entrypoint, imports all {category}/index.mjs
  index.mjs.map
  index.d.mts      # Declarations for ESM
  _chunk/
    *-[hash].js    # Small number of chunks produced in the CJS build
    *-[hash].js.map
  {category}/
    index.js       # CJS entrypoint, bundled code (may require from _chunk/*.js)
    index.js.map
    index.d.ts     # Declarations for CJS
    index.mjs      # ESM category entrypoint, imports from {category}/{fn}.mjs
    index.mjs.map
    index.d.mts    # Declarations for ESM
    {fn}.mjs       # ESM function implementation
    {fn}.mjs.map
    {fn}.d.mts     # Declarations for ESM
```

Things look maybe a bit easier to parse if we filter out by format. Here is the output structure for ESM:

```
dist/
  index.mjs        # root entrypoint, imports all {category}/index.mjs
  index.mjs.map
  index.d.mts
  {category}/
    index.mjs      # entrypoint, imports from {category}/{fn}.mjs
    index.mjs.map
    index.d.mts
    {fn}.mjs       # function implementation
    {fn}.mjs.map
    {fn}.d.mts
```

and for CJS:

```
dist/
  index.js         # root entrypoint, requires all {category}/index.js
  index.js.map
  index.d.ts
  _chunk/
    *-[hash].js    # small number of chunks (only 4 currently)
    *-[hash].js.map
  {category}/
    index.js       # entrypoint, bundled code (may require from _chunk/*.js)
    index.js.map
    index.d.ts
```

Note: for CJS there is some bundling going on, while for ESM we try to output a transpiled copy of the `src` dir (without tests) that mirrors its structure. I opted for this because that's what is going on currently with `tsup` (once you fix some useless files and some excessive indirection created by chunks).

It should be possible to do different things though:

1. Make `{category}/index` bundles of all the functions for that category, in both CJS and ESM builds.
2. Or make both CJS and ESM builds mirror the structure of the  `src` module graph (`{category}/index` entrypoints import function implementations from `{category}/{fn}`).

## Results of this PR

- I managed to get a `dist` output that mimics the current output, minus the outlined issues.
- Size of the generated `package.tgz`:
    - Before: 879 files, 309kB
    - After: 528 files, 164kB
- This requires roughly 200 LOC of configuration using Rollup, up from ~12 with `tsup`.
- Build times are a bit slower, but not by much: on my M1 Mac, `yarn pack` with `tsup` takes 3-4 seconds, and with Rollup it takes 4-5 seconds.

### Choice of bundler

The choice of bundler or transpiler is conditioned by the characteristics of this project:

- Sources use `.ts` extensions in import specifiers.
- Goal is to produce `.js` (CJS) and `.mjs` (ESM) builds.

This is a problem for both `tsc` and `esbuild`, since both have limitations (by design or de facto) when it comes to modifying extensions used in import specifiers.

With ESBuild, this can be worked around by using ESBuild to follow the whole module graph, then do code splitting to have it explode the module graph into many files. This looks like what `tsup` is configuring ESBuild to do. As a result the output has the configured extensions (`.js` or `.mjs`) in import specifiers, but a side effect is that we end up with most code in chunks at `dist/chunk-[hash].mjs`.

It looks like there is no way to get `tsup` and ESBuild to fulfill the requirements of this project.

When it comes to alternatives:

- I’ve used **Vite** to bundle libraries before. Its [library mode](https://vitejs.dev/guide/build.html#library-mode) is an opinionated configuration for Rollup. This _might_ work, but I suspect it might also fail some requirements.
- I knew that **Rollup** has a reputation of being a very flexible tool for compiling and/or bundling libraries, and I had some limited experience with it (and Vite). It's a lower-level tool than Vite (or `tsup`), but can be configured to control the output more closely through its options, existing plugins, or even custom plugins if need be.
- I haven't investigated `swc`, Parcel, or others.

### Duplication of code

With the 3 builds for this project (ESM, CJS, and the IIFE build in `umd/browser.global.js`), we end up with a lot of verbatim copies of the same code:

1. TypeScript declaration files for CJS and ESM have identical content.
    - For example, `dist/array/difference.d.ts` and `dist/array/difference.d.mts` are identical.
    - This is probably not fixable, as this is a requirement of TypeScript itself: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-7.html#packagejson-exports-imports-and-self-referencing

2. JSDoc comments for a function end up duplicated in many files.
    - For example, the source JSDoc comment for the `difference` function in `src/array/difference.ts` ends up copied verbatim into:
        ```
        dist/array/index.js          (CJS bundle)
        dist/array/difference.mjs    (ESM)
        dist/array/difference.d.ts   (CJS)
        dist/array/difference.d.mts  (ESM)
        ```
    - It should be possible to use TypeScripts `removeComments` option to not output comments in the `.js`/`.mjs` outputs, or in the `.d.ts`/`.d.mts` outputs. My intuition is that having JSDoc comments in the TypeScript declaration files only should be enough, but there might be IDEs or configurations where the IDE doesn't use the TypeScript declaration files but does use JSDoc comments from the imported `.js`/`.mjs` sources?
    - Edit: I checked the current output from `tsup`, and it looks like comments are kept in the declaration files but stripped in ESM/CJS modules. I'll update this PR to do the same.

3. Duplication of the original TypeScript sources:
    - Edit: discussion moved to https://github.com/toss/es-toolkit/issues/285

### Testing the output

I checked the generated `package.tgz` with https://arethetypeswrong.github.io/ and everything is green.

I would still like to verify the changes in this PR further:

- It might be possible to use `@arethetypeswrong/cli` on CI?
- See also https://github.com/toss/es-toolkit/issues/254: maybe that issue should be addressed first, by landing some integration tests, and this PR should only land after that?


raon0211:
Hello, thanks for your pull request! This would be a great help.

package.json
```diff
@@ -38,7 +38,7 @@
     "exports": {
       ".": {
         "import": {
-          "types": "./dist/index.d.mts",
+          "types": "./dist/index.d.ts",
```
raon0211, line 41: Please note that TypeScript type declarations have "types"; there's the CommonJS type declaration (`.d.cts`) and the ESM type declaration (`.d.mts`). 

We are checking them with https://arethetypeswrong.github.io.

fvsch: Are there specific tools that actually require the types for the ESM and CJS builds to be separate?

When I looked at the emmitted types with the tsup setup, I found that there were:

- `.d.ts` for CommonJS
- `.d.mts` for ESM

And checking a bunch of them, they were always identical. So having both seemed like needless duplication.

fvsch: Answering myself:

> It’s important to note that the CommonJS entrypoint and the ES module entrypoint each needs its own declaration file, even if the contents are the same between them. Every declaration file is interpreted either as a CommonJS module or as an ES module, based on its file extension and the "type" field of the package.json, and this detected module kind must match the module kind that Node will detect for the corresponding JavaScript file for type checking to be correct. Attempting to use a single .d.ts file to type both an ES module entrypoint and a CommonJS entrypoint will cause TypeScript to think only one of those entrypoints exists, causing compiler errors for users of the package.

From https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-7.html#packagejson-exports-imports-and-self-referencing

raon0211: You're right. Although the contents might be identical, it can bring a potential bug to our codebase; it is better for us to split the type declarations.

rollup.config.mjs
```diff
@@ -0,0 +1,172 @@
...
+ */
+const packageJson = createRequire(import.meta.url)('./package.json');
+
+const testPatterns = ['**/*.bench.ts', '**/*.spec.ts', '**/*.test.ts'];
+
+export default () => {
+  clearDir('dist');
+  clearDir('umd');
+
+  const entrypoints = Object.values(packageJson.exports).filter(f => /^(\.\/)?src\//.test(f) && f.endsWith('.ts'));
```
fvsch, line 24: After a bunch of testing: using the paths from `package.json#exports` (the dev version pointing to the `src` directory) as inputs to Rollup produces the best results:

- The whole module tree gets resolved correctly.
- This avoids pulling in test files, and modules not marked as test files but only consumed by tests (e.g. `src/compat/_internal/*.ts`).

Another approach is to use something like `globSync('src/**/*.ts')` and then do a lot of filtering, but I found that it's easy to forget something while filtering.

src/browser.ts
```diff
@@ -1,8 +0,0 @@
-import * as toolkit from './compat/index.ts';
-
-interface Window {
-  _: typeof toolkit;
-}
-
-declare let window: Window;
-window._ = toolkit;
```
fvsch, line 8: As described above, this is now handled with a Rollup build that looks like:

```js
const config = {
  input: 'src/compat/index.ts',
  output: {
    format: 'iife',
    name: '_',
    file: 'umd/browser.global.js',
  }
}
```

package.json
```diff
@@ -28,6 +28,7 @@
   },
   "files": [
     "dist",
+    "umd",
```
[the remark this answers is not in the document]
raon0211: Thanks for your catch!

raon0211 (dismissed):
This is decent work. Huge thanks to @fvsch !

raon0211 (approved):

raon0211 (approved):

PR #259, opened 2024-07-20
feat(fill): Add compatibility with lodash

### LINKED ISSUE #257: Support for compat fill
I will add a lodash test for fill

dayongkr:
Screenshot 2024-07-20 at 3 58 31 PM

I add a testcase based on [lodash/fill](https://github.com/lodash/lodash/blob/main/test/fill.spec.js) for compatibility at first, but it failed.

So I add a coercing integer by using `Math.floor` and add a converting to 0 from falsy value.

close #257


dayongkr:
``` javascript
it('should return a wrapped value when chaining', () => {
        const array = [1, 2, 3];
        const wrapped = _(array).fill('a');
        const actual = wrapped.value();

        expect(wrapped instanceof _)
        expect(actual).toBe(array);
        expect(actual, ['a', 'a').toEqual('a']);
});
```
I don't add this case, because this case is out of scope for es-toolkit/compat. Right?

raon0211 (approved):
Thanks a lot! This really helps.

PR #261, opened 2024-07-20
test(invert): add compatibility test case

### LINKED ISSUE #260: Support for compat/invert
Adding test cases based on Lodash helps me improve at finding edge cases. It's so cool!

dayongkr:
There are no fails. So I just add a test case and update compatibility docs.

close #260


raon0211 (approved):
Thanks!

PR #267, opened 2024-07-20
feat(isArray): Add `isArray`

mass2527:
This update introduces the [isArray](https://lodash.com/docs/4.17.15#isArray) function, which can be utilized to check if a value is an array. It also helps narrow the value type to an array or a readonly array.

Please review the changes and let me know if you have any feedback or suggestions.

### Benchmark

스크린샷 2024-07-20 오후 10 24 00

### DIFF
--- a/benchmarks/performance/isArray.bench.ts
+++ b/benchmarks/performance/isArray.bench.ts
@@ -1,0 +1,17 @@
+import { bench, describe } from 'vitest';
+import { isArray as isArrayToolkit } from 'es-toolkit';
+import { isArray as isArrayLodash } from 'lodash';
+
+describe('isArray', () => {
+  bench('es-toolkit/isArray', () => {
+    isArrayToolkit([1, 2, 3]);
+    isArrayToolkit('abc');
+    isArrayToolkit(() => {});
+  });
+
+  bench('lodash/isArray', () => {
+    isArrayLodash([1, 2, 3]);
+    isArrayLodash('abc');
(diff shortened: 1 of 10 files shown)


raon0211:
Looks great! Meanwhile, I think we should add this to our compatibility library since using `Array.isArray(...)` is much simpler in modern JavaSCript.

mass2527:
> Looks great! Meanwhile, I think we should add this to our compatibility library since using `Array.isArray(...)` is much simpler in modern JavaScript. Let me move this to our compatibility library.

Thanks for investing your time in reviewing my pull request!

The reason I added this function, even though `Array.isArray` can be simply and naturally used in modern JavaScript, is that `Array.isArray` always predicates value as `any[]` as we can see in the type definition:

```ts
interface ArrayConstructor {
   // ...
   isArray(arg: any): arg is any[];
}
```

This makes handling arrays uncomfortable. For example:

```ts
const arr1 = ['abc', () => {}, [1, 2, 3]];
const result1 = arr1.filter(Array.isArray);
// ^? any[][]

const arr2 = ['abc', () => {}, [1, 2, 3]];
const result2 = arr2.filter(isArray);
// ^? number[][]
```

In the above examples, `Array.isArray` results in `any[][]`, while using the custom `isArray` function results in `number[][]`, which provides more specific type information and makes array handling more convenient.

So I thought that this function can be even more useful than `Array.isArray` in modern JavaScript.

PR #269, opened 2024-07-20
test(initial,last,takeRight,without,uniq): Add lodash compat tests

KiKoS0:
Add the lodash compat tests for:
- [initial](https://lodash.com/docs/4.17.15#initial)
- [last](https://lodash.com/docs/4.17.15#last)
- [takeRight](https://lodash.com/docs/4.17.15#takeRight)
- [without](https://lodash.com/docs/4.17.15#without)
- [uniq](https://lodash.com/docs/4.17.15#uniq)

It also makes `takeRight` handle negative count correctly, let me know if it should be in a different PR or a different intended behavior is expected.


raon0211 (approved):
Thanks for your great help!

PR #273, opened 2024-07-21
fix(zipObject, zipObjectDeep, debounce, throttle, isEqual): type problems

nia3y:


raon0211 (dismissed):
Thanks!

src/function/throttle.ts
```diff
@@ -30,7 +30,7 @@ export function throttle<F extends (...args: any[]) => void>(func: F, throttleMs
   const throttledFunction = function (...args: Parameters<F>) {
     const now = Date.now();
 
-    if (lastCallTime == null || now - lastCallTime >= throttleMs) {
+    if (lastCallTime === null || now - lastCallTime >= throttleMs) {
```
raon0211, line 33:
```suggestion
    if (lastCallTime == null || now - lastCallTime >= throttleMs) {
```

PR #274, opened 2024-07-21
feat(take): add better return types

StyleShit:
I was bored and found this cool project, so... 🤷‍♂️😅


raon0211:
Hello, thanks for your contribution. 

Since we value [simplicity of the implementation](https://github.com/toss/es-toolkit/blob/main/.github/CONTRIBUTING.md#12-simplicity), we decided not to merge complex types in our repository.

However, other types of contributions such as adding new functions would be really helpful for us. Looking forward to your next contribution.

PR #275, opened 2024-07-22
feat(size, compat): implement size function

sunrabbit123:
This PR contains docs, compat test case, test case, benchmark and implement code about isLength.

Please review the changes and provide feedback or approval for merging. Thank you!


raon0211 (approved):
Thanks! Let me add this in our `compat` library, since I think we should use `Array#length` or `Set#size` instead when using TypeScript.

PR #277, opened 2024-07-22
feat: add test case of isNil (compat version)

choi2021:
image

I implemented isNil in the compatibility version.

The test codes are referenced from [lodash isNil.spec.js](https://github.com/lodash/lodash/blob/main/test/isNil.spec.js).

I skipped `'should work with nils from another realm'` test case.


raon0211 (approved):

raon0211:
Thanks for your pull request!

PR #288, opened 2024-07-23
test(ci): check the generated dist and package.tgz

### LINKED ISSUE #254: Testing the dist output?

fvsch:
Fixes #254

This PR adds a couple minimal checks of the `dist` output and `package.tgz` contents:

1. Use `@arethetypeswrong/cli` to check the contents of a generated `package.tgz`:
    - Running in CircleCI's `pre-pack` job.
    - Running in the GitHub Actions' `release` workflow.

2. Add a Node test (using `node --test` and the `node:assert` and `node:test` modules) that looks at `package.json#publishConfig#exports` and tries to import/require the ESM/CommonJS exports listed there.
    - Running in CircleCI's `pre-pack` job.
    - NOT running in the release workflow.

Ideally, for the second check, we would need a sub-package that consumes a generated `package.tgz` and tries to import/require from `es-toolkit` directly. But I'm not sure how to set that up, and a single test using Node's built-in test runner and ESM/CJS loaders (no Vitest or other test harness hiding resolution issues!) was a low hanging fruit.

Regarding where and when to run such checks: the `pre-pack` job looked like a good fit.

I wasn't sure if the `release` workflow should perform any check. After all, we don't run the test suite there. But we do run the `tsup` build, so maybe it makes sense to still have some basic checks to make sure that build is not broken?

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -42,6 +42,8 @@ jobs:
           mv jsr.json.tmp jsr.json
       - run: yarn install
       - run: 'mkdir -p out && yarn pack --out out/%s-%v.tgz'
+      - name: Check Package Entrypoints
+        run: 'yarn attw out/*.tgz'
       - id: extract-changelog
         uses: dahlia/submark@5a5ff0a58382fb812616a5801402f5aef00f90ce
         with:
--- a/.scripts/check-dist.mjs
+++ b/.scripts/check-dist.mjs
@@ -1,0 +1,77 @@
+import { deepStrictEqual, doesNotThrow } from 'node:assert';
+import path from 'node:path';
+import { describe, test } from 'node:test';
+import { createRequire } from 'node:module';
+
+const rootDir = path.resolve(import.meta.dirname, '..');
+const require = createRequire(import.meta.url);
+const { publishConfig } = require('../package.json');
+
+const entrypoints = getEntrypoints(rootDir, publishConfig.exports);
+
+test('all entrypoints are configured', () => {
+  deepStrictEqual(
+    entrypoints.map(entry => entry.name),
+    [
+      'es-toolkit',
+      'es-toolkit/array',
+      'es-toolkit/compat',
+      'es-toolkit/function',
+      'es-toolkit/math',
+      'es-toolkit/object',
+      'es-toolkit/predicate',
+      'es-toolkit/promise',
+      'es-toolkit/string',
+    ]
+  );
+});
+
+for (const { name, importFullPath, requireFullPath } of entrypoints) {
+  describe(`entrypoint: ${name}`, () => {
+    test(`can import esm`, async () => {
+      doesNotThrow(async () => await import(importFullPath));
+    });
+
+    test(`can require cjs`, async () => {
+      doesNotThrow(() => require(requireFullPath));
+    });
+
+    test(`identical esm and cjs export names`, async () => {
+      const esm = await import(importFullPath);
+      const cjs = require(requireFullPath);
+      const esmMap = objTypeMap(esm);
+      const cjsMap = objTypeMap(cjs);
+      deepStrictEqual(esmMap, cjsMap);
+    });
+  });
+}
+
+function getEntrypoints(baseDir = '', exports = {}) {
+  const entrypoints = [];
+  for (const [key, config] of Object.entries(exports)) {
+    if (
+      key.includes('.json') ||
+      typeof config?.import?.default !== 'string' ||
+      typeof config?.require?.default !== 'string'
+    ) {
+      continue;
+    }
+
+    entrypoints.push({
+      name: key.replace(/^\./, 'es-toolkit'),
+      importFullPath: path.resolve(baseDir, config.import.default),
+      requireFullPath: path.resolve(baseDir, config.require.default),
+    });
+  }
+
+  entrypoints.sort((a, b) => (a.name > b.name ? 1 : -1));
+  return entrypoints;
+}
+
+function objTypeMap(obj) {
+  return Object.fromEntries(
+    Object.keys(obj)
+      .sort()
+      .map(key => [key, typeof obj[key]])
+  );
+}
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -131,6 +131,7 @@
     }
   },
   "devDependencies": {
+    "@arethetypeswrong/cli": "^0.15.3",
```
fvsch, line 134: This dependency does add some weight to installed packages. I don't remember the numbers printed by Yarn, but with npm it takes:

- 37 packages
- 35 MB on disk

.github/workflows/release.yml
```diff
@@ -42,6 +42,8 @@ jobs:
           mv jsr.json.tmp jsr.json
       - run: yarn install
       - run: 'mkdir -p out && yarn pack --out out/%s-%v.tgz'
+      - name: Check Package Entrypoints
+        run: 'yarn attw out/*.tgz'
```
fvsch, line 46: This is the part I’m not sure about.

Should we have any smoke test before pushing the package, or is this not the right place?

.scripts/check-dist.mjs
```diff
@@ -0,0 +1,77 @@
+import { deepStrictEqual, doesNotThrow } from 'node:assert';
+import path from 'node:path';
+import { describe, test } from 'node:test';
+import { createRequire } from 'node:module';
+
+const rootDir = path.resolve(import.meta.dirname, '..');
+const require = createRequire(import.meta.url);
```
fvsch, line 7: Using an ES module so that we can use both `import()` and `require()`.

.scripts/check-dist.mjs
```diff
@@ -0,0 +1,77 @@
...
+    [
+      'es-toolkit',
+      'es-toolkit/array',
+      'es-toolkit/compat',
+      'es-toolkit/function',
+      'es-toolkit/math',
+      'es-toolkit/object',
+      'es-toolkit/predicate',
+      'es-toolkit/promise',
+      'es-toolkit/string',
```
fvsch, line 24: Fixed list of entrypoints that should be defined.

This is meant to prevent issues where we accidentally drop some values in `package.json#publishConfig#exports`.

raon0211: This looks great, while I think we could configure this using vitest, not Node.js' native test module. (It would be great if we could test this altogether with other unit tests.)

Let me modify this a little bit with vitest after merging this.

package.json
raon0211: Great comment! It might seem a little big, but I guess it's worth it, and it's devDependencies anyway.

raon0211 (approved):

.scripts/check-dist.mjs
fvsch: One possible issue with Vitest is that it's using Vite, which is built to be more tolerant of things like importing CJS sources as ESM modules, or ESM sources in a CJS context.

Using Node's testing capabilities is a bit "closer to the metal", which can be useful for this specific purpose.

But there's value in using Vitest for all the things too.

Just something to keep in mind.

Issue #271, opened 2024-07-21
Support for `mapKeys` and `mapValues`.


seungrodotlee:
`mapKeys` and `mapValues` can be useful when manipulating objects. If @raon0211 agrees to the implementation of these two functions, I will proceed to implement them.

raon0211:
These were one of the top priorities for us; It would be a great help if you help us!

raon0211:
This was added in our library. Thanks for your contribution!

- https://es-toolkit.slash.page/reference/object/mapKeys.html
- https://es-toolkit.slash.page/reference/object/mapValues.html

PR #291, opened 2024-07-23
feat(mapKeys,mapValues): add mapKeys and mapValues

seungrodotlee:
image

image

Both `mapKeys` and `mapValues` are little bit faster than `lodash`.

See #271


seungrodotlee:
One concern regarding the implementation is that while the current two functions do not modify the original object when calling iteratee, the object passed to iteratee is still mutable. 

This means that the actions of a previous iteratee can affect the actions of the next iteratee. Initially, I considered passing a newly cloned object with each iteratee call to address this issue, but this approach is performance-wise unfavorable, so I am exploring other options.

PR #310, opened 2024-07-24
docs(negate): fix link of function-negate on ko docs

shinwonse:
While looking at the official documentation in Korean, I saw that the link to 'function-negate' was specified as a link to the official documentation in English, so I created a pull request.


raon0211 (approved):
Thanks!

PR #312, opened 2024-07-24
fix(cloneDeep): SharedArrayBuffer could be undefined

hainan-612:
According to MDN

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer#description

the website must enable
```
Cross-Origin-Opener-Policy: same-origin
Cross-Origin-Embedder-Policy: require-corp
```
Otherwise, you'll get error: SharedArrayBuffer undefined

### DIFF
--- a/src/object/cloneDeep.ts
+++ b/src/object/cloneDeep.ts
@@ -87,7 +87,7 @@ export function cloneDeep<T>(obj: T): Resolved<T> {
 
   if (
     obj instanceof ArrayBuffer ||
-    (typeof window.SharedArrayBuffer !== 'undefined' && obj instanceof SharedArrayBuffer)
+    (typeof SharedArrayBuffer !== 'undefined' && obj instanceof SharedArrayBuffer)
   ) {
     return obj.slice(0) as Resolved<T>;
   }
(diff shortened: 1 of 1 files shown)


src/object/cloneDeep.ts
```diff
@@ -85,7 +85,10 @@ export function cloneDeep<T>(obj: T): Resolved<T> {
     return result as Resolved<T>;
   }
 
-  if (obj instanceof ArrayBuffer || obj instanceof SharedArrayBuffer) {
+  if (
+    obj instanceof ArrayBuffer ||
+    (typeof window.SharedArrayBuffer !== 'undefined' && obj instanceof SharedArrayBuffer)
```
raon0211, line 90:
```suggestion
    (typeof SharedArrayBuffer !== 'undefined' && obj instanceof SharedArrayBuffer)
```

Since this library can be used in Node.js, we should not implicitly use window here.

raon0211 (approved):
Thanks!

MangMax:
This problem also occurs in clone

PR #314, opened 2024-07-24
chore: move browser build from umd dir to dist

### LINKED ISSUE #284: Publishing issues with the browser.global.js build
The browser build, as published today (https://unpkg.com/browse/es-toolkit@1.13.1/umd/browser.global.js), has a couple issues:

1. It's referencing a sourcemap (``), but that sourcemap is not published in the package (https://unpkg.com/es-toolkit@1.13.1/umd/browser.global.js.map is a 404).
2. Its path is `umd/browser.global.js`, but its format is not UMD, it's an IIFE with a side-effect of adding an object as `window._`. [UMD is a specific format](https://github.com/umdjs/umd), and it would be best to not advertise UMD when that format is not actually provided.

Personally, I don't think UMD is useful nowadays. Using a simple IIFE format is a fine approach for CDN use. So I would just change the path, maybe output this bundle at `dist/browser.global.js`?

fvsch:
Fixes #284.

There were two issues reported in #284:

1. The source map for the `umd/browser.global.js` bundle was generated but not published in the package: fixed in #255.
2. The `umd/browser.global.js` path conveys the wrong information, because this bundle uses a simpler IIFE format declaring a global `_` variable, and not the UMD format.

This PR moves this bundle to `dist/browser.global.js`.

Because the Rollup config takes this path from `package.json#publishConfig#browser`, we only have to update it in `package.json`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,6 @@
   },
   "files": [
     "dist",
-    "umd",
     "*.d.ts"
   ],
   "publishConfig": {
@@ -36,7 +35,7 @@
     "main": "./dist/index.js",
     "module": "./dist/index.mjs",
     "types": "./dist/index.d.ts",
-    "browser": "./umd/browser.global.js",
+    "browser": "./dist/browser.global.js",
     "exports": {
       ".": {
         "import": {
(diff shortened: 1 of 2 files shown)


raon0211 (approved):

raon0211:
Thanks a lot!

PR #320, opened 2024-07-26
config(build): remove sourcemaps for esm/cjs builds

### LINKED ISSUE #285: Duplication of TypeScript sources in sourcemaps
(I’m extracting this question from #255 to keep the scope of that PR manageable.)

The current build of this library with `tsup`, and the proposed build config with Rollup in #255, produce an output where the TypeScript sources of this library are copied 3 times in published sourcemaps.

For example, let's take the contents of the source file `src/array/difference.ts`.

With the current `tsup` build config, they end up being copied into the `sourcesContent` of 8 sourcemaps:

```
dist/browser.js.map           (CJS sourcemap)
dist/chunk-JPUKH67K.mjs.map   (ESM sourcemap)
dist/index.js.map             (CJS sourcemap)
dist/array/difference.js.map  (CJS sourcemap)
dist/array/index.js.map       (CJS sourcemap)
dist/array/xor.js.map         (CJS sourcemap)
dist/compat/index.js.map      (CJS sourcemap)
umd/browser.global.js.map     (IIFE sourcemap)
```

(The CommonJS build is a big culprit here. This duplication internal to the CJS build is fixed in #255.)

With the proposed Rollup config in #255, this duplication is down to 3 files:

```
dist/array/index.js.map        (CJS sourcemap)
dist/array/difference.mjs.map  (ESM sourcemap)
umd/browser.global.js.map      (IIFE sourcemap)
```

But it's still three copies of the source TypeScript file.

I'm opening this issue to check that this duplication is okay, and ask whether we want to eliminate this duplication by publishing the TS files in the package.

It should be possible to publish the `src/**/*.ts` files (adding `src/**/*.spec.ts` to `.npmignore`), and set TypeScript's `inlineSources` option to `false` when building, leaving individual source maps looking like this:

```json
{
  "version": 3,
  "sources": ["../../src/array/difference.ts"],
  "sourcesContent": [],
  "mappings": ";;;;;;;;;;;;;;;;;;;;AAAA;AAAA;AAAA;AAAA;AAAA;AAsBO,SAAS,WAAc,UAAwB,WAA8B;AAClF,QAAM,YAAY,IAAI,IAAI,SAAS;AAEnC,SAAO,SAAS,OAAO,UAAQ,CAAC,UAAU,IAAI,IAAI,CAAC;AACrD;",
  "names": []
}
```

vs the current output which looks like:

```json
{
  "version": 3,
  "sources": ["../../src/array/difference.ts"],
  "sourcesContent": [
    "/**\n * Computes the difference ... firstArr.filter(item => !secondSet.has(item));\n}\n"
  ],
  "mappings": ";;;;;;;;;;;;;;;;;;;;AAAA;AAAA;AAAA;AAAA;AAAA;AAsBO,SAAS,WAAc,UAAwB,WAA8B;AAClF,QAAM,YAAY,IAAI,IAAI,SAAS;AAEnC,SAAO,SAAS,OAAO,UAAQ,CAAC,UAAU,IAAI,IAAI,CAAC;AACrD;",
  "names": []
}
```

Possible impacts:

- Positive: smaller size of package on disk.
- Positive but perhaps negligible: smaller size of package over the wire (this might be marginal, if gzipping handles most of the overhead from this duplication.
- Unknown risk: could it be that some tools only handle sourcemaps with `sourcesContent`, and do not load content from the `sources` paths?

Personally, I don't have a strong opinion either way. (For a library like this I would tend to not publish sourcemaps at all in the first place, as long as I’m publishing unminified code that is readable. But that's just a personal preference.)

fvsch:
Fixes #285.

The goal of this PR is to minimize the download and disk size of the `es-toolkit` package, while keeping a good developer experience for users of the package.

### Context for changes

See https://github.com/toss/es-toolkit/issues/285#issue-2424501137 for an overview of the problem. In summary: the current configuration ships the contents of this library's TypeScript sources several times in sourcemaps.

@raon0211 suggested that sourcemaps might not be necessary if the code shipped in `dist` was itself human readable (https://github.com/toss/es-toolkit/issues/285#issuecomment-2248418461).

The other option discussed was shipping the `src/**/*.ts` files in the package (excluding `src/**/*.spec.ts`), and sourcemaps that reference those files but don't copy their contents (this looks like: `"sources": ["../../src/array/difference.ts"], "sourcesContent": [null]`).

I tried many permutations of options for:

- sourcemaps with inline sources
- sourcemaps with refrences to `src` files only
- no sourcemaps
- JSDoc comments in JS output or not

and have posted my results and analysis here:
https://github.com/toss/es-toolkit/issues/285#issuecomment-2250501965

My takeaways were:

- Publishing `src/**/*.ts` and using sourcemaps with references to those files is not much of a win, compared to sourcemaps with inline sources.
- Best trade-off between package size and DX seems to be: no sourcemaps for ESM/CJS outputs, but with JSDoc comments conserved in those outputs.

### Summary of changes

As a result, this PR proposes the following changes:

1. Add JSDoc comments in ESM/CJS outputs (such as `dist/array/index.js` and `dist/array/difference.mjs`).
   - Cost: increases package size a little bit (mostly on disk, because those comments are identical between the `.js`, `.mjs`, `.d.ts` and `.d.mts` outputs, and should gzip well).
   - Benefit: better DX in contexts where TypeScript types are not used or not relevant, e.g. when stepping in and out of code in a debugger.

2. Remove sourcemaps for ESM/CJS outputs.
   - Cost: in some cases, debuggers might be able to map debugged code (such as a ESM module) back to the original TS source, and users might benefit from seeing the TS types in the debugger.
   - Mitigation of cost: shipping JSDoc comments in the debugged modules (see first point) is a decent alternative for this use case.
   - Benefit: big reduction in package size.

3. Keep a sourcemap for the browser bundle, and generate the browser bundle from the ESM build. Given that the ESM build will contain human-readable sources with JSDoc comments, treating it as the “source of truth” for the browser bundle is reasonable.
    - Cost: users of an in-browser debugger (e.g. Chrome DevTools's Sources panel) will be able to inspect unminified and JSDoc-documented sources for the toolkit, but won't see the TS types in those sources (they will see `dist/array/difference.mjs` and not `src/array/difference.ts`).
    - Benefit: small reduction in package size.

4. Tweak the browser bundle's minification to make the output more human-readable once the code is beautified (which DevTools does automatically).
    - Benefit: errors in the DevTools Console will use original function names; following the code in the beautified bundle in DevTools Sources will be easier.
    - Cost: +11% increase of `browser.global.js` (

### File and package size changes

| File | Bytes: before | Bytes: after |
| --- | --- | --- |
| `package.tgz` \* | 179,598 | 111,247 (-38%) |
| `package.tar` \* | 1,512,448 | 958,464 (-37%) |
| `dist/browser.global.js` | 16,762 | 18,686 (+11%) |
| `dist/browser.global.js.gz` \* | 5,169 | 5,573 (+8%) |
| `dist/browser.global.js.map` | 189,974 | 191,442 (+1%) |

Note: file names followed by a `\*` are not part of the package, but reflect what the size of the package itself might be over the wire or on disk:

- `package.tgz`: size of the package downloaded from repository
- `package.tar`: rough proxy for the size of the package on disk (e.g. at `node_modules/es-toolkit`)
- `dist/browser.global.js.gz`: size of the browser bundle served from a CDN (over HTTP(S), with gzip compression)

### DIFF
--- a/rollup.config.mjs
+++ b/rollup.config.mjs
@@ -114,7 +114,20 @@ function browserBuildConfig({ inputFile, outFile, name, sourcemap }) {
       }),
     ],
     output: {
-      plugins: [terserPlugin()],
+      plugins: [
+        // Minify with terser, but with a configuration that optimizes for
+        // readability in browser DevTools (after re-indenting by DevTools).
+        terserPlugin({
+          // Terser defaults to ES5 for syntax it adds or rewrites
+          ecma: 2018,
+          // Readable function names (not just in final export)
+          keep_fnames: true,
+          // Turn off compress.sequences to keep the assignments to the toolkit
+          // object readable, instead of turning them into a huge list of
+          // comma-separated expressions.
+          compress: { sequences: false },
+        }),
+      ],
       format: 'iife',
       name,
       file: outFile,
--- a/rollup.config.mjs
+++ b/rollup.config.mjs
@@ -2,12 +2,11 @@
 
 import fs from 'node:fs';
 import { createRequire } from 'node:module';
-import path from 'node:path';
-import { dirname } from 'node:path';
+import { dirname, join } from 'node:path';
+import { fileURLToPath } from 'node:url';
 import terserPlugin from '@rollup/plugin-terser';
 import tsPlugin from '@rollup/plugin-typescript';
 import dtsPlugin from 'rollup-plugin-dts';
-import { fileURLToPath } from 'node:url';
 
 // eslint-disable-next-line @typescript-eslint/naming-convention
 const __dirname = dirname(fileURLToPath(import.meta.url));
@@ -33,12 +32,14 @@ export default () => {
       extension: 'mjs',
       entrypoints,
       outDir: 'dist',
+      sourcemap: false,
     }),
     libBuildOptions({
       format: 'cjs',
       extension: 'js',
       entrypoints,
       outDir: 'dist',
+      sourcemap: false,
     }),
     declarationOptions({
       entrypoints,
@@ -48,6 +49,7 @@ export default () => {
       inputFile: './src/compat/index.ts',
       outFile: packageJson.publishConfig.browser,
       name: '_',
+      sourcemap: true,
     }),
   ];
 };
@@ -58,9 +60,10 @@ export default () => {
  *   format: 'esm' | 'cjs';
  *   extension: 'js' | 'cjs' | 'mjs';
  *   outDir: string;
+ *   sourcemap: boolean;
  * }) => import('rollup').RollupOptions}
  */
-function libBuildOptions({ entrypoints, extension, format, outDir }) {
+function libBuildOptions({ entrypoints, extension, format, outDir, sourcemap }) {
   const isESM = format === 'esm';
 
   return {
@@ -69,8 +72,8 @@ function libBuildOptions({ entrypoints, extension, format, outDir }) {
       tsPlugin({
         exclude: [...testPatterns],
         compilerOptions: {
-          sourceMap: true,
-          inlineSources: true,
+          sourceMap: sourcemap,
+          inlineSources: sourcemap || undefined,
           declaration: false,
           removeComments: true,
         },
@@ -83,7 +86,7 @@ function libBuildOptions({ entrypoints, extension, format, outDir }) {
       // Using preserveModules disables bundling and the creation of chunks,
       // leading to a result that is a mirror of the input module graph.
       preserveModules: isESM,
-      sourcemap: true,
+      sourcemap,
       generatedCode: 'es2015',
       // Hoisting transitive imports adds bare imports in modules,
       // which can make imports by JS runtimes slightly faster,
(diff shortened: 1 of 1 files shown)


rollup.config.mjs
```diff
@@ -69,10 +72,10 @@ function libBuildOptions({ entrypoints, extension, format, outDir }) {
...
         exclude: [...testPatterns],
         compilerOptions: {
-          sourceMap: true,
-          inlineSources: true,
+          sourceMap: sourcemap,
+          inlineSources: sourcemap || undefined,
+          removeComments: !sourcemap,
           declaration: false,
-          removeComments: true,
         },
```
fvsch, line 79: When compiling the ESM/CJS outputs (.js and .mjs), depending on the `sourcemap: boolean` value, this eithers:

1. generates sourcemaps but strips comments from the generated .js and .mjs outputs;
2. or generates no sourcemaps, and keeps comments in the generated .js and .mjs outputs.

Note that the present configuration only ever calls this function with `{ sourcemap: false }` (second option), so we could hardcode this config as:

```js
tsPlugin({
  compilerOptions: {
    sourceMap: false,
    removeComments: false,
    declaration: false,
  }
})
```

to remove one level of indirection in this config.

rollup.config.mjs
```diff
@@ -94,28 +97,41 @@ function libBuildOptions({ entrypoints, extension, format, outDir }) {
...
+        terserPlugin({
+          // Terser defaults to ES5 for syntax it adds or rewrites
+          ecma: 2018,
+          // Readable function names (not just in final export)
+          keep_fnames: true,
+          // Turn off compress.sequences to keep the assignments to the toolkit
+          // object readable, instead of turning them into a huge list of
+          // comma-separated expressions.
+          compress: { sequences: false },
+        }),
```
fvsch, line 129: A few changes to the Terser defaults to produce a more readable `dist/browser.global.js` file (once beautified):

- Keep function names. When this is off, the properties added to the `_` object are using the expected names, but most of them call to functions declared earlier in the bundle, and those function names are mangled (with code looking like `t.difference = function (...args) { return x$(...args) }`), resulting in JS errors being logged as an error in the `x$` function, which is not ideal.
- The `compress.sequences` feature of Terser leads to a particularly arcane output looking like:

```js
// line 300
return (
  t.fn1 = /* ... */,
  t.fn2 = /* ... */,
  /* ... */

// line 800
  t.fn93 = /* ... */,
  t
)
```

which is really hard to follow, and not actually smaller than the output with this option turned off:

```js
// line 300
t.fn1 = /* ... */;
t.fn2 = /* ... */;
/* ... */
// line 800
t.fn93 = /* ... */;
return t;
```

raon0211 (approved):
Looks great. Thanks for your great contribution!

PR #321, opened 2024-07-26
feat(without): improve types

StyleShit:


raon0211:
Hello, thanks for your contribution to our library! While I understand the reason of providing exact types, I am not sure they are worth it, considering maintainability and readability of complex types. Let me think a little bit about whether or not to include complex types to our library.

raon0211:
Hello, thanks for your contribution. 

Since we value [simplicity of the implementation](https://github.com/toss/es-toolkit/blob/main/.github/CONTRIBUTING.md#12-simplicity), we decided not to merge complex types in our repository.

However, other types of contributions such as adding new functions would be really helpful for us. Looking forward to your next contribution.

PR #322, opened 2024-07-27
fix(isLength): modernize checking a safe integer

dayongkr:
## description

- By using `Number.isSafeInteger`, I simplify `typeof value === 'number' && value % 1 === 0 && value <= Number.MAX_SAFE_INTEGER` to `Number.isSafeInteger(value)`
- And I also change `value > -1` to `value >=0`, because I think that after code is more acceptable.
- There are no difference of performance.

## after

``` typescript
export function isLength(value: unknown): value is number {
  return Number.isSafeInteger(value) && (value as number) >= 0;
}
```

Screenshot 2024-07-27 at 3 09 32 PM

## previous

``` typescript
export function isLength(value: unknown): value is number {
  return typeof value === 'number' && value > -1 && value % 1 === 0 && value <= Number.MAX_SAFE_INTEGER;
}
```

Screenshot 2024-07-27 at 3 09 50 PM

### DIFF
--- a/src/predicate/isLength.ts
+++ b/src/predicate/isLength.ts
@@ -1,3 +1,3 @@
 export function isLength(value: unknown): value is number {
-  return typeof value === 'number' && value > -1 && value % 1 === 0 && value <= Number.MAX_SAFE_INTEGER;
+  return Number.isSafeInteger(value) && (value as number) >= 0;
 }


raon0211 (approved):
Thanks for modernizing our codebase!

PR #323, opened 2024-07-27
feat(padStart): add padStart

### LINKED ISSUE #280: Support for `padStart`
According to the [compatibility documentation](https://es-toolkit.slash.page/compatibility.html), `padStart` has not been implemented yet, and I have not found any related issues either. Can I work on this?

pkovzz:
### Solution

Instead of implementing a custom algorithm, I have utilized the built-in [String.prototype.padStart()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padStart) function in a more type safe way.

### Benchmark
Screenshot 2024-07-27 at 15 40 14

Resolves #280


raon0211 (approved):
Thanks!

PR #325, opened 2024-07-28
feat(ary/unary): implement ary/unary

D-Sketon:
test case from 
- https://github.com/lodash/lodash/blob/main/test/unary.spec.js
- https://github.com/lodash/lodash/blob/main/test/ary.spec.js

benchmark
image

I'm not good at Korean, so I didn't add the corresponding documentation.


raon0211:
Thanks! Let me review this today.

raon0211:
I had some issues this week, let me review this until tomorrow!

raon0211 (dismissed):
Thanks for your contribution!

Seems that `guard` might not be necessary for our library, let me check this!

raon0211 (approved):

PR #330, opened 2024-07-30
feat(flatten): add compatibility with lodash

dayongkr:
## Description

- **Array Type Checking:** Added a check to ensure the input value is an array. If the value is not an array, the function returns an empty array.
- **Flattening Items:** Enhanced the flattening mechanism for items that are either:
  - Marked with `item[Symbol.isConcatSpreadable]` set to `true`
    - Even though the `Array.prototype.flat` method does not support this, it was added for compatibility with lodash.
  - Instances of `arguments`
    - Instead of using the `isArguments` function, expressions are directly added for better performance.
  - Items are forced to be of array type using `Array.prototype.from` for flattening the above types.

## benchmark

Screenshot 2024-07-30 at 5 52 14 PM


raon0211 (approved):
Thanks!

PR #331, opened 2024-07-31
docs(isPrimitive): fix isPrimitive function documentations

ssi02014:
Object wrapper `object` makes it a little simpler to determine if it's a primitive.

This logic is possible because

1. undefined, null values return a new object. 

```ts
Object(); // {}
Object(undefined); // {}
Object(null); // {}
```

2. The primitive value is returned as an object wrapped in a wrapper object.
```ts
Object(true); // Boolean {true}
Object(1); // Number {1}
Object('str'); // String {''}
```

3. If the argument is an object, return it as is. 
```ts
const obj = { a: 1 }
const newObj = Object(obj); // { a: 1 }

obj === newObj; // true

const set = new Set();
const newSet = Object(set);

set === newSet; // true
```

So if the return value of `Object(value)` is different from `value`, it means it's an object.
Doesn't this make the implementation logic much simpler?

### DIFF
--- a/docs/ko/reference/predicate/isPrimitive.md
+++ b/docs/ko/reference/predicate/isPrimitive.md
@@ -7,16 +7,16 @@ JavaScript 원시 값은 null, undefined, string, number, symbol, bigint를 말�
 ## 인터페이스
 
 ```typescript
-function isPrimitive(x: unknown): x is null | undefined | string | number | boolean | symbol | bigint;
+function isPrimitive(value: unknown): value is null | undefined | string | number | boolean | symbol | bigint;
 ```
 
 ### 파라미터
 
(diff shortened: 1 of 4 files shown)


raon0211:
Hello, thanks for your pull request!

Personally, I think it is less intuitive to understand `Object(value) !== value`. I think the current conditions are more intuitive and explicit. Would it be okay if we keep the current conditions?

ssi02014:
@raon0211 Yes! I follow the maintainer's opinion.
I'll revert the implementation logic.
But I think the jsdoc and some of the `documentation` will need to be fixed!

raon0211 (approved):
Thanks!

PR #333, opened 2024-07-31
docs(size): fix size title

vjo:
# Problem

The `reference/array/size.html` documentation page has the wrong "shuffle" title in the menu.
See:
Screenshot 2024-07-31 at 21 48 23

# Solution

Fix it.


raon0211 (approved):
Thanks!

PR #334, opened 2024-08-01
feat(flattenDeep, flattenDepth): add compatibility with lodash

### LINKED ISSUE #262: Support for compat/faltten, flatten{Depth, Deep}
In lodash, They wrote test cases for three methods in [one file](https://github.com/lodash/lodash/blob/main/test/flatten-methods.spec.js). 

``` javascript
describe('flatten methods', () => {
    const array = [1, [2, [3, [4]], 5]];
    const methodNames = ['flatten', 'flattenDeep', 'flattenDepth'];

...
```

Do we write them separately, or do we follow the same approach?

dayongkr:
## description 

I add `compat/flattenDeep` and `compat/flattenDepth` method based on `compat/flatten` function.

## benchmark

Screenshot 2024-08-01 at 4 57 59 PM

I don't add flattenDepth benchmark, because `flattenDepth` is same as `flatten` function.

close #262


dayongkr:
close #262

PR #337, opened 2024-08-03
feat(bind): implement bind

D-Sketon:
test case from https://github.com/lodash/lodash/blob/main/test/bind.spec.js, missing `chain` and `_.placeholder` related implementations.

benchmark: 
image


raon0211:
This looks great, but I guess this function has two features:

1. Binding `this` to the provided function.
2. Partially applying the function's argument.

I wish we support only (1) in `es-toolkit`; however we might support (2) in `es-toolkit/compat`.

Let me rewrite the implementation after merging this.

PR #339, opened 2024-08-03
docs(endsWith): Fix wrong returns in Examples

yhmpc:
endsWith('fooBar', 'foo') // returns true -> false
endsWith('fooBar', 'Bar') // returns false -> true


raon0211 (approved):
Thanks for finding this!

PR #345, opened 2024-08-04
fix(bind, CD): add sloppy-import flag and explicit symbol type for publishing deno

dayongkr:
## Description

Normally, deno doesn't infer file extensions from imports.

But we add `--unstable-sloppy-imports` flag, then deno will infer file extensions.

> ref. https://docs.deno.com/runtime/manual/tools/unstable_flags/#--unstable-sloppy-imports

And I also add explicit symbol type in `bind` function for fixing deno error.

## Before

Screenshot 2024-08-04 at 9 44 46 PM

## After

Screenshot 2024-08-04 at 9 44 52 PM

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -74,7 +74,7 @@ jobs:
           else
             npm publish --provenance --access public --tag dev *.tgz
           fi
-          deno publish --allow-dirty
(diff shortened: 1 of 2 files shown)


raon0211 (approved):
Thanks!

PR #346, opened 2024-08-05
feat(concat): Add concat funciton

ssi02014:
Is there a reason why `concat` only exists in `compat`?

First of all, the benchmark test files need to be modified. The rest seems to be worth checking.


raon0211:
In my opinion, using `Array#concat` is the recommended way to concat arrays; there is no need to use `concat` from `es-toolkit` in modern JavaScript. (I guess there might be performance implications, though..) What do you think?

ssi02014:
@raon0211 Oh, I get it, I agree.
Thanks for the feedback.

PR #350, opened 2024-08-06
chore(ci): add a branch linting all files in CI when .eslintrc.js, .prettierrc, .prettierignore files are changed 

### LINKED ISSUE #340: Check all files in ci, when `.eslintrc.js`, `.prettierrc`, `.prettierignore` files are changed
Now, we check linting or formatting only changed files in ci jobs.

But if we change configure files of linting or formatting library, then we should not check only changed files.

dayongkr:
## Description

- I removed a unused slack orb.
- I add a branch linting all files in CI when .eslintrc.js, .prettierrc, .prettierignore files are changed.
  - Because when we change the configure files, then maybe their rules are changed. So we need to lint all files that may have lint errors.

close #340

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,10 +29,25 @@ jobs:
       - run:
           name: ESLint
           shell: bash -e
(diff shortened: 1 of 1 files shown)


raon0211 (approved):
This is great. Thanks a lot!

PR #358, opened 2024-08-08
docs: fix Vercel deployment failures by async loading Sandpack component

kangju2000:
This PR fixes Vercel deployment failures by implementing async loading for the Sandpack component in our vitepress theme config. 

I believe the issue is caused by `vitepress-plugin-sandpack`'s Sandbox component using [DOM APIs on the server side](https://github.com/jerrywu001/vitepress-plugin-sandpack/blob/main/src/components/SandpackUtil.ts#L96). This change ensures Sandpack only loads client-side, preventing SSR-related errors. 

We can't verify the fix in this PR as Vercel deployment doesn't run on PRs. We'll need to merge to confirm the solution works.


raon0211 (approved):
Thanks, let me try this again.

raon0211:
Seems that it's not working, but I will open building docs in the pull request. Could you give it another try?

PR #357, opened 2024-08-08
feat(orderBy, isSymbol): add compatibility with lodash 

dayongkr:
## Description

I implement a `compat/orderBy`.

It is slower than `es-toolkit/orderBy`, but it is faster than `lodash/orderBy`. This is because lodash converts same strings to paths (e.g., `"user.name"` to `["user", "name"]`) during each comparison, whereas I convert the string to a path only once before sorting or comparing.

## Benchmark

Screenshot 2024-08-08 at 11 12 11 PM

close #328

PR #359, opened 2024-08-08
feat(isSymbol): implement isSymbol

D-Sketon:
image


src/predicate/isSymbol.ts
```diff
@@ -0,0 +1,15 @@
...
+ *
+ * This function can also serve as a type predicate in TypeScript, narrowing the type of the argument to `symbol`.
+ *
+ * @param {unknown} value The value to check.
+ * @returns {value is symbol} Returns `true` if `value` is a symbol, else `false`.
+ */
+export function isSymbol(value: unknown): value is symbol {
+  return (
+    typeof value === 'symbol' || (value !== null && typeof value === 'object' && getTag(value) === '[object Symbol]')
+  );
```
raon0211, line 14: While this is great, I think supporting `Object(symbol)`s is not desirable in our library (since I think it is an edge case, and it makes our implementation complicated).

Why not add this branch in our compatibility library, `es-toolkit/compat`?

For example, we have two versions of `isBoolean`:

- [one](https://github.com/toss/es-toolkit/blob/main/src/predicate/isBoolean.ts) in our original library
- [another](https://github.com/toss/es-toolkit/blob/main/src/compat/predicate/isBoolean.ts) in our compatibility library.

dayongkr:
In PR(#357), I already added this function in `es-toolkit/compat` because I need this function for implementing a `compat/orderBy`.

So I think that you don't need to add this function in `es-toolkit/compat`.

I'm sorry for the confusion. Thank you!

D-Sketon:
fixed

raon0211 (approved):
Thanks!

PR #365, opened 2024-08-09
feat(orderBy): add handling one is a string and the other is a number case

### LINKED ISSUE #361: Unexpected behavior using orderBy
Something wrong if the last element has another type 

Using `orderBy(data, ["value"], ["asc"])`

```
const data = [
  { id: 1, value: "apple" },
  { id: 2, value: "banana" },
  { id: 3, value: "cherry" },
  { id: 4, value: 10 },
  { id: 5, value: 5 },
  { id: 7, value: 20 },
  { id: 6, value: "banana" },
];

// Result:
[
  { id: 1, value: "apple" },
  { id: 2, value: "banana" },
  { id: 3, value: "cherry" },
  { id: 5, value: 5 },
  { id: 4, value: 10 },
  { id: 7, value: 20 },
  { id: 6, value: "banana" },
]
```

But if we change a bit data, its OK:
```
const data = [
  { id: 1, value: "apple" },
  { id: 2, value: "banana" },
  { id: 3, value: "cherry" },
  { id: 4, value: 10 },
  { id: 5, value: 5 },
  { id: 6, value: "banana" },
  { id: 7, value: 20 },
];

// Result:
[
  { id: 1, value: "apple" },
  { id: 2, value: "banana" },
  { id: 6, value: "banana" },
  { id: 3, value: "cherry" },
  { id: 5, value: 5 },
  { id: 4, value: 10 },
  { id: 7, value: 20 },
]
```
Example https://codesandbox.io/p/sandbox/es-toolkit-orderby-22zkzm

dayongkr:
## Description

If value `a` is a `'hello'` and value `b` is a `24`, then  value `a` will be converted to `NaN`.  So, `a > b` and `a < b` will be `false` in this case.

> Because Javascript converts to number, when types of the compared values are not same.

So in this update, I enhanced the `compareValues` function to handle cases where the types of `a` and `b` differ between strings and numbers. 

> Handling the other cases like object or array type is not good for our simplicity, So I just return `0` in these cases.

The function now converts number to string when necessary to ensure they can be compared correctly without causing NaN errors. 

Specifically, if `a` is a string and `b` is a number (or vice versa), the number is converted to a string before comparison. 

This adjustment prevents issues that might arise when comparing mixed types and ensures consistent behavior when sorting or ordering data.

If this enhancement is acceptable, then I will add this description in a document.

## Example

``` typescript
const data = [
  { id: 1, value: 'a' },
  { id: 2, value: 2 },
  { id: 12, value: 1 },
  { id: 5, value: 'b' },
  { id: 4, value: 2 },
  { id: 43, value: 'c' },
  { id: 24, value: 3 },
  { id: 3, value: '3a' },
  { id: 6, value: '2a' },
  { id: 7, value: '1cs' },
];

console.log(orderBy(data, ['value', 'id'], ['asc', 'asc']))
/*
[
  { id: 12, value: 1 },
  { id: 7, value: '1cs' },
  { id: 2, value: 2 },
  { id: 4, value: 2 },
  { id: 6, value: '2a' },
  { id: 24, value: 3 },
  { id: 3, value: '3a' },
  { id: 1, value: 'a' },
  { id: 5, value: 'b' },
  { id: 43, value: 'c' },
]
*/
```

close #361


src/array/_internal/compareValues.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+    expect(compareValues(1, 1, 'desc')).toBe(0);
+  });
+
+  it('should handle the case where a is string and b is number', () => {
+    const a = ['a', '1b', '3', '2'];
+    const actual = a.map(value => compareValues(value, 1, 'asc'));
+    const expected = a.map(() => 1);
+
+    expect(actual).toEqual(expected);
+  });
```
raon0211, line 35: I guess we could improve the test code like this:

```tsx
  it('should compare numbers and strings by converting number into a string', () => {
    expect(compareValues('0', 1, 'asc')).toBe(-1);
    expect(compareValues('a', 1, 'asc')).toBe(1);
    expect(compareValues('1b', 1, 'asc')).toBe(1);
    expect(compareValues('3', 1, 'asc')).toBe(1);
    expect(compareValues('2', 1, 'asc')).toBe(1);
   });
```

I think it is much easier to understand, and it is more comprehensive (since we handle the (`'0'` case)

src/array/_internal/compareValues.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  it('should handle the case where a is number and b is string', () => {
```
raon0211, line 37: We might improve the test code here, too.

src/array/_internal/compareValues.spec.ts
dayongkr: I agree!

raon0211 (approved):
Thanks!

PR #369, opened 2024-08-10
docs(orderBy): add information about PR #365

dayongkr:
## Description

I add information about PR #365.

I don't know chinese, so I only add this information in Korean and English docs.

### Korean docs

image

### English docs

image


raon0211:
Looking at the description, I am not sure if the feature we added was actually desirable. 

The key design principle of `es-toolkit` is performance and simplicity (please refer to our [contributing guide](https://github.com/toss/es-toolkit/blob/main/.github/CONTRIBUTING.md)), but it seems that what we added violates the two values.

1. It will have negative impacts on performance since we implicitly cast values.
2. Our implementation has become slightly complex than before.

Let me revert the original implementation asap (since it was deployed less than an hour ago), and let us discuss this again.

dayongkr:
> Looking at the description, I am not sure if the feature we added was actually desirable.
> 
> The key design principle of `es-toolkit` is performance and simplicity (please refer to our [contributing guide](https://github.com/toss/es-toolkit/blob/main/.github/CONTRIBUTING.md)), but it seems that what we added violates the two values.
> 
> 1. It will have negative impacts on performance since we implicitly cast values.
> 2. Our implementation has become slightly complex than before.
> 
> Let me revert the original implementation asap (since it was deployed less than an hour ago, let's regard this as a bug), and let us discuss this again.

I also worried about these points, and agree that PR #365 makes complex and slower.

Thank you!

PR #375, opened 2024-08-11
chore: move some of files to performance folder and change path of importing correctly

dayongkr:
## Description

- Some of files are moved to performance folder.
- A `set` function is only in `es-toolkit/compat`, so I changed a path of importing to `es-toolkit/compat`.
- `benchmarks/zeepObjectDeep.bench.ts` file is removed, because this file are also in `benchmarks/performance` folder.


raon0211 (approved):
Thanks!

PR #378, opened 2024-08-11
feat(indexOf): implement `indexOf` in `es-toolkit/compat`

### LINKED ISSUE #377: Support for `compat/indexOf`
I think ES already has a `Array.prototype.indexOf` function, so we don't need in `es-toolkit`.

So I am going to implement only in `es-toolkit/compat`

dayongkr:
## Description

I implement this function based on `Array.prototype.indexOf`.

Its behavior is  already similar with `lodash/indexof`.

> https://tc39.es/ecma262/#sec-array.prototype.indexof

But it can't find a `NaN` searchElement, so I add codes handling this case.

### Benchmark

Screenshot 2024-08-11 at 5 15 02 PM

close #377

### DIFF
--- a/benchmarks/performance/indexOf.bench.ts
+++ b/benchmarks/performance/indexOf.bench.ts
@@ -1,0 +1,23 @@
+import { bench, describe } from 'vitest';
+import { indexOf as indexOfToolkitComapt } from 'es-toolkit/compat';
+import { indexOf as indexOfLodash } from 'lodash';
+
+describe('indexOf', () => {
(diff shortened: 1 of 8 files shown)


dayongkr:
As I mentioned on a comment of issue(#377), I think we have to add this function also in `es-toolkit`.

Because `Array.prototype.indexOf` doesn't support finding `NaN` value.

raon0211 (approved):
Thanks for adding this!

dayongkr:
I will add docs of this function soon. Thanks!

PR #380, opened 2024-08-11
feat(compat): add `padEnd`

### LINKED ISSUE #332: Support for `padEnd`
According to the [compatibility documentation](https://es-toolkit.slash.page/compatibility.html), `padEnd` has not been implemented yet, and I have not found any related issues either. Can I work on this?

pkovzz:
### Solution

We can use the native [padEnd](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padEnd) API instead of a custom solution.

Screenshot 2024-08-11 at 18 17 00

Resolves #332


raon0211 (approved):
Thanks!

PR #381, opened 2024-08-12
feat(sortBy): implement `es-toolkit/sortBy`

### LINKED ISSUE #85: Support for `sortBy`
Support for [sortBy](https://lodash.com/docs/4.17.15#sortBy) is needed.

dayongkr:
## Description

I implement a `es-toolkit/sortBy` that sorts an array of objects based on the given iteratees or keys in ascending order.

This function have a custom key function feature differ from `es-toolkit/orderBy`.

> This function is based by `lodash/sortBy`, but more simple.

### Benchmarks

Screenshot 2024-08-12 at 10 43 26 AM

close #85

### DIFF
--- a/src/array/sortBy.ts
+++ b/src/array/sortBy.ts
@@ -7,7 +7,7 @@ type Iteratee<T> = (object: T) => T[keyof T];
  * It returns the sorted array, ordering by each key according to ascending order.
  * If values for a key are equal, it moves to the next key to determine the order.
  *
- * @template T - The type of elements in the array.
+ * @template T - The type of the objects in the array.
  * @param {T[]} collection - The array of objects to be sorted.
  * @param {Array<Iteratee<T>> | Array<keyof T>} iteratees - The array of iteratees or keys to sort by.
  * @returns {T[]} - The ascendingly sorted array of objects.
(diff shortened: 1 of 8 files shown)


src/array/sortBy.ts
```diff
@@ -0,0 +1,68 @@
...
+ * // results will be:
+ * // [
+ * //   { user : 'bar', age: 7 },
+ * //   { user : 'bar', age: 29 },
+ * //   { user : 'foo', age: 8 },
+ * //   { user : 'foo', age: 24 },
+ * // ]
+ */
+export function sortBy<T extends object>(collection: T[], iteratees: Array<Iteratee<T> | keyof T>): T[] {
+  const compareValues = (a: T[keyof T], b: T[keyof T]) => {
```
raon0211, line 36: I think this function is repeated quite often; maybe we might split it into an internal utility function sometime.

src/array/sortBy.ts
```diff
@@ -0,0 +1,68 @@
...
+
+      if (result !== 0) {
+        return result;
+      }
+    }
+
+    return 0;
+  });
+
+  return sortedCollection;
```
raon0211, line 67: Maybe just inlining `sortedCollection` and `shallowCopiedCollection` would make the code simpler.

```suggestion
  return collection.slice().sort((a, b) => {
    for (let i = 0; i < iteratees.length; i++) {
      const iteratee = iteratees[i];
      const iterateeIsFunction = typeof iteratee === 'function';

      const valueA = iterateeIsFunction ? iteratee(a) : a[iteratee];
      const valueB = iterateeIsFunction ? iteratee(b) : b[iteratee];

      const result = compareValues(valueA, valueB);

      if (result !== 0) {
        return result;
      }
    }

    return 0;
  });
```

raon0211 (approved):
Looks great to me. Thanks for your great contribution!

PR #382, opened 2024-08-15
refactor(sortBy, orderBy): use same compareValues function in internal and fix type

dayongkr:
## Description

I think that `orderBy` function also has to use `_internal/compareValues` function.

So I add a `order` argument in `_internal/compareValues` function.

## ⚠️ And I fix type of a criteria argument, it forgot parentheses.


raon0211 (approved):
Thanks for catching this!

Issue #383, opened 2024-08-15
Support custom key functions in `es-toolkit/orderBy` and `es-toolkit/compat/orderBy`


dayongkr:
## Problems

1. At now, we have a custom key function in `sortBy`, but `orderBy` doesn't have this feature.

Maybe, this makes confusion.

<img width="673" alt="Screenshot 2024-08-15 at 10 36 53 PM" src="https://github.com/user-attachments/assets/3008ff6c-b924-4660-a5e2-486f7a396034">

2. `lodash/orderBy` also supports this feature, but our `es-toolkit/compat/orderBy` doesn't support it.

## Solutions

1. Add it in `es-toolkit/orderBy`.
2. Add it in `es-toolkit/compat/orderBy`.

But after we also supports it in `es-toolkit/orderBy`, then  `es-toolkit/orderBy` looks like duplication of `es-toolkit/sortBy`.

> Difference of these functions is just it has a `order` argument or not.

So I think we have to integrate codes of these functions.

## Related issue

> https://github.com/toss/es-toolkit/issues/361#issuecomment-2280358793
> I guess in this case, the ideal way of using `orderBy` would be:
> 
> ```ts
> orderBy(data, x => x.value.toString(), ["asc"])
> ```
> 
> This would ensure that all numbers and strings are converted into string before comparison.

dayongkr:
I will work separately with two PRs.

At first, I will make a PR for supporting custom key functions in `es-toolkit/compat/orderBy`.

And a rest of PR is supporting it in `es-toolkit/orderBy` and integrating with a `es-toolkit/sortBy`

dayongkr:
If we add this feature in `es-toolkit/orderBy`, then how about makes `orderBy` deprecated and add a `orders` arguments in `sortBy`.

Two functions is too similar.

raon0211:
We might support both `sortBy` and `orderBy`, even though they’re pretty similar.

`sortBy` usually sorts in ascending order, like `Array#sort`. If we need to sort in descending order, that’s where `orderBy` comes in.

dayongkr:
> We might support both `sortBy` and `orderBy`, even though they’re pretty similar.
> 
> `sortBy` usually sorts in ascending order, like `Array#sort`. If we need to sort in descending order, that’s where `orderBy` comes in.

Almost programming languages have one sort function that has a reverse option, And javascript also have one sort function that has custom compare function.

And if we supports this feature in `orderBy` then `sortBy` code will be like this:

[...]

PR #385, opened 2024-08-16
feat(orderBy): implement a feature of custom key functions in a `compat/orderBy`

dayongkr:
## Description

image

`lodash/orderBy` has a feature of custom key functions, but `es-toolkit/compat/orderBy` doesn't have it.

I think that it makes confusion, so I added it.

### Benchmarks

Screenshot 2024-08-16 at 3 38 40 PM

related #383, #361


raon0211 (approved):
Thanks for your contribution!

PR #392, opened 2024-08-18
feat(join): add join

kim-sung-jee:
# Description

add `join` function

image

### DIFF
--- a/benchmarks/performance/join.bench.ts
+++ b/benchmarks/performance/join.bench.ts
@@ -1,0 +1,16 @@
+import { bench, describe } from 'vitest';
+import { join as joinToolkit } from 'es-toolkit';
+import { join as joinLodash } from 'lodash';
+
+describe('join', () => {
+  const arr = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10];
+  const separator = ',';
+
+  bench('es-toolkit/join', () => {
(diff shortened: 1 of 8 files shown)


dayongkr:
We already have a native api(Array.prototype.join), so maybe it needs only in es-toolkit/compat.

kim-sung-jee:
I understand that the es-toolkit may not include a join function as a feature. 
Does that seem correct?

dayongkr:
You can see a same case in [here(`es-toolkit/compat/concat`)](https://es-toolkit.slash.page/reference/compat/array/concat.html).

> This function is only available in es-toolkit/compat for compatibility reasons. It either has alternative native JavaScript APIs or isn’t fully optimized yet.

I hope this information is helpful.

kim-sung-jee:
@dayongkr 

oh thank you!
I'll close the PR.

dayongkr:
I think that you open the PR again after changing PR for adding `es-toolkit/compat/join`! Thank you!

And I think that you don't have to close this PR.

We have to decide this after reviewing @raon0211.

dayongkr:
I think that you don't have to close this PR at now, it is just my opinion.

Maybe, you decide whether close or not after reviewing @raon0211.

Thanks!

PR #396, opened 2024-08-18
chore(dev-deps): upgrade to eslint 9

### LINKED ISSUE #395: Upgrade to ESLint 9
ESLInt 9 is out and it seems like all the plugins are supported, so we can probably update to the latest version

StyleShit:
Closes #395

Updated the rules according to the [eslint](https://eslint.org/docs/latest/use/migrate-to-9.0.0#eslint-recommended) & [tseslint](https://typescript-eslint.io/blog/announcing-typescript-eslint-v8#rule-breaking-changes) docs

I can't make it work with VSCode though...
Updated the SDKs, but I'm not familiar with that, will be glad to get some help here 😅


src/compat/object/set.spec.ts
```diff
@@ -10,7 +10,6 @@ describe('set', () => {
       a: number;
     }
     const result = set<Test>({}, 'a', 1);
-    result.a;
```
StyleShit, line 13: wasn't sure why it was there, and eslint threw an error
is it necessary?

raon0211: Seems that it was from lodash's test code, I don't think it's necessary.

Issue #304, opened 2024-07-23
Support for `upperCase`


TiBianMod:
Support for [upperCase](https://lodash.com/docs/4.17.15#upperCase) is needed.

ivan-kleshnin:
How it that different from `str.toUpperCase`? 
I thought the goal of this library is to not reimplement functionality that exists (and is properly typed) in JS/TS natively 🤔

TiBianMod:
@ivan-kleshnin is not the same as `str.toUpperCase()`, take a look on lodash [upperCase](https://lodash.com/docs/4.17.15#upperCase) 

```js
_.upperCase('--foo-bar'); // => 'FOO BAR'
 
_.upperCase('fooBar'); // => 'FOO BAR'
 
_.upperCase('__foo_bar__'); // => 'FOO BAR'
```

TiBianMod:
already implemented by PR https://github.com/toss/es-toolkit/pull/402

PR #402, opened 2024-08-20
feat: add `upperCase` function

mattiacoll:
Solves issue #304 

## Benchmark

Screenshot 2024-08-20 alle 12 02 27


raon0211:
For long strings, we might try with the manual for loops instead of `Array#map` -- probably that's the reason why we are slow if the text is long.

raon0211 (approved):
Let me fix the implementation of `upperCase` a little bit to imperative `for` loops.

PR #405, opened 2024-08-20
fix(minBy): Fix typing of `minBy`

raviqqe:
Maybe, we should enable `noUncheckedIndexedAccess` compiler option for TypeScript too...


raon0211:
Hi, you're right about the suggestion. However, our current code convention is that functions selecting an element from an array shouldn't return `undefined`. I'd like our library to return `undefined` only when noUncheckedIndexedAccess is enabled. But, as far as I know, TypeScript doesn't allow reading compiler options directly.

raviqqe:
@raon0211 I changed the implementation to guarantee that an array length is always more than zero.

We can also allow both behaviour if it's preferrable:

```typescript
export function minBy<T>(
  items: readonly T[],
  getValue: (element: T) => number
): typeof items extends [T, ...T[]] ? T : T | undefined {
  // ...
}
```

raon0211:
What about we use function overloading as in [head](https://es-toolkit.slash.page/reference/array/head.html)? It would be much readable.

raviqqe:
It's ready for another review! 😄

raon0211 (approved):
Thanks!

PR #406, opened 2024-08-20
chore(debounce): fix return type

dayongkr:
# Description

Before the changes, return type used a Template type `F`, but we don't return a `func` function.

Actually, we return a debounced function. So, I change return type.

And more accuracy, I change type of extends target and add a test case.

## Before

Screenshot 2024-08-20 at 4 39 21 PM

> Correct return type is  `(a: number) => void) & { cancel: () => void; }`, but before type inference is not.

## After

Screenshot 2024-08-20 at 4 42 45 PM


src/function/debounce.spec.ts
```diff
@@ -17,6 +17,21 @@ describe('debounce', () => {
     expect(func).toHaveBeenCalledTimes(1);
   });
 
+  it('should not immediately call the function when debounce time is 0', async () => {
```
raon0211, line 20: Is this actually what we offer with debounce? I thought this was implementation detail.

raon0211 (commented):
Looks great! Please check one comment.

PR #407, opened 2024-08-20
feat(orderBy, sortBy): support custom key functions in `orderBy` and integrate `sortBy` with `orderBy`

### LINKED ISSUE #383: Support custom key functions in `es-toolkit/orderBy` and `es-toolkit/compat/orderBy`

dayongkr:
# Description

## `orderBy`

1. I add a feature of custom key functions in `orderBy`.
2. I changed handling a case that the number of orders is less than the number of criteria, for performance.
3. And I also integrate signature and docs based on `sortBy`.

## `sortBy`

Two functions are very similar, so I integrate `sortBy` with `orderBy`.

## Benchmarks

Screenshot 2024-08-19 at 11 00 43 AM

reopen #398 (I delete Repository that had the PR's branch by mistake. So I create new PR.)
close #383


raon0211 (approved):
Seems like what we needed. Thanks!

PR #408, opened 2024-08-20
chore(deps): remove unused babel and tsup package

dayongkr:
## Description

As I know, we use rollup for build.

So I think that we can remove the tsup and babel package.

reopen #401 (I deleted a repository that had the PR's branch by mistake. So I create new PR.)


raon0211 (approved):
Seems that it is unnecessary. Thanks!

PR #414, opened 2024-08-22
feat(throttle): add promise based feature

dayongkr:
# Description

At now, `es-toolkit/throttle` has only a time based feature.

But if we also support promise based feature, we can do this:

``` typescript
const throttledFetch = throttle(fetch);

throttledFetch('https://es.toolkit/api/login') // Start the first fetching 
throttledFetch('https://es.toolkit/api/login') // Ignored
// End the first fetching
throttledFetch('https://es.toolkit/api/login') // Start the second fetching
// End the second fetching
```

> Currently, if the `throttleMs` parameter is not provided, then It works promise based.

I think that this feature is also good for preventing double(or nth)-clicked.

I'm not sure this feature is good for `es-toolkit`, so I stop working. But this approach is acceptable, then I am going to improve interface and change docs.

### DIFF
--- a/src/function/throttle.spec.ts
+++ b/src/function/throttle.spec.ts
@@ -40,4 +40,17 @@ describe('throttle', () => {
     expect(func).toHaveBeenCalledTimes(1);
     expect(func).toHaveBeenCalledWith('test', 123);
   });
+
+  it('should work correctly if the throttled function is Promise-based and throttleMs is not provided', async () => {
+    const func = vi.fn(() => delay(1000));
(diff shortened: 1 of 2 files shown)


raon0211:
I am not really sure if we add this feature to throttle. However, we might add this called `throttleAsync` as an async library.

dayongkr:
> I am not really sure if we add this feature to throttle. However, we might add this called `throttleAsync` as an async library.

I agree with adding a method called throttleAsync. I will reopen the issue for further discussion.

PR #416, opened 2024-08-22
docs(isObjectLike): Update docs for `isObjectLike`

mass2527:
I've fixed the incorrect return type in the JSDoc and added the missing result in the example.


raon0211 (approved):
Thanks!

PR #417, opened 2024-08-23
feat(isRegExp): add `isRegExp` function

### LINKED ISSUE #299: Support for `isRegExp`
Support for [isRegExp](https://lodash.com/docs/4.17.15#isRegExp) is needed.

hyesungoh:
close #299 

> Hi there, I appreciate your great works for environment!

I append `isRegExp` function.

스크린샷 2024-08-23 오전 9 17 04


src/compat/predicate/isRegExp.ts
hyesungoh, line 1: I am confused for this compat function needed 🤔

> Does it always have to be the same as the lodash function?

raon0211: Yes, for `es-toolkit/compat`, it has to behave exactly the same with lodash. We try to copy the test cases from lodash like [here](https://github.com/lodash/lodash/blob/main/test/isRegExp.spec.js).

raon0211: Let me merge this and add the missing test cases!

raon0211 (approved):
Thanks for your contribution!

PR #420, opened 2024-08-23
fix(cloneDeep): Copy all properties from the source using `Reflect.ownKeys()`

ShaoClean:
```js
const obj = {
    a: 1,
    b: 2,
    [Symbol('1')]: 1,
};
Object.defineProperty(obj, 'c', {
    value: 3,
    enumerable: false,
});

console.log(Object.keys(obj));
// ["a", "b"]
console.log(Reflect.ownKeys(obj));
// ["a", "b", "c", Symbol(1)]
```


raon0211 (approved):
Thanks! Seems to be much better.

raon0211:
Seems that the new implementation is causing problems, let me check.

image

PR #421, opened 2024-08-23
fix(timeout): Add missing export for `timeout`

mass2527:
Looks like the `timeout` function was mistakenly omitted from index.ts.


raon0211 (approved):
Thanks for pointing this out!

PR #424, opened 2024-08-24
chore(lint): fix ignores option

dayongkr:
# Description

Before, We ignored files in benchmarks, docs and tests. But we have to lint these files.

So I removed that part and add some patterns for built and node_modules.


yhay81:
Looks good. I forgot to run prettier in #432.

raon0211 (approved):
Thanks!

PR #426, opened 2024-08-24
feat(rearg): implement rearg

D-Sketon:
benchmark
image


raon0211 (approved):
Thanks for your work! Let me add this to our compat library (since I think using a inline function, like `(a, b, c) => func(b, c, a)` would be much explicit

PR #427, opened 2024-08-24
fix(compat/orderby): add missed test cases and fix the problematic parts

dayongkr:
# Description

`compat/orderby` missed including these testcases(https://github.com/lodash/lodash/blob/6a2cc1dfcf7634fea70d1bc5bd22db453df67b42/test/sortBy-methods.spec.js) and hidden case(only implemented). And it also missed some of feature, so I added.

## ! Important

Currently, we had a bug that makes by below code:

``` typescript
// src/compat/array/orderBy.ts:81
keys = keys.map(key => getPath(key, collection[0]));
```

Before the changes, it checked the key is not a deep path only with the first element of the collection. So if the first element is different with others, then throw the TypeError like this:

``` typescript
orderBy([{ a: 1 }, { 'a.b': 3 }, { 'a.b': 1 }], ['a.b'], ['asc']) // TypeError: Cannot read properties of undefined (reading 'b')
orderBy([{ 'a.b': 3 }, { a: 1 }, { 'a.b': 1 }], ['a.b'], ['asc']) // success
```

So I changed the logic of checking keys similar with lodash(slower but safer). 

## Benchmarks

Screenshot 2024-08-25 at 12 16 31 AM

The performance result is changed, you can check result by a comment.


dayongkr:
I Improved performance by preparing all cases of criterion. So we don't need to using `isKey` (heavy function) each criterion.


Screenshot 2024-08-25 at 11 18 23 AM

src/compat/_internal/memoizeCapped.ts
```diff
@@ -0,0 +1,15 @@
+import { memoize } from '../../function';
+
+export function memoizeCapped<F extends (...args: Parameters<F>) => unknown>(func: F) {
```
raon0211, line 3: Personally, I really didn't want to add this function to `es-toolkit/compat`.. Does this make a big difference when it comes to performance?

dayongkr: Screenshot 2024-08-25 at 10 52 09 PM

This is result that not using it.

I think that it is good for performance.

dayongkr: Screenshot 2024-08-25 at 11 14 48 PM

I removed it!

Now, we just decrease counts of calling `toPath` function instead of memoizing.

dayongkr: While changing the logic, I realized that the existing code had too many conditional statements, so I also proceeded with refactoring.

src/compat/array/orderBy.ts
```diff
@@ -31,62 +33,85 @@ import { getPath } from '../_internal/getPath';
...
-      for (let i = 0; i < path.length; i++) {
-        value = value[path[i] as keyof typeof value];
-      }
+    if (typeof criterion === 'function') {
+      return criterion(object);
+    }
 
-      return value;
+    if (Array.isArray(criterion)) {
+      return getVaueByNestedPath(object, criterion);
```
raon0211, line 80:
```suggestion
      return getValueByNestedPath(object, criterion);
```

src/compat/array/orderBy.ts
```diff
@@ -31,62 +33,85 @@ import { getPath } from '../_internal/getPath';
...
+      return object[criterion as keyof typeof object];
     }
 
-    return obj[path as keyof typeof obj];
+    // Case for possible to be a deep path
+    if (Object.hasOwn(object, criterion.key)) {
+      return object[criterion.key as keyof typeof object];
+    }
+
+    return getVaueByNestedPath(object, criterion.path);
```
raon0211, line 92:
```suggestion
    return getValueByNestedPath(object, criterion.path);
```

dayongkr:
> Thanks for your great work!

After merging this PR, I am going to implement `compat/sortBy` based on `compat/orderBy`.

dayongkr:
> Thanks for your great work!

After merging this PR, I am going to implement compat/sortBy based on compat/orderBy.

dayongkr:
> Thanks for your great work!

After merging this PR, I am going to implement compat/sortBy based on compat/orderBy!

PR #428, opened 2024-08-25
feat(spread): Implement spread

D-Sketon:
benchmark
image


raon0211 (approved):
I think `spread` in the main version of `es-toolkit` should not accept the `startIndex` parameter, because mixing normal arguments and spread arguments is very confusing, anyway. Let me fix this a little bit after merging this.

PR #433, opened 2024-08-26
fix(countBy): Support nemeric and symbol keys in `countBy`

mass2527:
For consistency, the countBy function should support readonly arrays and non-string keys, like `groupBy`, so I made the change.


raon0211 (approved):
Thanks!

PR #434, opened 2024-08-26
feat(groupConsecutiveBy): Implement a new function on array to chunk consecutive elements with same key into groups

lqhuang:
Hi there,

Thanks for such amazing lib. I'm trying to transform consecutive discriminated union elements have the same key in an array into a new array with grouped unions, but I only found `groupBy` function.

Hence, here I want to propose a new function on array toolkits to transform things I described above.

```typescript
const array = [
  { category: 'fruit', name: 'apple' },
  { category: 'fruit', name: 'banana' },
  { category: 'vegetable', name: 'carrot' }
  { category: 'vegetable', name: 'tomato' }
  { category: 'fruit', name: 'orange' }
];
const result = groupConsecutiveBy(array, item => item.category);
// result will be:
// [
//   [
//     { category: 'fruit', name: 'apple' },
//     { category: 'fruit', name: 'banana' }
//   ],
//   [
//     { category: 'vegetable', name: 'carrot' }
//     { category: 'vegetable', name: 'tomato' }
//   ],
//   [
//     { category: 'fruit', name: 'orange' }
//   ]
// ]
```

You could see the details in src changes. And there are some notable comments

- [x] test cases added and passed. Exported in `index.ts`
- [x] en docs added
- [ ] ko docs
- [ ] zh-hans docs
- [ ] I'm not very sure the function name `groupConsecutiveBy` is enough concise and short. The function name could be changed if we have a better one.
- [ ] type hints are implemented as [Distributive Conditional Types](https://www.typescriptlang.org/docs/handbook/2/conditional-types.html#distributive-conditional-types). 
  - The type signature in docs is a simpler version, because it obviously not only works for discriminated union but has more general cases.
  - The problem is I don't know how to write type hints in the more generic shapes. Want helps. Of course, current type signature probably is usable for most cases?

It will be greatly appreciated if anyone provide suggestions to improve it!

Regards

### DIFF
--- a/docs/reference/array/groupConsecutiveBy.md
+++ b/docs/reference/array/groupConsecutiveBy.md
@@ -1,0 +1,51 @@
+# groupConsecutiveBy
+
+Groups the consecutive elements (like discriminated unions, but not limit only discriminated unions)
+of an array based on a provided key-generating function.
+
+This function takes an array and a function that generates a key from each element. It returns
+a array where the keys are the generated keys and the values are arrays of elements that share
+the same key.
+
+## Signature
+
+```typescript
+function groupConsecutiveBy<T, K extends PropertyKey>(arr: T[], getKeyFromItem: (item: T) => K): T[][];
+```
+
+### Parameters
+
+- `arr` (`T[]`): The array to group.
+- `getKeyFromItem` (`(item: T) => K`): A function that generates a key from an element.
+
+### Returns
+
+(`T[][]`) A array where sub element is an array of elements that
+share that key and are consecutive neighbors in the original array (also order preserved).
+
+## Examples
+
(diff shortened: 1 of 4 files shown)


raon0211:
Hello :) Thanks for the suggestion! I see the need for this function, but I'm still not sure how frequently it will be used by our users. We might consider defining it in each repository where it's needed. What do you think?

lqhuang:
@raon0211 Thanks for reply! Never mind, it's fine to close. I totally understand this function doesn't include in `loadash` either. I just thought `es-toolkit` could have more usage cases than `lodash`.

Feel free to close 😄, or just keep it open until other people back this feature or provide more feedbacks.

PR #437, opened 2024-08-28
docs(camelCase): Fix JSDoc for `camelCase`

kingston:
The JSDoc description for camel case describes snake case instead of camel case. This fixes it to be the correct definition.


raon0211 (approved):
Thanks for catching this!

PR #443, opened 2024-08-29
refactor(tail): Improve code readability in `tail`

mass2527:
this change not only improve code readability but also improve performance in large array.

i ran this bench to check `tail` in large array

```ts
import { tail as tailToolkit } from 'es-toolkit';
import { tail as tailLodash } from 'lodash';

// Helper function to generate a large array
function generateLargeArray(size) {
  return Array.from({ length: size }, (_, index) => index);
}

describe('tail', () => {
  const largeArray = generateLargeArray(1000000);

  bench('es-toolkit/tail', () => {
    tailToolkit(largeArray);
  });

  bench('lodash/tail', () => {
    tailLodash(largeArray);
  });
});
```

updated code is more faster than original code.

### before

스크린샷 2024-08-29 오후 10 02 55

### after

스크린샷 2024-08-29 오후 10 02 42

### DIFF
--- a/src/array/tail.ts
+++ b/src/array/tail.ts
@@ -68,13 +68,9 @@ export function tail<T, U>(arr: readonly [T, ...U[]]): U[];
 export function tail<T>(arr: readonly T[]): T[];
 
 export function tail<T>(arr: readonly T[]): T[] {
-  const len = arr.length;
-  if (len <= 1) {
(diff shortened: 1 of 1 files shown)


raon0211 (approved):
Thanks! Wondering why we implemented this in the original version..

PR #444, opened 2024-08-29
feat(initial): Support readonly array in `initial`

mass2527:
It would be great if `initial` support readonly array.


raon0211 (approved):
Thanks!

PR #445, opened 2024-08-29
feat(take): Add default `count` parameter to `take`

mass2527:
I think `take` function should also have default `count` parameter like [`takeRight`](https://github.com/toss/es-toolkit/blob/main/src/array/takeRight.ts)

lodash [`take`](https://lodash.com/docs/4.17.15#take) and [`takeRight`](https://lodash.com/docs/4.17.15#takeRight) both have default parameter.

### DIFF
--- a/src/array/take.ts
+++ b/src/array/take.ts
@@ -5,7 +5,7 @@
  * @template T - Type of elements in the input array.
  *
  * @param {T[]} arr - The array to take elements from.
- * @param {number} count - The number of elements to take.
+ * @param {number} [count=1] - The number of elements to take.
  * @returns {T[]} A new array containing the first `count` elements from `arr`.
  *
  * @example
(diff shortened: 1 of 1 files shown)


raon0211:
Actually, if `takeRight` has a default parameter, it is not what we intended. We would like our users to provide explicit arguments to our functions.

mass2527:
@raon0211 
Thanks for reviewing.

I assumed that the es-toolkit decided to add a default parameter, similar to lodash, since [takeRight](https://github.com/toss/es-toolkit/blob/main/src/array/takeRight.ts) already had one. If this wasn't intentional, we might consider removing the default parameter from `takeRight`.

raon0211:
Yes, that's right. We might remove the default parameter from `takeRight` -- that was by accident.

mass2527:
then i will close this PR and make another pull request that remove default parameter from `takeRight`

PR #453, opened 2024-08-31
fix(takeRight): Remove default parameter from `takeRight`

mass2527:
[As we discussed](https://github.com/toss/es-toolkit/pull/445#issuecomment-2322771234), I removed the unintended default parameter from takeRight.


mass2527:
Since this change can bring breaking change to library user, i will close this PR.

PR #454, opened 2024-08-31
chore(docs): sort sidebar items

dayongkr:
# Description

Screenshot 2024-08-31 at 2 43 59 PM

Currently, es-toolkit docs aren't sorted, I think that it makes some confusions.

So I fixed that it will sort in build time.


raon0211:
Thanks for your contribution. Maintaining docs is a big thing right now, I think we should automate the doc generation in the long run...

raon0211 (approved):

dayongkr:
> Thanks for your contribution.
> 
> Maintaining docs is a burden right now, I think we should automate the document generation in the long run...

I completely agree that we need automation. When I have some ideas, I'll share them.

PR #455, opened 2024-08-31
fix(negate): Fix parameter type of `negate`

mass2527:
Before the change, I encountered a type error when using the negate function with a function that has parameters.

스크린샷 2024-08-31 오후 3 04 48

I have updated the parameter type of negate so that it now supports functions with parameters.


raon0211 (approved):
Thanks! Technically it should be `never`, but I guess `any` is more readable..

PR #456, opened 2024-08-31
fix(delay): Ensure event listener is removed after timeout

mass2527:
Currently, if a signal is passed but not aborted, the event listener will not be removed, since the event listener with the [once](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#once) option is only automatically removed when it is invoked.

Therefore, I have updated the code to explicitly remove the event listener when a timeout occurs.


raon0211 (approved):
Seems that we need this. Thanks!

PR #463, opened 2024-08-31
feat(fromEntries):implement fromEntries

knott11:
微信图片_20240831225940


src/compat/object/fromEntries.ts
```diff
@@ -0,0 +1,27 @@
...
+ * const object = fromEntries(entries);
+ * console.log(object);
+ * // Output:
+ * // {
+ * //   a: 1,
+ * //   b: 2
+ * // }
+ */
+
+export function fromEntries<T extends PropertyKey, U>(
```
raon0211, line 19: Oh! Why not we just name this as `fromPairs` as in lodash, since it is for compatibility purposes?

knott11: I think the "fromPairs" currently can be considered as part of the JavaScript syntax.but I can change the name if it's not standard.

PR #465, opened 2024-09-01
feat(dropWhile): Add `dropWhile` in compat

mass2527:
I added a dropWhile function that is compatible with lodash.


raon0211 (approved):
Thanks for your great work!

raon0211:
Let me add some more compatibility with lodash, since we need perfect compatibility when added in `es-toolkit/compat`.

raon0211 (approved):

PR #467, opened 2024-09-01
test: Remove duplicated test code

mass2527:
I have removed the duplicate test code for `initial` and `without` that already exists in the `src/compat/array` folder.


raon0211 (commented):
Actually the test cases in `compat` is to make sure our `compat` functions work 100% identically with `lodash`, so duplicated test code is by purpose.

mass2527:
You're right. However, both "initial" and "without" have two test files in the `compat` folder: one in `compat/` and the other in `compat/array`. So i deleted files in `compat/`.

Is this also intentional?

raon0211:
This is intentional to keep track of whether we have compatibility with lodash.

mass2527:
Thanks for the explanation. I'll close this PR.

PR #469, opened 2024-09-01
feat(pad): add `pad` function

### LINKED ISSUE #386: Support for `pad`
The library already provides the [padStart](https://github.com/toss/es-toolkit/issues/280) and [padEnd ](https://github.com/toss/es-toolkit/issues/332)functions, so using them, we could support the [pad](https://lodash.com/docs/4.17.15#pad) function as well! 
However, unlike `padStart` and `padEnd`, `pad` does not part of the native API. Can I work on this? If so, does it belong to the `compat` or `string` module?

pkovzz:
### Solution

This pull request introduces a [pad](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html) function that pads a string on both the left and right sides if it is shorter than the specified length. The padding characters are truncated if they cannot be evenly divided by the length.

### Benchmark

Screenshot 2024-09-01 at 21 21 44

Resolves #386

### DIFF
--- a/benchmarks/performance/pad.bench.ts
+++ b/benchmarks/performance/pad.bench.ts
@@ -1,0 +1,15 @@
+import { bench, describe } from 'vitest';
+import { pad as padStartToolkit } from 'es-toolkit';
+import { pad as padStartLodash } from 'lodash';
+
+describe('pad', () => {
+  bench('es-toolkit/pad', () => {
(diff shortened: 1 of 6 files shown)


raon0211 (approved):
Thanks!

PR #470, opened 2024-09-02
chore: remove types of jsdoc

dayongkr:
# Description

We define types using TypeScript, so it doesn't seem necessary to specify types with JSDoc as well. Specifying both can lead to mistakes where different types are inadvertently documented, and it's quite cumbersome to update the JSDoc every time the TypeScript types are changed.

Moreover, even if we extract JSDoc to automate documentation, it’s pointless because JSDoc doesn’t fully support all of TypeScript’s types.

> It seems more accurate to use a TypeScript parser to extract types instead.

Screenshot 2024-09-02 at 11 15 28 AM

Additionally, even if we specify types in JSDoc, tools like VS Code only display type information based on TypeScript, just as described above.


dayongkr:
@raon0211 
If this PR is not a good idea, then It's ok closing this PR.

PR #472, opened 2024-09-02
docs: automate the English documentation

dayongkr:
# Description 

I’ve implemented an automation feature using VitePress’s `dynamic routes` and `comment-parser` with `@typescript-eslint/parser` to generate documentation(only `reference/array`).

The approach is as follows:

1. Dynamic routes and Node.js `fs` module are used to generate paths for each function (`reference/array/[function].md`, `reference/array/[function].paths.mts` are for Dynamic routes).
2. The source files of each function are parsed with `comment-parser` to extract JSDoc comments (e.g., function descriptions, parameter details) and with `@typescript-eslint/parser` to retrieve Function signatures and types.
3. The extracted data is then used to generate content strings through abstracted functions.

This setup allows for the English documentation to be created by simply adding a folder and two files for `Dynamic routes` per function category.

Before proceeding further, I’ve opened a PR to get feedback on the current implementation way. I would appreciate your feedback.

## Rough flow chart

Flowchart drawio (1)


dayongkr:
First of all, thank you for your feedback. Here are my thoughts in response:

1. I chose dynamic routes for full automation, as using a draft-based approach would still require modifications if the type changes. However, I completely agree that the draft approach would be more adaptable. I also could find many edge cases when I implemented this.
2. Both cases you mentioned are fully supported at now. Please refer to the images below for reference. Additionally, it also supports scenarios with multiple jsDOCs.

Screenshot 2024-09-04 at 10 45 29 AM Screenshot 2024-09-04 at 10 46 07 AM

I briefly looked into `@deno/doc`, and it seems quite powerful. However, it doesn’t provide support for the required scope, meaning that I would need to generate the types manually through AST. This adds a lot of branching and code, which raises concerns regarding maintainability and stability. But it still looks great!

By the way, the development process itself was fun. So, I'm totally fine with closing this PR, merging [this codes](https://github.com/toss/es-toolkit/blob/format/.scripts/docs/generate-docs.mts) into the main branch, and then finding areas for improvement later on!

Thanks!

dayongkr:
This issue is resolved by a9e7edde14e6f2b5a1e61262344de366879a9aee.

So I close this PR!

PR #480, opened 2024-09-07
feat(isArrayLikeObject): Add `isArrayLikeObject` in compat

### LINKED ISSUE #479: Support for `isArrayLikeObject`(compat)
I will be working on adding the `isArrayLikeObject` function to the compatibility layer.

mass2527:
resolve #479

I have added `isArrayLikeObject` in compat. To clarify the difference between `isArrayLike` (which is described as 'checks if a value is an array-like object') and `isArrayLikeObject`, I have defined `isArrayLikeObject` as checking if a value is a non-primitive, array-like object. Please feel free to modify the documentation if you think a better explanation exists.


raon0211 (approved):
The documentation looks nice. Thanks for your contribution!

PR #482, opened 2024-09-07
feat(sortBy): add `compat/sortBy` and fix `compat/orderBy`

dayongkr:
# Description

I added `compat/sortBy` that just calling a `compat/orderBy` function. So I have to change a `compat/orderBy`.

Changes:

- `Symbol`, `null`, `undefined`, and `NaN` should always be placed at the end, so I assigned priorities to values. (Lodash implemented this using [complex conditions](https://github.com/lodash/lodash/blob/ddfd9b11a0126db2302cb70ec9973b66baec0975/lodash.js#L4600), but I improved readability and performance by using a priority function).
  - `Array.prototype.sort` always places `undefined` at the end, but Lodash needed `NaN` to appear after `undefined`, so I wrapped the values in an object to sort them, then mapped them back to return the original values.
- Additionally, some conditions were reordered or added to pass test cases.
- I made it so that localeCompare() is used for strings.

According to benchmarks, the performance is similar to Lodash, but since Lodash’s performance is due to memoization, I think this is a meaningful improvement.

### `compat/sortBy`
Screenshot 2024-09-07 at 5 28 10 PM

### `compat/orderBy`

Screenshot 2024-09-07 at 5 20 37 PM


raon0211 (approved):
Thanks for your work!

PR #489, opened 2024-09-08
feat(attempt): implement attempt

D-Sketon:
benchmark
image
I'm not sure if `attempt` should be under `es-toolkit/function` or not.


raon0211 (approved):
Thanks for your contribution! Meanwhile, I guess this function should be added to `es-toolkit/compat`, since I guess there is a better interface for this..

PR #491, opened 2024-09-08
feat(conforms/conformsTo): implement conforms&conformsTo

D-Sketon:
benchmark
image
I'm not sure this function should be implemented entirely in compat


raon0211 (approved):
Thanks for your help!

raon0211:
I am not sure either, I guess it is best if we just write `obj => obj.x > 1 && obj.y < 2` instead of `conforms({ x: val => val > 1, y: val => val < 2 })`...

raon0211:
Let me add this in the compat library, and see if other people asks for it to add in our main library.

PR #492, opened 2024-09-08
fix(benchmarks): update concat's benchmarks

moonheekim0118:
fixed the incorrect benchmark for concat

# Benchmarks
스크린샷 2024-09-08 오후 9 39 12


raon0211 (approved):
Thanks!

PR #493, opened 2024-09-08
fix(countBy): add missing example in jsdoc

jsparkdev:
Added missing example in jsdoc of `countBy` function.

### DIFF
--- a/src/array/countBy.ts
+++ b/src/array/countBy.ts
@@ -14,6 +14,16 @@
  * @param {(item: T) => K} mapper - The transformation function that maps each item to a key.
  * @returns {Record<K, number>} An object containing the transformed items as keys and the
  * counts as values.
+ *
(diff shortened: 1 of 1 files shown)


src/array/countBy.ts
```diff
@@ -14,6 +14,16 @@
  * @param {(item: T) => K} mapper - The transformation function that maps each item to a key.
  * @returns {Record<K, number>} An object containing the transformed items as keys and the
  * counts as values.
+ *
+ * @example
+ * const array = ['a', 'b', 'c', 'a', 'b', 'a'];
+ * const result = countBy(array, String);
```
raon0211, line 20:
```suggestion
 * const result = countBy(array, x => x);
```

raon0211 (approved):
Thanks!

PR #494, opened 2024-09-08
fix(benchmark): add missing bench

D-Sketon:
image


raon0211 (approved):
Thanks for your great help!

PR #496, opened 2024-09-09
feat(clone): enhance clone function to handle various data types 

dayongkr:
# Description 

Changes:

- It supports `DataView`, `Error`, `TypedArray`, `File`, `ArrayBuffer`, `Blob`, `Buffer`, `SharedArrayBuffer`.
- I simplified the codes
  - Using `prototype.constructor()` instead of specified constructors  like `new Date()`(Also it's good for the exteneded Error instance.)
  - Using `obj` instead of property value like `obj.gettime()`, `obj.source`
  - Removing `isPrimitive()` that is a duplicated function.
- I fixed that we forget to copy `lastIndex` property in a `RegExp` object.

## Benchmark

Screenshot 2024-09-10 at 1 17 47 PM

### DIFF
--- a/src/object/clone.spec.ts
+++ b/src/object/clone.spec.ts
@@ -160,6 +160,10 @@ describe('clone', () => {
   });
 
   it('should clone File', () => {
+    if (typeof File === 'undefined') {
+      return;
+    }
+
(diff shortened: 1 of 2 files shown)


raon0211 (approved):
Looks nice. Thanks!

PR #501, opened 2024-09-09
feat(compat): Add `isObject` function 

WISUNGWON:
benchmark

image

There is an isObjectLike function, but not an isObject, so I implemented it.

I would appreciate any feedback and reviews.

### DIFF
--- a/src/compat/predicate/isObject.ts
+++ b/src/compat/predicate/isObject.ts
@@ -1,0 +1,29 @@
+/**
+ * Checks if the given value is the language type of Object.
+ * https://262.ecma-international.org/7.0/#sec-ecmascript-language-types
+ *
+ * This function tests whether the provided value is an object or not.
+ * It returns `true` if the value is an object, and `false` otherwise.
+ *
+ * This function can also serve as a type predicate in TypeScript, narrowing the type of the argument to an object value.
+ *
+ * @param {unknown} value - The value to check if it is an object.
+ * @returns {value is object | Function} `true` if the value is the language type of Object, `false` otherwise.
+ *
+ * @example
+ * const value1 = {};
+ * const value2 = [1, 2, 3];
+ * const value3 = () => {};
+ * const value4 = null;
+ *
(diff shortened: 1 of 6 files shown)


src/compat/predicate/isObject.ts
```diff
@@ -0,0 +1,28 @@
...
+ * const value3 = () => {};
+ * const value4 = null;
+ *
+ * console.log(isArray(value1)); // true
+ * console.log(isArray(value2)); // true
+ * console.log(isArray(value3)); // true
+ * console.log(isArray(value4)); // false
+ */
+
+export function isObject(value: unknown): value is object | ((...args: any[]) => unknown) {
```
raon0211, line 25: `object` is a supertype of `(...args: any[]) => unknown`; isn't `value is object` enough?

raon0211:
Thanks for your pull request!

src/compat/predicate/isObject.ts
WISUNGWON: `value is object` enough! thanks for your review

WISUNGWON: If the return type is object, Typescript infers that the value is of type ‘Function’ in the `if (typeof value === 'function')` statement. so that I'm wondering if it's better to specify Function as the return type.

```
function isObject(value: unknown): value is object {
    const valueType = typeof value;
    return valueType !== null && (valueType === 'object' || valueType === 'function');
}

const value: unknown = () => {}

if (isObject(value)) {
    console.log(typeof value) // object
    if (typeof value === 'function') {
        console.log(typeof value) // Function
    } else {
        console.log(typeof value) // object
    }
} else {
    console.log(value) // unknown
}
```

raon0211: I understand that `typeof` returns different values for functions and objects; but they are all `object`s in TypeScript. So I think in this case `object` might be enough.

raon0211 (approved):
Thanks for your contribution!

PR #504, opened 2024-09-09
fix(negate): add missing example

jsparkdev:
Added missing example in jsdoc of `negate` function.


raon0211 (approved):
Thanks for your help!

PR #507, opened 2024-09-10
feat(camelCase): implement `camelCase` in compat layer

dayongkr:
# Description 

I added `camelCase` in compat layer.
And I add a `normalizeForCase` internal function, because the other case functions also need to remove constraction apostrophes and handle case that the `str` argument is a `object`.

## Benchmarks

Screenshot 2024-09-10 at 5 38 35 PM

related #508


raon0211 (approved):
Thanks for your help!

PR #510, opened 2024-09-10
feat(compat): implement ceil/floor/round

D-Sketon:
benchmark
image
Since the document for "round" already exists, I don't know how to handle it.
In addition, to ensure accuracy, I have added additional test cases. However, since the implementation is different from that of lodash, I still suspect that there may be problems.


raon0211 (approved):
The implementation and test case seems nice, we might go for it and seek feedback..

PR #513, opened 2024-09-11
perf(deburr): improve performance

D-Sketon:
before
image
after
image

**However, if the input string does not have any characters that need to be deburr'd, the performance of the lodash implementation improves dramatically in long words case. Is this acceptable?**
image


raon0211 (approved):
Thanks! Is using `Map` more performant than `object`s?

PR #515, opened 2024-09-12
feat(capitalize): Add lodash test case

youznn:
# Description

I added [lodash test case](https://github.com/lodash/lodash/blob/main/test/capitalize.spec.js) for capitalize.


raon0211 (approved):
Thanks for your awesome work!

PR #516, opened 2024-09-12
chore(test): add script for transforming lodash test codes

dayongkr:
# Description

I added codemod for transform lodash test code based on `jscodeshift`.

It also format transformed source codes by `prettier`.

## Usage

1. Choose the test code in [here](https://github.com/lodash/lodash/tree/main/test) and copy it.
2. Create a test file such as `src/compat/array/at.spec.ts`, then paste the code into the file.
3. Run the following command in the terminal to execute the transform:
``` bash
yarn transform "./src/compat/array/at.spec.ts"
```

## Supported cases

- Import Statements:
  - Automatically adds `import { describe, it, expect } from 'vitest';` at the top of the file.
  - Removes import statements from the lodash library like `import lodashStable from lodash`.
  - Changes tools imported from `./utils` to `../internal/` and splits them accordingly. (e.g. `import {stubTrue, stubFalse} from './utils';` => `import {stubTrue} from '../internal/stubTrue'; import {stubFalse} from '../internal/stubFalse';`
- Handling lodashStable Methods:
  - Replaces methods like each and map with the corresponding `Array.prototype` methods. (e.g. `lodashStable.map([1,2], stubTrue)` => `[1,2].map(stubTrue)`
  - Other usages of lodashStable are removed for now. (e.g. `lodashStable.constant()` => `cosntant()`
- Syntax Adjustments (via regex):
  - Removes [expressions that delete local variables](https://github.com/lodash/lodash/blob/6a2cc1dfcf7634fea70d1bc5bd22db453df67b42/test/curryRight.spec.js#L67).
  - Fixes [broken expressions](https://github.com/lodash/lodash/blob/6a2cc1dfcf7634fea70d1bc5bd22db453df67b42/test/clone-methods.spec.js#L420) within `.toEqual` statements.

## Additional

I think that we have to add descriptions in [here](https://github.com/toss/es-toolkit/issues/235), after merging it.

### DIFF
--- a/.scripts/tests/_internal/formatter/brokenSyntax.ts
+++ b/.scripts/tests/_internal/formatter/brokenSyntax.ts
@@ -1,0 +1,21 @@
+export function formatBrokenSyntax(source: string): string {
+  // Fix broken syntax
+  const brokenMatched = source.match(/(?<=,)[\s\d\w?]+\[.+\).toEqual\((?!\[).+\]\);/g);
+  if (brokenMatched != null) {
+    for (const match of brokenMatched) {
+      const splited = match.split(').toEqual(');
+      const formatted = `).toEqual(${splited.join(', ')}`;
+      source = source.replace(match, formatted);
+    }
+  }
+
+  // Remove deleting local variable
+  const deleteMatched = source.match(/delete [\w\d]+;/g);
+  if (deleteMatched != null) {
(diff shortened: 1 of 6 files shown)


raon0211 (approved):
Thanks! Let me give it a try.

PR #517, opened 2024-09-12
feat(string): add `snakeCase`, `startCase`, `lowerCase`, `kebabCase` in compat layer

### LINKED ISSUE #508: Support for camelCase, kebabCase, lowerCase, snakeCase, startCase in compat layer
After merging #507, I am going to add rests.

Thanks!

dayongkr:
# Description

I added these functions in compat layer

### `snakeCase`

Screenshot 2024-09-12 at 10 46 31 PM

### `startCase`

Screenshot 2024-09-12 at 10 44 22 PM

### `lowerCase`

Screenshot 2024-09-12 at 10 45 56 PM

### `kebabCase`

Screenshot 2024-09-12 at 10 46 55 PM

close #508 

## Additional

I found a bug in `startCase` that is also in lodash.(you can see a related comment in [here](https://github.com/lodash/lodash/pull/908#issuecomment-72353907).

> Bug: `startCase` has to convert `FOO BAR` to `Foo Bar`, but it converts to `FOO BAR`.

I think that we have to fix this bug in `es-toolkit`


src/compat/_internal/normalizeForCase.ts
```diff
@@ -1,6 +1,6 @@
 export function normalizeForCase(str: string | object): string {
   if (typeof str === 'object') {
-    str = str.toString();
+    str = String(str);
```
raon0211, line 3: Does this improve performance?

D-Sketon: in my computer it becomes slower

dayongkr: > Does this improve performance?

I changed this code for the case that  `str` is `null` or `undefined`. Before codes throws error in this case.

But we already guided that don't insert `null` and `undefiend` and I also founded this changes makes 1.75x slower. 

So I reverted! Thanks

dayongkr: But Lodash don't throw error in that case. So I think `String(str)` is better way, because this function is for compatibility.

D-Sketon: Have you tried template strings? Is it faster?

dayongkr: > Have you tried template strings? Is it faster?

Screenshot 2024-09-12 at 10 44 22 PM

it seems slow like `String()`

dayongkr: Template literals also use complex and safe `toString`. So I think that above result make sense.

- https://tc39.es/ecma262/multipage/abstract-operations.html#sec-tostring
- https://tc39.es/ecma262/multipage/ecmascript-language-expressions.html#sec-template-literals-runtime-semantics-evaluation

raon0211 (approved):

src/compat/_internal/normalizeForCase.ts
raon0211: Why not we just add a null and undefined check at the top?

dayongkr: Yes we can do that, but I will compare with Lodash source codes for more compatible!

dayongkr: I think that we have to implement `estoolkit/compat/toString` before merging this. So i'm going to implement it at first!

raon0211: Okay!

PR #518, opened 2024-09-12
perf(has/get): improve performance

D-Sketon:
before
image

after
image


raon0211 (approved):
Thanks! Seems that changes like this can make a huge difference.

PR #519, opened 2024-09-12
fix(isNotNil): simplify `isNotNil` and add testcases

dayongkr:
# Description

I just simplify `isNotNil` code.

And I also fix a mistake in `isBoolean.spec.ts`.


raon0211 (approved):
Thanks!

PR #520, opened 2024-09-13
fix : wrong docs in isObject docs

DonghyunKim98:
All example code that should have been written as `isObject` has been modified to be written as `isArray`.


raon0211 (approved):
Good catch! Thanks for your help!

PR #521, opened 2024-09-13
feat(toString): add `toString` in compat layer

dayongkr:
# Description

I added `toString` in compat layer. It is just simple function, so it doesn't have Improved performance. 

I focused on modernized codes.

## Benchmarks

Screenshot 2024-09-13 at 5 15 18 PM

### DIFF
--- a/src/compat/string/toString.ts
+++ b/src/compat/string/toString.ts
@@ -1,0 +1,32 @@
+/**
+ *  Converts value to a string.
+ *  An empty string is returned for `null` and `undefined` values.
+ *  The sign of `-0` is preserved.
+ *
+ * @param {unknown} value - The value to convert.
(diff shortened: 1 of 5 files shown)


src/compat/string/toString.ts
```diff
@@ -0,0 +1,32 @@
+/**
+ *  Converts value to a string.
+ *  An empty string is returned for `null` and `undefined` values.
+ *  The sign of `-0` is preserved.
```
raon0211, line 4:
```suggestion
 * Converts `value` to a string.
 *
 * An empty string is returned for `null` and `undefined` values.
 * The sign of `-0` is preserved.
```

raon0211 (approved):

PR #522, opened 2024-09-13
fix(string): coerce to string for cases methods

dayongkr:
# Description 

As I mentioned [before](https://github.com/toss/es-toolkit/pull/517#discussion_r1758344272), I coerce to string by `es-toolkit/compat/toString`. And I also compare between`str.toString` and `toString(str)`, there are not significantly difference of performance.

### Before (`str.toString`)
Screenshot 2024-09-13 at 5 49 12 PM

### After (`toString(str)`)
Screenshot 2024-09-13 at 5 49 33 PM


src/compat/_internal/normalizeForCase.ts
```diff
@@ -1,9 +1,11 @@
-export function normalizeForCase(str: string | object): string {
-  if (typeof str === 'object') {
-    str = str.toString();
+import { toString } from '../string/toString';
```
raon0211, line 1:
```suggestion
import { toString } from '../util/toString';
```

raon0211 (approved):
Thanks!

PR #530, opened 2024-09-14
docs: replace `pnpm install` with `pnpm add`

jsparkdev:
The pnpm official documentation recommends using the `add` command instead of `install` when installing new packages as follows:

- https://pnpm.io/cli/add
- https://pnpm.io/cli/install

Currently it works the same without any differences, but the document clearly distinguishes these two commands.

This means that you have the flexibility to adapt to changes, such as changing or removing commands in future versions.


raon0211 (approved):
Thanks!

Issue #531, opened 2024-09-14
`drop` - Compatibility with lodash original functions


DonghyunKim98:
### In lodash

```typescript
function drop(array, n = 1) {
    const length = array == null ? 0 : array.length;
    return length ? slice(array, n < 0 ? 0 : toInteger(n), length) : [];
}
```

The code is written assuming that null enters array. However, in drop.ts of es-toolkit, the type of array is written as `readonly T[]`, so null cannot be entered at all.

Of course, the code below is not intuitive, but shouldn't exception handling be done for compatibility?
```
drop(null,2); // expected []
```

raon0211:
We might add these kind of behavior in our compatibility library. Please refer to functions like [orderBy](https://github.com/toss/es-toolkit/blob/main/src/compat/array/orderBy.ts#L41).

PR #532, opened 2024-09-15
feat(drop): Added compatibility for `drop`

DonghyunKim98:
|`drop.spec.ts`|
|---------------|
||

As mentioned in  Issue #531, which I opened, I wrote code to add compatibility to `drop.ts`. I also wrote additional test code and uploaded screenshots showing that it works properly.


raon0211 (approved):
Thanks!

PR #533, opened 2024-09-15
feat(curry): enhance type

Plumbiu:
I noticed that the type tag of the curry function is enumerated. When the number of parameters exceeds a certain value, the type will be any.

There are three cases for currying function types:

1. no parameter, return `() => R`
2. one parameter, return `(p: P) => R`
3. More than one parameter, return `(p: P) => Curried<Rest, P>`

```ts
type Curried<A extends any[], R> = A extends []
  // If the function parameter length is 0, return () => R
  ? () => R
  : A extends [infer P]
    // If the function parameter length is 1, return (p: P) => R
    ? (p: P) => R
    : A extends [infer P, ...infer Rest]
      // If the function has more than 1 parameter, return (p: P) => Curried<Rest, R>
      ? (p: P) => Curried<Rest, R>
      : never;
export function curry<A extends any[], R>(func: (...args: A) => R): Curried<A, R> {
  if (func.length === 0 || func.length === 1) {
    return func as Curried<A, R>;
  }

  return function (arg) {
    return makeCurry(func, func.length, [arg]);
  } as Curried<A, R>;
}
```

### DIFF
--- a/src/function/curry.ts
+++ b/src/function/curry.ts
@@ -2,119 +2,8 @@
  * Curries a function, allowing it to be called with a single argument at a time and returning a new function that takes the next argument.
  * This process continues until all arguments have been provided, at which point the original function is called with all accumulated arguments.
  *
- * @param {() => R} func - The function to curry.
- * @returns {() => R} A curried function.
- *
- * @example
- * function noArgFunc() {
- *   return 42;
- * }
- * const curriedNoArgFunc = curry(noArgFunc);
- * console.log(curriedNoArgFunc()); // 42
- */
-export function curry<R>(func: () => R): () => R;
-
-/**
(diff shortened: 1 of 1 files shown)


raon0211:
Hello, thanks for your suggestion!

The design principle of `es-toolkit` is to [offer the simplest interface and implementation for 85% of major use cases](https://github.com/toss/es-toolkit/blob/main/.github/CONTRIBUTING.md#12-simplicity). We recognize that the type you suggested covers all scenarios, but we believe that currying functions with more than 5 parameters isn’t very common for most users. That’s why we decided to stick with the current type interface.

What do you think about this?

raon0211:
Closing the pull request for now, feel free to open the pull request again :)

PR #534, opened 2024-09-15
feat(initial): Add overloaded signatures for `initial`

mass2527:
I added function overloading signatures for `initial` similar to [`tail`](https://github.com/toss/es-toolkit/blob/main/src/array/tail.ts) and included corresponding examples to aid understanding.


raon0211 (approved):
Thanks!

PR #536, opened 2024-09-15
fix(spread): fix doc

D-Sketon:
Documentation in other languages may also need to be modified


raon0211 (approved):
Thanks for catching this!

PR #537, opened 2024-09-15
feat(string): implement `compat/upperCase` and fix types and jsdoc of case methods

dayongkr:
# Description

I added a `upperCase` in compat layer.

And I fixed function interface similar with `@types/lodash`.

## Benchmark (`upperCase`)

Screenshot 2024-09-15 at 6 41 20 PM


raon0211 (approved):
Thanks!

PR #538, opened 2024-09-15
feat(isNaN): add `isNaN` function

benzyminzy:
Hi there👋 I've implemented the isNaN function. 

I welcome any feedback for improvements. Thank you.

image

### DIFF
--- a/src/predicate/isNaN.ts
+++ b/src/predicate/isNaN.ts
@@ -1,0 +1,15 @@
+/**
+ * Checks if the value is NaN.
+ *
+ * @param {unknown} value - The value to check.
+ * @returns {value is typeof NaN} `true` if the value is NaN, `false` otherwise.
+ *
+ * @example
+ * isNaN(NaN); // true
+ * isNaN(0); // false
+ * isNaN('NaN'); // false
+ * isNaN(undefined); // false
+ */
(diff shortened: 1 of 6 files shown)


src/predicate/isNaN.ts
```diff
@@ -0,0 +1,15 @@
...
+ * @returns {value is typeof NaN} `true` if the value is NaN, `false` otherwise.
+ *
+ * @example
+ * isNaN(NaN); // true
+ * isNaN(0); // false
+ * isNaN('NaN'); // false
+ * isNaN(undefined); // false
+ */
+export function isNaN(value: unknown): value is typeof NaN {
+  return typeof value === 'number' && value !== value;
```
raon0211, line 14: What about we use `Number.isNaN` in this case? We should check the performance differences, though..

benzyminzy: I checked and found that Number.isNaN has better performance. 
I think it would be better to implement it the way you said.

I think I forgot to use the native function as I focused on utilizing the properties of NaN. 
Thank you for pointing this out!😀
image

benzyminzy:
@raon0211 Updated to use `Number.isNaN`. Thanks for the feedback!👍

raon0211 (approved):
Thanks!

PR #539, opened 2024-09-15
perf(zip): Optimize performance of `zip`

mass2527:
I've improved the performance of `zip` by defining the array size. The updated code is 1.41x faster than the previous version.

스크린샷 2024-09-15 오후 8 36 11


raon0211 (approved):
Thanks for your work!

PR #541, opened 2024-09-16
docs(curry): Update docs for `curry`

mass2527:
I've updated the documentation to fix inaccuracies and improve readability. I also removed sections that I think are unnecessary.

### DIFF
--- a/src/function/curry.ts
+++ b/src/function/curry.ts
@@ -51,18 +51,17 @@ export function curry<P1, P2, R>(func: (p1: P1, p2: P2) => R): (p1: P1) => (p2: 
  * Curries a function, allowing it to be called with a single argument at a time and returning a new function that takes the next argument.
  * This process continues until all arguments have been provided, at which point the original function is called with all accumulated arguments.
  *
- * @param {(p1: P1, p2: P2, p3: P3, p4: P4) => R} func - The function to curry.
- * @returns {(p1: P1) => (p2: P2) => (p3: P3) => (p4: P4) => R} A curried function.
+ * @param {(p1: P1, p2: P2, p3: P3) => R} func - The function to curry.
+ * @returns {(p1: P1) => (p2: P2) => (p3: P3) => R} A curried function.
  *
  * @example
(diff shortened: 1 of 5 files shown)


src/function/curry.ts
```diff
@@ -100,63 +99,15 @@ export function curry<P1, P2, P3, P4, R>(
...
- * // The parameter `a` should be given the value `10`.
- * const sum10 = curriedSum(10);
- *
- * // The parameter `b` should be given the value `15`.
- * const sum25 = sum10(15);
- *
- * // The parameter `c` should be given the value `5`. The function 'sum' has received all its arguments and will now return a value.
- * const result = sum25(5);
- */
-export function curry(func: (...args: any[]) => any): (...args: any[]) => any;
```
mass2527, line 135: I think if we don’t support more than 5 parameters, we don’t need to write an overload signature for it. I think not supporting it is better than supporting it with incorrect types. What do you think about this?

src/function/curry.ts
```diff
@@ -100,63 +99,15 @@ export function curry<P1, P2, P3, P4, R>(
...
- *
- * // The parameter `a` should be given the value `10`.
- * const sum10 = curriedSum(10);
- *
- * // The parameter `b` should be given the value `15`.
- * const sum25 = sum10(15);
- *
- * // The parameter `c` should be given the value `5`. The function 'sum' has received all its arguments and will now return a value.
- * const result = sum25(5);
- */
```
mass2527, line 159: Since the overloaded signatures already have good JSDoc, we can delete this jsdoc of function implementation and keep only the necessary sections.

src/function/curry.ts
raon0211: Actually I think we need interfaces for functions more than 5 parameters. The types are not precise, but at least it is usable. I think it is better to have a fallback.

mass2527: I've thought about this again and changed my mind. I agree with you—keeping this signature will be useful for flexibility.

Thanks for sharing your opinion!

raon0211 (approved):
Thanks for your suggestion!

PR #542, opened 2024-09-17
test(partial): add curry test

D-Sketon:
In fact, partial and partialright have passed lodash's test completely!


raon0211 (approved):
Thanks!

PR #543, opened 2024-09-17
feat(isNumber): add `isNumber` function

shan-mx:
This PR introduces isNumber function to check if the given value is a number.

image


src/predicate/isNumber.ts
```diff
@@ -0,0 +1,20 @@
...
+ * const value1 = 123;
+ * const value2 = 'abc';
+ * const value3 = true;
+ *
+ * console.log(isNumber(value1)); // true
+ * console.log(isNumber(value2)); // false
+ * console.log(isNumber(value3)); // false
+ */
+export function isNumber(value: unknown): value is number {
+  return typeof value === 'number' && !Number.isNaN(value);
```
raon0211, line 19: I am not sure if we should return `false` for `NaN` (`lodash` also returns `true` for `NaN`). Is there some background for checking the value is not `NaN`?

shan-mx: Forgot to notice that, will align this behavior same to lodash.

raon0211 (approved):
Thanks! Let me move this into our compatibility library, since we can easily check if a value is a number using `typeof` checks natively.

PR #545, opened 2024-09-17
fix(curry): Fix curry types and logic.

seungrodotlee:
current version of `curry` has issue with `optional parameters`.

```
it('should work with optional parameters', () => {
  const fn = (a: number, b: number, c?: number, d?: number) => [a, b, c, d];
  const curried = curry(fn);
  const expected = [1, 2, undefined, undefined];

  expect(curried(1)(2)).toEqual(expected);
});
```
image

The current version of `curry` ignores optional parameters and infers types. 
Therefore, when two argument values are passed, The curried function looks like return a value, but the actual behavior does not reflect this because in the built JS file, `fn.length` still remains 4. 

I believe this could be a significant issue. 
Thus, although it slightly compromises the conciseness of the function, it seems necessary to adopt a method that allows developers to explicitly specify the number of arguments they can input using arity.

```
const fn = (a: number, b: number, ...rest: number[]) => [a, b, ...rest];
const curriedA = curry(fn, 2); // receive 2 arguments
const curriedB = curry(fn) // receive 4 arguments

curriedA(1)(2) // Expect [1, 2, undefined, undefined]
curriedB(1)(2)()() // Expect [1, 2, undefined, undefined]
```

Additionally, the current version of `curry` only guarantees types for a maximum of 5 parameters. 
However, in real-world scenarios where the curry function is used, there may be situations that require more parameters.
Consequently, we will improve the typing to accommodate this.

Suggestion for enhance of types is optional and It would be great if it's simplified! **But, I think sync types with arity is required.**


raon0211:
Hi! Thanks for your suggestion. I understand that the current `curry` type doesn't handle optional arguments and is limited to 5 arguments. However, I have a few concerns:

- The new `CurryResult<F, A>` type is less intuitive compared to the original one, which means users would need to learn what the type signifies.
- The new types are much more complex to maintain, which goes against our [design principles for implementation simplicity](https://github.com/toss/es-toolkit/blob/main/.github/CONTRIBUTING.md#12-simplicity).

Also, I'm not convinced that supporting more than 5 arguments or currying functions with default parameters is a major use case. In cases like these, we can use lambda functions to create a curried version:

```tsx
function minus(a: number, b: number, c?: number, d?: number) {
   return a - b - (c ?? 1) - (d ?? 0);
}

function curriedMinus(a: number) {
  return function (b: number, c?: number, d?: number) {
    return minus(a, b, c, d);
  }
}
```

I'm not sure if `const curriedMinus = curry(minus, 4)` would be more intuitive and readable than the approach shown above.

seungrodotlee:
I understand and agree with importance of simplicity.

In that case, how about clearly stating in the documentation that using optional parameters is not supported, and that it is recommended to implement currying directly using lambda? This is because the current implementation does not allow for proper use when optional parameters are involved.

D-Sketon:
When optional parameters are used, we can suggest using `curry/compat`

seungrodotlee:
> When optional parameters are used, we can suggest using curry/compat

Oh, I missed that😅
Close this issue!

PR #546, opened 2024-09-18
chore(script): enhance transforming lodash test codes script

dayongkr:
# Description

I improved transforming lodash test codes script:

1. Change `delete localVariable` to comment, Because this expression is not allowed in js strict mode.
2. Enhance fixing a broken syntax like `expect(a, [1, 2).toEqual(3])`.
  a. Now, we can handle cases like `expect(func()`, `expect({a: {b: [1,2]}`, etc...
3. Change `assert.methods(a,b,c)` to `expect(a,c).toMethods(b)`.
  b. e.g. `assert.strictEqual(a,b,c)` to `expect(a, c).toBe(b)`
4. Change `lodashStable.methods` to `esToolit.methods`
5. Change `import method from '../src/method'` to `import { method } from '../index'`

## Test results

Screenshot 2024-09-18 at 1 17 46 PM

At now, We can transform all of the lodash testcode except 2 files.

I have automated the transformation of all lodash test codes as much as possible, but there are cases where the test code itself is inaccurate. Therefore, we shouldn’t rely on the tool too much. It would be better if it is only used as an auxiliary tool.

### DIFF
--- a/.scripts/tests/_internal/formatter/brokenSyntax.ts
+++ b/.scripts/tests/_internal/formatter/brokenSyntax.ts
@@ -1,22 +1,54 @@
+function fixExpectFunction(source: string): string {
+  const expectMatched = source.match(/expect(\(.+\)).toEqual.+\);/g);
+
+  if (!expectMatched) {
+    return source;
+  }
+
+  const splitMatched = expectMatched.map(match => match.split('.toEqual('));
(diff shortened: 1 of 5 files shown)


raon0211 (approved):
Looks really great. Thanks a lot!

PR #547, opened 2024-09-18
docs: update wrong headings and contents

jsparkdev:
- update untranslated headings (mainly for Korean documents)
- update incorrect heading levels.
- update untranslated or incorrect contents.
- improve formatting consistency.


raon0211 (approved):
Thanks for catching these! :)

PR #552, opened 2024-09-19
docs(range): Update docs for `range`

mass2527:
I've updated the JSDoc for `range` to describe only the corresponding overloaded function signatures.

### DIFF
--- a/src/math/range.ts
+++ b/src/math/range.ts
@@ -1,39 +1,21 @@
 /**
- * Returns an array of numbers from `start` to `end`, incrementing by `step`.
+ * Returns an array of numbers from `0` (inclusive) to `end` (exclusive), incrementing by `1`.
  *
- * If `step` is not provided, it defaults to `1`.
- *
- * @param {number} start - The starting number of the range (inclusive).
- * @param {number} [end] - The end number of the range (exclusive).
- * @param {number} [step] - The step value for the range. (default: 1)
- * @returns {number[]} An array of numbers from `start` to `end` with the specified `step`.
+ * @param {number} end - The end number of the range (exclusive).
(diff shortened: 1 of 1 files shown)


src/math/range.ts
```diff
@@ -56,24 +38,24 @@ export function range(start: number, end: number): number[];
...
+  * @throws {Error} Throws an error if the step value is not a non-zero integer.
+  *
+  * @example
+  * // Returns [0, 1, 2, 3]
+  * range(4);
+  *
+  * @example
+  * // Returns [0, -1, -2, -3]
+  * range(0, -4, -1);
+  */
```
raon0211, line 58:
```suggestion
 * Returns an array of numbers from `start` to `end`, incrementing by `step`.
 *
 * If only one argument is provided, it returns an array from `0` to `start`.
 *
 * @param {number} start - The starting number of the range (inclusive).
 * @param {number} [end] - The end number of the range (exclusive).
 * @param {number} [step] - The step value for the range. (default: 1)
 * @returns {number[]} An array of numbers from `start` to `end` with the specified `step`.
 * @throws {Error} Throws an error if the step value is not a non-zero integer.
 *
 * @example
 * // Returns [0, 1, 2, 3]
 * range(4);
 *
 * @example
 * // Returns [0, -1, -2, -3]
 * range(0, -4, -1);
 */
```

raon0211:
Restore the JSDoc comment at the end because it is actually used in Deno.

raon0211 (approved):
Thanks!

mass2527:
> Restored the JSDoc comment at the end, because it is actually used in Deno.

Thanks for explanation.
Now i can understand why it is necessary.

PR #553, opened 2024-09-19
docs: Delete wrong matched md file in references/math

VVSOGI:
maxBy utils belongs to the array category but, its explanation was included in docs/ko/reference/math, 
so I deleted it from there.


raon0211 (approved):
Thanks!

PR #554, opened 2024-09-19
docs: Add detail examples to docs in maxBy, minBy

VVSOGI:
I found the maxBy and minBy utils quite useful, but noticed there weren't many examples. 
I thought that adding more examples might encourage developers to use these functions more frequently, 
so I've included some additional ones.

If this approach is helpful, I'd be glad to provide similar examples for other utils as well. 
Please let me know if you think this would benefit the project. 👍


raon0211:
This looks nice. Could you add the examples to English, Japanese, and Simplified Chinese docs too?

VVSOGI:
> This looks nice. Could you add the examples to English, Japanese, and Simplified Chinese docs too?

Sure! I'll add it and PR it again.

raon0211 (approved):
Thanks!

PR #556, opened 2024-09-19
fix(merge): Handle edge-case for `merge`

mass2527:
I understood how the `merge` function works by testing Lodash's `merge` function. I then realized there was a difference between the two.

Although this is not `compat/merge`, the behavior should be consistent. Please let me know if the current behavior is intended.

### DIFF
--- a/src/object/merge.spec.ts
+++ b/src/object/merge.spec.ts
@@ -60,8 +60,9 @@ describe('merge', () => {
   });
 
   it('should handle merging arrays into non-array target values', () => {
-    const target = { a: 1, b: {} };
     const numbers = [1, 2, 3];
+    
+    const target = { a: 1, b: {} };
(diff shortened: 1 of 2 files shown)


raon0211 (approved):

rwe:
Looks like this should be applied to `mergeWith` as well, which still has identical logic to `merge` before this PR: https://github.com/toss/es-toolkit/blob/f1c7f923b1315bb05cb3a5e4070e6fa50d3127c5/src/object/mergeWith.ts#L74-L78

Additionally, I wonder if that type check should be done to at the top level of "source" and "target", rather than the `target[key] = …`, in order to keep the recursive behavior consistent with `merge(a, b)`. As it is, `merge(["1"], {a: 2})` behaves differently than `merge({x: ["1"]}, {x: {a: 2}}).x`, which is probably unexpected.

PR #563, opened 2024-09-20
feat(isJSONObject, isJSONValue, isJSONArray): Implement `isJSONObject`, `isJSONValue`, `isJSONArray`

knott11:
I will add 'isJSONString' in the next few days.


raon0211 (dismissed):
Thanks for your contribution! Refactored the implementation a bit so that it clearly matches the JSON spec.

PR #564, opened 2024-09-20
Add `fp` export

seungrodotlee:
Add `es-toolkit/fp` export to support functional programming and pipeline syntax.

```ts
// example

pipe(
  [1, 2, 3, 4, 5, 6, 7, 8, 9, 10],
  filter(isEven),
  map(double),
  join(", "),
) // Expect: "4, 8, 12, 16, 20"
```

### Scripts
- [x] Write a script to clone and convert existing functions into 'pipable' functions for pipelining.
- [ ] Enhance a script to clone and convert existing specs to fp-version specs.

### Codes
- [ ] Write basic functions for fp (ex. `pipe`, `map`, `filter` ...)
- [ ] Convert all available existing functions

### Documents
- [ ] Write documents about `es-toolkit/fp`


gwansikk:
Hello, I find this task interesting and believe it will bring high productivity and convenience. Are there any additional updates? I’d love to collaborate on this.

How about structuring the API interface as shown below? (This is just one suggestion.)

```ts
toolkit([1, 2, 3, 4, 5, 6, 7, 8, 9, 10]).pipe(
  filter(isEven),
  map(double),
  join(", "),
) // Expect: "4, 8, 12, 16, 20"
```

This structure holds several potential advantages

- Clear input, making the flow of data transformation intuitive.
- Flexible scalability, allowing for future expansion not only in pipe but also in other interfaces.
- Consistent API interface patterns that maintain uniformity.
- Enhanced TypeScript type inference for greater precision.

D-Sketon:
any progress on this thread?

seungrodotlee:
While I am trying to work as quickly as possible, but it is not a very high priority, so I am falling behind. I hope to be able to work on it this week. 🥲

seungrodotlee:
The syntax for the pipe function is planned as follows.

```typescript
function toString(value: unknown) {
  return `string:${value}`;
}

function toStringAsync(value: unknown) {
  return Promise.resolve(`string:${value}`);
}

function length(value: string) {
  return value.length;
}

const result = pipe(1, toString, length);
console.log(result); // 8

// Use pipe with async function
const asyncResult = await pipe(1, toStringAsync, length);
console.log(asyncResult); // 8

// Use pipe with curried function
const mapKeyResult = await pipe(
  { a: 1, b: 2 },
  mapKeys((value, key) => key + value)
);
console.log(mapKeyResult); // { a1: 1, b2: 2 }
```

- receive the initial value as first parameter
- receive functions that process the value from the second parameter onward
- can handle async functions

seungrodotlee:
image

bench result of `pipe`

sunrabbit123:
As far as I remember, this team was not interested in that work, but it is a work that I am interested in.

But is there a reason why I have to copy the implementation?
It also looks good to refer to the original function with curring to take the consistency of the implementation.


```ts
const pipeableCountBy = <T>(fn: (arg: T) => boolean) => (arr: Array<T>) => countBy(arr, fn);
pipe(
   [1, 2, 3, 4],
   pipeableCountBy(v => v % 2 == 0)
);
```

If you're not going to make it in consideration of curring from the beginning, like fxts, that seems to be the best option.


The advantage of p.spipe comes as a delay assessment.
That's the linq of dotnet, that's the fxts, that's the effect-ts.

seungrodotlee:
Maybe we don't need to copy original implementation even if there is a script for copying it... I agree with that idea.

But it would be better to keep current design that 'auto-currying' the function by number of received argument to avoid situations like the one below. 

```tsx
// BAD
import { omit } from "es-toolkit";
import { omit as pipableOmit } from "es-toolkit/fp";

const omitted = omit(obj, ['a', 'b']); // usage without pipe

// ...

pipe(
  obj,
  pipableOmit(['a', 'b'])
); // usage with pipe
```

```tsx
// GOOD
import { omit } from "es-toolkit/fp";

const omitted = omit(obj, ['a', 'b']); // usage without pipe

// ...

pipe(
  obj,
  omit(['a', 'b'])
); // usage with pipe
```

I gonna fix the implementation and script to  refer it, not copy the implementation of original one.

sunrabbit123:
I think that's a good idea, too.
The core of my story is to refer to the feature implementation to follow the original, but only to change the interface.

seungrodotlee:
## Move functions to `fp`

### array
[x] at
[x] chunk
[x] compact
[x] countBy
[x] difference
[ ] ~~differenceBy~~
[ ] ~~differenceWith~~
[ ] drop
[ ] dropRight
[ ] dropRightWhile
[ ] dropWhile
[ ] fill
[ ] flatMap
[ ] flatMapDeep
[ ] flatten
[ ] flattenDeep
[ ] forEachRight
[ ] groupBy
[ ] head
[ ] initial
[ ] intersection
[ ] intersectionBy
[ ] intersectionWith
[ ] isSubset
[ ] keyBy
[ ] last
[ ] maxBy
[ ] minBy
[ ] orderBy
[ ] partition
[ ] pullAt
[ ] sample
[ ] sampleSize
[ ] shuffle
[ ] sortBy
[ ] tail
[ ] take
[ ] takeRight
[ ] takeRightWhile
[ ] takeWhile
[ ] toFilled
[ ] union
[ ] unionBy
[ ] unionWith
[ ] uniq
[ ] uniqBy
[ ] uniqWith
[ ] unzip
[ ] unzipWith
[ ] without
[ ] xor
[ ] xorBy
[ ] xorWith
[ ] zip
[ ] zipObject
[ ] zipWith

### function
[ ] after
[ ] ary
[ ] before
[ ] curry
[ ] curryRight
[ ] debounce
[ ] flow
[ ] flowRight
[ ] identity
[ ] memoize
[ ] negate
[ ] noop
[ ] once
[ ] partial
[ ] partialRight
[ ] rest
[ ] spread
[ ] throttle
[ ] unary

### math
[ ] clamp
[ ] inRange
[ ] mean
[ ] meanBy
[ ] median
[ ] medianBy
[ ] random
[ ] randomInt
[ ] range
[ ] rangeRight
[ ] round
[ ] sum
[ ] sumBy

### object
[ ] clone
[ ] cloneDeep
[ ] flattenObject
[ ] invert
[ ] mapKeys
[ ] mapValues
[ ] merge
[ ] mergeWith
[ ] omit
[ ] omitBy
[ ] pick
[ ] pickBy
[ ] toMerged

### string
[ ] camelCase
[ ] capitalize
[ ] constantCase
[ ] deburr
[ ] escape
[ ] escapeRegExp
[ ] kebabCase
[ ] lowerCase
[ ] lowerFirst
[ ] pad
[ ] pascalCase
[ ] snakeCase
[ ] startCase
[ ] trim
[ ] trimEnd
[ ] trimStart
[ ] unescape
[ ] upperCase
[ ] upperFirst

### util
[ ] invariant

nabato:
Will `fp` package implement structural sharing based on object prototyping like `lodash/fp` does?

dayongkr:
I'm planning to close this PR since it was not updated it for a long time.

If you plan to resume work on it, feel free to reopen.

amrsalama:
Hi everyone, thanks a lot for your awesome work and dedication.
Is there any update regarding this? `pipe` and `/fp`?

PR #565, opened 2024-09-21
docs: fix doc

D-Sketon:


docs/zh_hans/reference/compat/array/castArray.md
```diff
@@ -11,7 +11,7 @@
 ## 签名
 
 ```typescript
-function castArray<T>(value?: T | T[]): T[];
+function castArray<T>(value?: T | readonly T[]): T[];
```
raon0211, line 14:
```suggestion
function castArray<T>(value?: T | T[]): T[];
```
Actually we are not printing `readonly` in our signature.

raon0211 (approved):
Thanks!

PR #566, opened 2024-09-21
perf(flatten): improve performance

D-Sketon:
before
{90138E23-BFCD-49D9-8548-20F66804BEA2}
after
{D71EC599-8FC8-41BA-A3E1-C8CC8B771C06}


raon0211 (approved):
Thanks! This is a common code pattern that we have to improve.. Wondering if we can catch this with eslint rules or something.

PR #568, opened 2024-09-21
feat(isJSONString):implement `isJSONString`

knott11:
I think it's more flexible to support both objects and arrays.


knott11:
> I think it's more flexible to support both objects and arrays.

maybe it's JSON object string and JSON array string.

PR #569, opened 2024-09-21
feat(utils): add `toNumber`, `toFinite`, `toInteger` in compat layer

### LINKED ISSUE #567: Support for `toFinite`, `toInteger`, `toNumber` in compat layer
These needs to implement `compat/includes`, so I am going to implementing them.

dayongkr:
# Description

I added these functions in `es-toolit/compat/utils`.

We we don't have `es-toolkit/utils`, so I didn't add these functions in `es-toolkit/utils`.

If these functions also need in `es-toolkit`, could you recommend which folder would be best to create it in?

###  `toNumber`

Screenshot 2024-09-21 at 4 48 32 PM

### `toFinite`

Screenshot 2024-09-21 at 4 48 49 PM

### `toInteger`

Screenshot 2024-09-21 at 4 48 19 PM

close #567


raon0211 (approved):
I think the current implementation is fine; if other people ask for `toFinite`, `toInteger`, or `toNumber` in our original library, let's add it.

PR #570, opened 2024-09-21
docs(fetchData): Improve example of `withTimeout`

mass2527:
I've updated the `withTimeout` example to improve understanding.


raon0211 (approved):
Seems that the new example is much more easier to understand. Thanks for your help!

PR #573, opened 2024-09-21
fix(predicate): enhance performance and simplify codes of `isNumber`, `isString`, `isSymbol`, `isRegExp`, `isBoolean`

dayongkr:
# Description

If Class like `String` exists, I think that we don't need to compare with string like`[Object string]`, when checking wrapped value like `Object('str')`.

Additionally, I changed the parameter name of `isBoolean` to value for consistency, and added optional types to both `isBoolean` and `isString` for compatibility with Lodash.

#### Before
Screenshot 2024-09-21 at 7 59 30 PM

#### After

Screenshot 2024-09-21 at 7 59 38 PM

---

#### Before

Screenshot 2024-09-21 at 8 00 31 PM

#### After

Screenshot 2024-09-21 at 8 00 58 PM

---

#### Before

Screenshot 2024-09-21 at 8 01 56 PM

#### After

Screenshot 2024-09-21 at 8 02 10 PM

---

#### Before

Screenshot 2024-09-21 at 8 03 18 PM

#### After

Screenshot 2024-09-21 at 8 04 11 PM

---

#### Before

Screenshot 2024-09-21 at 8 05 37 PM

#### After

Screenshot 2024-09-21 at 8 05 52 PM

### DIFF
--- a/benchmarks/performance/isBoolean.bench.ts
+++ b/benchmarks/performance/isBoolean.bench.ts
@@ -10,6 +10,8 @@ describe('isBoolean', () => {
     isBooleanToolkit(undefined);
     isBooleanToolkit('');
     isBooleanToolkit(123);
+    isBooleanToolkit({});
+    isBooleanToolkit(Boolean());
   });
 
   bench('es-toolkit/compat/isBoolean', () => {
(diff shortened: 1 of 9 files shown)


raon0211 (approved):
Seems that our tests are passing. Thanks for your help!

PR #579, opened 2024-09-22
feat(isError): append `isError`

hyesungoh:
Hi there 👍 

I append `isError`.

스크린샷 2024-09-22 오후 4 48 40


raon0211 (approved):
Thanks!

PR #581, opened 2024-09-22
feat(defer): implement defer

D-Sketon:
benchmark
{E96CF4C6-6EDD-4792-A880-F77C994030B2}


raon0211 (approved):
Thanks for your contribution!

PR #586, opened 2024-09-23
docs: Delete wrong matched docs in references/math in ja

VVSOGI:
I found same problems like #553 while adding examples to docs
No more files related to maxBy.md and minBy.md in references/math, 
only in each language docs of references/array.


raon0211 (approved):
Thanks!

PR #587, opened 2024-09-23
chore(lint): add lint plugin for using `for loop` instead of `for...of` over array

dayongkr:
# Description

I have developed a plugin that prohibits the use of `for...of` when iterating over arrays, due to performance issues that arise when arrays are accessed via iterators. The goal is to encourage the use of traditional for loops, which are generally more efficient in these scenarios.

> The idea for this plugin came from [this comment](https://github.com/toss/es-toolkit/pull/566#pullrequestreview-2321201991).

## The plugin prevents the following cases:

```javascript
const array = [1, 2, 3];
const object = { a: 1 };

for (const item of array) {} // Not allowed
for (const item of Object.entries(object)) {} // Not allowed

for (const item of new Set()) {} // Allowed
for (const item of new Map() {} // Allowed
```

Since TypeScript is used to check if the value next to `of` is an array, a reporting is triggered if it is.

Therefore, when using functions like `Object.entries()` or `Object.keys()`, which return arrays, a reporting will also be generated in these cases.

## Additional

You can seeing source of this plugin in [here](https://github.com/dayongkr/eslint-plugin-no-for-of-array).

> I found that having the ESLint plugin in this repository felt a bit unnatural, so I implemented a separate repository.


dayongkr:
The test file `src/compat/function/debounce.spec.ts` intermittently fails. Since it has been failing after merging from main branch, it seems to be an issue with the main branch.

raon0211 (approved):
This is great! Thanks for your contribution!

raon0211:
> The test file src/compat/function/debounce.spec.ts intermittently fails. Since it has been failing after merging from main branch, it seems to be an issue with the main branch.

Seems that the current test is flaky. Let me check this.

PR #313, opened 2024-07-24
feat(camelCase/capitalize/kebabCase/lowerCase/snakeCase/startCase): improve return types

D-Sketon:
Even though the type passes the test, I can't guarantee that it covers all boundary cases

PR #591, opened 2024-09-23
fix(startCase): convert the non-first char to lowercase

### LINKED ISSUE #580: startCase wrong behaviour/s
`startCase` is not working properly when an uppercase string is provided, you have to `toLowerCase` it first.

Also, it's removing commas. 

Is this the intended behaviour?

hyesungoh:
Hi there 👍 

I saw some [strange behavior at startCase](https://github.com/toss/es-toolkit/issues/580)

so I handle this issue at original path's startCase (non-compat)

> I did not handle compat's startCase, but if you think it's necessary, I'll apply it! 
> Please feel free to talk about it 😄 

## benchmark

스크린샷 2024-09-24 오전 12 28 06

fix #580 

## related pr

- #517 
- #313
- #203


raon0211 (approved):
Thanks for fixing the bug!

PR #594, opened 2024-09-24
feat(includes): add `includes` in `es-toolkit/object`

dayongkr:
# Description

I have added an `includes` method for `Object`. Unlike lodash, it doesn't allow `Array` and `String` value. This is because both `String` and `Array` already have native `includes` methods like `Array.prototype.includes`.

My implementation uses the `SameValueZero` comparison algorithm, similar to `Array.prototype.includes`.

### 1. Not Using `Object.values`

Screenshot 2024-09-24 at 6 56 20 PM

Implementing it like `Object.values(obj).includes(value)` forces access to both all of keys and all of values as shown as a image above(under the hood of Object.values). This approach is less efficient.

Therefore, I chose not to use `Object.values` and `Array.prototype.includes`. Instead, I implemented the function using `Object.keys` and manually implemented the `SameValueZero` comparison. This allows early return upon finding the value, improving performance.

Screenshot 2024-09-24 at 6 42 47 PM

> In the above image, `include` is implemented using `Object.keys`, and `include2` using `Object.values`.

### ~~2. Symbol-keyed property~~

~~By default, `Object.keys` only returns `enumerable string-keyed properties`. However, users may want to check `symbol-keyed properties`, both enumerable and non-enumerable. To address this, I added an option `options.allowSymbolKeyed`. If enabled, the function will check symbol-keyed properties using `Reflect.ownKeys()`.~~

~~Since `Reflect.ownKeys()` has some performance overhead, this feature is optional. I’d like feedback on whether this interface design is user-friendly or if improvements are needed.~~

In my opinion, there's no need to implement this all the way from the initial interface. I removed that option because other includes methods only support enumerable string keyed. Additionally, since this case is likely to be very minor, I removed it to adhere to the principle of simplicity.

## Benchmark

Screenshot 2024-09-24 at 6 48 15 PM

### DIFF
--- a/benchmarks/performance/includes.bench.ts
+++ b/benchmarks/performance/includes.bench.ts
@@ -1,0 +1,39 @@
+import { bench, describe } from 'vitest';
+import { includes as includesToolkit } from 'es-toolkit';
+import { includes as includesLodash } from 'lodash';
+
+describe('includes', () => {
+  const object = {
+    a: 1,
+    b: 'a',
+    c: NaN,
+    d: undefined,
+    e: null,
+    f: Infinity,
+    g: -0,
+  };
+
+  bench('es-toolkit/includes', () => {
+    includesToolkit(object, 1);
+    includesToolkit(object, 'a');
+    includesToolkit(object, NaN);
+    includesToolkit(object, undefined);
+    includesToolkit(object, null);
+    includesToolkit(object, Infinity);
+    includesToolkit(object, Symbol('sym1'));
+    includesToolkit(object, -0);
+  });
+
+  bench('lodash/includes', () => {
+    includesLodash(object, 1);
+    includesLodash(object, 'a');
+    includesLodash(object, NaN);
+    includesLodash(object, undefined);
+    includesLodash(object, null);
+    includesLodash(object, Infinity);
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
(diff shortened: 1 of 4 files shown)


raon0211:
Thanks for your contribution! I’m unsure if this function is needed in `es-toolkit`.

- The name `includes` suggests it should work for arrays and strings. If it doesn’t, it could confuse users. If we add it, we should support those types as well.
- While I see that `es-toolkit`’s `includes` is faster than `Object.values(...).includes(...)`, the syntax `includes({ a: 1, b: 2, c: 3 }, 2)` can be confusing. The native method is more intuitive, especially since it's unclear whether it checks keys or values. I believe `es-toolkit` should focus on very intuitive functions.

dayongkr:
> Thanks for your contribution! Meanwhile, I’m unsure if this function is needed in `es-toolkit`.
> 
> * The name `includes` suggests it should work for arrays and strings. If it doesn’t, it could confuse users. If we add it, we should support those types as well.
> * While I see that `es-toolkit`’s `includes` is faster than `Object.values(...).includes(...)`, the syntax `includes({ a: 1, b: 2, c: 3 }, 2)` can be confusing. It's unclear whether it checks keys or values. I guess the native method is more intuitive, and I believe `es-toolkit` should focus on very intuitive functions.

Thank you for explaining the reason! I will close the PR.

PR #596, opened 2024-09-25
feat(isSet): add isSet

chhw130:
Hi!🖐️
I appreciate your great works for environment!
I added `isSet` function.

isSet bench


raon0211 (approved):
Thanks for your work!

PR #600, opened 2024-09-26
feat(isMap): add isMap

chhw130:
Hi there!
I add `isMap` function.

isMap

### DIFF
--- a/src/predicate/isMap.ts
+++ b/src/predicate/isMap.ts
@@ -1,0 +1,21 @@
+/**
+ * Checks if a given value is `Map`.
+ *
+ * This function can also serve as a type predicate in TypeScript, narrowing the type of the argument to `Map`.
+ *
+ * @param {unknown} value The value to check if it is a `Map`.
+ * @returns {value is Map<any, any>} Returns `true` if `value` is a `Map`, else `false`.
+ *
+ * @example
+ * const value1 = new Map();
+ * const value2 = new Set();
+ * const value3 = new WeakMap();
+ *
+ * console.log(isMap(value1)); // true
+ * console.log(isMap(value2)); // false
+ * console.log(isMap(value3)); // false
+ */
+
+export function isMap(value: unknown): value is Map<any, any> {
+  return value instanceof Map;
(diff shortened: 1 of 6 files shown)


src/predicate/isMap.ts
```diff
@@ -0,0 +1,21 @@
...
+ * const value2 = new Set();
+ * const value3 = new WeakMap();
+ *
+ * console.log(isMap(value1)); // true
+ * console.log(isMap(value2)); // false
+ * console.log(isMap(value3)); // false
+ */
+
+export function isMap(value: unknown): value is Map<any, any> {
+  return value instanceof Map;
```
BasixKOR, line 20: This may not work if the map object is from other realms, see: https://web.mit.edu/jwalden/www/isArray.html

chhw130: Thank you for the feedback!

`instanceof` may not work accurately in independent context environments (especially iframe), but I've confirmed that `Array.isArray` provides consistent results in such cases.

As far as I know, there is no built-in method to properly check for a `Map`. So would using the following approach be the best option?

```javascript
Object.prototype.toString.call(new Map()) // '[object Map]'
```
Please let me know if you need any adjustments! 🙇

dayongkr: Managing cases with JavaScript realm is out of our scope, we can see related contents in [design principles of es-toolkit/compat](https://es-toolkit.slash.page/compatibility.html#design-principles).

But it is for `es-toolkit/compat`, So, I’m not sure if this applies to es-toolkit as well. I would like to hear the maintainer’s opinion on this.

raon0211: I would like to go with the current implementation.

As @dayongkr mentioned, our design principle is to [support 85% usecases with simple implementation](https://github.com/toss/es-toolkit/blob/main/.github/CONTRIBUTING.md#12-simplicity). I understand that the current implementation is not 100% correct; but it will work 99% of the time.

raon0211 (approved):
Thanks for your help!

PR #602, opened 2024-09-26
fix(startCase): add missing tests and separate logic for `es-toolkit` and `es-toolkit/compat`

dayongkr:
Due to [the change](https://github.com/toss/es-toolkit/pull/591) in the behavior of `startCase` in `es-toolkit`, the tests for `startCase` in `es-toolkit/compat` should have failed, but they did not. Upon investigation, I found that some test cases were either incorrect or missing.

I have added the missing test cases and separated the code according to the differences in behavior.


raon0211 (approved):
Thanks for catching this!

PR #604, opened 2024-09-26
docs(predicate): append docs fot `isSet` and `isMap`

hyesungoh:
I append ja and zh_hans docs for `isSet` and `isMap` 👍 

## related pr

- #596
- #600


raon0211 (approved):
thanks for your help!

PR #607, opened 2024-09-27
chore: dedupe dependencies

wojtekmaj:
This is just the output of running `yarn dedupe`. Improves CI time, project's install time and disk space usage. My suggestion would be to make it a habit to dedupe after updating dependencies of any kind to avoid this problem in the future.


raon0211 (approved):
You're true; maybe we might add this to a git hook or CI check.

PR #608, opened 2024-09-27
chore(deps): bump vitest from 1.5.2/2.0.5 to 2.1.1, fix unit test failing in Vitest 2

wojtekmaj:
Blocked by #607


D-Sketon:
Since we need https://github.com/vitest-dev/vitest/pull/6541, we may also need to bump version after the new version of vitest is released.

raon0211 (approved):
Thanks for updating!

PR #609, opened 2024-09-27
chore: fix test command

wojtekmaj:
`vitest run` implied non-development environment, disabling features like watching. So, now you can e.g.:

```
yarn test cloneDeep --watch
```

and work on your utility in peace.

At the same time, added `watch: false` to `vitest.config.mts` ensures the expected behavior of all test related commands won't change.


raon0211 (approved):
Thanks!

PR #610, opened 2024-09-27
feat(filter): add `filter` function 

k-jeonghee:
Hello 😀

I have implemented the `filter` function. I wrote test cases based on lodash’s test cases and added other possible cases to achieve 100% coverage. However, since it’s a compat function, its performance is a bit lower compared to lodash (especially when filtering objects, the performance difference is quite noticeable).

It was faster when I implemented the condition checks myself, but I decided that using the existing matches and property features would be better overall.

If you have any suggestions for further improvements, I’d be happy to make changes. Thank you! 

image1

image2

- Except object case

image3


raon0211 (dismissed):
Thanks for your awesome contribution!

src/compat/index.ts
```diff
@@ -45,6 +45,7 @@ export { sortBy } from './array/sortBy.ts';
 export { size } from './array/size.ts';
 export { some } from './array/some.ts';
 export { zipObjectDeep } from './array/zipObjectDeep.ts';
+export { filter } from './array/filter.ts';
```
raon0211, line 48: I guess reordering exports should be done in a separate pull request 😢

raon0211 (approved):
Thanks!

src/compat/index.ts
k-jeonghee: I missed that part😢 I will keep it in mind for my next contribution. Thank you for the advice!

PR #611, opened 2024-09-27
fix(throttle): not throttling after initial throttleMs

jonganebski:
## Description

- `throttle` fails to throttle its callback function after initial `throttleMs` delay. 
- Fixed the issue by updating the `pendingAt` timestamp when the callback function is executed. 

```ts
import { throttle } from "es-toolkit";

const startedAt = Date.now();
const throttledFunc = throttle(() => console.log(Date.now() - startedAt), 1_000);
setInterval(throttledFunc, 100);

// AS-IS: 100, 1102, 1202, 1301, ...
// TO-BE: 100, 1102, 2104, 3108, ...
```

### DIFF
--- a/src/function/throttle.spec.ts
+++ b/src/function/throttle.spec.ts
@@ -28,19 +28,19 @@ describe('throttle', () => {
     throttledFunc(); // should be ignored
     expect(func).toHaveBeenCalledTimes(1);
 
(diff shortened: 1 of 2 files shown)


raon0211 (approved):
Thanks for catching this!

PR #613, opened 2024-09-28
docs(isSet): Update docs for `isSet`

mass2527:


raon0211 (approved):
Thanks for catching this!

PR #614, opened 2024-09-29
fix(matchesProperty): Add boolean type guard 

k-jeonghee:
Hello, I hope you're having a great weekend. 😊

I found that I needed to add a type guard for boolean types in the `matchesProperty` function. In a previous PR ([#610](https://github.com/toss/es-toolkit/pull/610)), I noticed a problem where, when using a tuple as a predicate in the `filter` function, the test case didn’t pass if the value was boolean.

So, I added new logic to fix this, and I also added a test case for matchesProperty to make sure the test coverage is 100%. I checked, and there’s almost no performance difference.

Could you check if this logic is really necessary? Thank you!

#### The failing cases in the filter function
```typescript
const users = [
  { user: 'barney', age: 36, active: true },
  { user: 'fred', age: 40, active: false },
];

it(`should filter objects based on the specified key and value pair`, () => {
  expect(filter(users, ['active', false])).toEqual([{ user: 'fred', age: 40, active: false }]);
});
```

### DIFF
--- a/src/compat/predicate/matchesProperty.spec.ts
+++ b/src/compat/predicate/matchesProperty.spec.ts
@@ -8,6 +8,13 @@ import { numberProto } from '../_internal/numberProto';
 import { cloneDeep } from '../object/cloneDeep';
 
 describe('matchesProperty', () => {
+  it('should return true when the property value matches the boolean source', () => {
+    const object = { a: false };
+    const matches = matchesProperty('a', false);
+
+    expect(matches(object)).toBe(true);
+  });
+
   it('should create a function that performs a deep comparison between a property value and `srcValue`', () => {
     let object: any = { a: 1, b: 2, c: 3 };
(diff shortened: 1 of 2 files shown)


raon0211:
Hello, thanks for your pull request! 😃 

In `lodash`, `matchesProperty` is a convenient utility for `isMatch`. So the implementation should be fixed on the side of `isMatch`.

We implemented a fix here: Could you have a look at this?

https://github.com/toss/es-toolkit/commit/ac1e79e0c48f3e24dfacd47cdf4a2da1fa9eacc9

k-jeonghee:
@raon0211 

Yes, I’ve checked the fix and everything looks good.
It was a great opportunity to learn a lot.

Thank you again for reviewing it! ✨

raon0211:
Thanks for looking at this! Closing the pull request for now.

PR #615, opened 2024-09-29
docs(dropWhile): Update docs for `dropWhile`

mass2527:
Thanks for adding full compatibility in this [commit](https://github.com/toss/es-toolkit/commit/c0efc6536091aaf5774bffb9698378af52607ae7). I learned a lot from it. While reviewing, I noticed a missing part, so I added it.


raon0211 (approved):
Thanks for catching this!

PR #616, opened 2024-09-29
chore: delete unused babel config file

dayongkr:
# Description

- Delete `babel.config.js` - we don't use babel.


raon0211 (approved):
Thanks! Probably this is from when we started the project, and the build configuration was different..

PR #617, opened 2024-09-29
docs(every): Update docs for `every`

mass2527:
- Added missing signature.
- Fixed incorrect results.
- Updated variable name for better clarity.


raon0211 (approved):
This is a really great help! 😃

PR #618, opened 2024-09-29
docs(flip): Update docs for `flip`

mass2527:
i've made documentation consistent with other languages


raon0211 (approved):
Thanks for catching this!

PR #619, opened 2024-09-29
fix(filter): Ensure `filter` provide correct arguments to the predicate function

mass2527:
- I've fixed `filter` to provide correct arguments when it is used with object.
- Updated docs

### DIFF
--- a/src/compat/array/filter.ts
+++ b/src/compat/array/filter.ts
@@ -170,7 +170,8 @@ export function filter<T>(
 
         for (let i = 0; i < entries.length; i++) {
           const entry = entries[i];
-          const [key, value] = entry;
+          const key = entry[0];
+          const value = entry[1];
 
           if (predicate(value, key, source)) {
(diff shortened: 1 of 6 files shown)


src/compat/array/filter.ts
```diff
@@ -165,6 +164,22 @@ export function filter<T>(
 
   switch (typeof predicate) {
     case 'function': {
+      if (!Array.isArray(source)) {
+        const result: T[] = [];
+        const entries: any[] = Object.entries(source);
+
+        for (let i = 0; i < entries.length; i++) {
+          const entry = entries[i];
+          const [key, value] = entry;
```
raon0211, line 173:
```suggestion
          const key = entry[0];
          const value = entry[1];
```
We're accessing `key` and `value` with indices for performance reasons.

raon0211 (approved):
Thanks! Wondering why we have wrong examples here..

PR #620, opened 2024-09-29
chore(lint): Add `prefer-object-has-own` eslint rule and fix lint error

dayongkr:
# Description

`Object.hasOwn` behaves the same as `Object.prototype.hasOwnProperty.call`, but with more explicit syntax. For the sake of consistency and better maintainability, I am submitting this PR.


raon0211 (approved):
Didn't know there was an alternative API. Thanks!

PR #627, opened 2024-09-30
fix(includes): Ignore inherited value when searching object

mass2527:
Fixed `includes` to ignore the inherited value.

### DIFF
--- a/src/compat/array/includes.ts
+++ b/src/compat/array/includes.ts
@@ -121,7 +121,9 @@ export function includes(
 
   // Not using `Object.keys` because it can't access inherited properties.
   for (const key in source) {
-    keys.push(key);
+    if (Object.hasOwn(source, key)) {
+      keys.push(key);
+    }
   }
 
   if (fromIndex < 0) {
--- a/src/compat/array/includes.ts
+++ b/src/compat/array/includes.ts
@@ -117,14 +117,7 @@ export function includes(
     return source.includes(target, fromIndex);
   }
 
-  const keys = [];
-
-  // Not using `Object.keys` because it can't access inherited properties.
-  for (const key in source) {
-    if (Object.hasOwn(source, key)) {
-      keys.push(key);
-    }
-  }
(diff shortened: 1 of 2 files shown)


src/compat/array/includes.ts
dayongkr, line 1: We can use `Object.keys()` and for-loop

```javscript
const keys = Object.keys(source)

   keys.push(key);
}
  }

> https://developer.mozilla.org/en-US/docs/Web/JavaScript/Enumerability_and_ownership_of_properties

mass2527: @dayongkr 
Thanks for the review! 
As you mentioned, my implementation can be replaced with `Object.keys`. 
I am going to refactor this.

dayongkr: Thank you for accepting my suggestion!

raon0211:
I'm not sure, in lodash, is `includes` supposed to include inherited properties?

dayongkr:
> I'm not sure, in lodash, is `includes` supposed to include inherited properties? I thought this was intended behavior.

It seems correct to fix it. If we look at the [baseKeys](https://github.com/lodash/lodash/blob/ddfd9b11a0126db2302cb70ec9973b66baec0975/lodash.js#L3488) function used in includes, it uses hasOwnKeys to skip inherited keys.

mass2527:
lodash `includes` function retrieves values of object using [`values`](https://github.com/lodash/lodash/blob/ddfd9b11a0126db2302cb70ec9973b66baec0975/lodash.js#L13902-L13930).

The documentation for `values` describes it as follows: 'Creates an array of the **own enumerable string-keyed** property values of object.' And that is exactily what [`Object.keys`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Enumerability_and_ownership_of_properties#traversing_object_properties) does. So, I believe this is correct.

raon0211 (approved):
Thanks!

PR #631, opened 2024-10-01
docs(curryRight): Rename parameters for clarity and update docs

mass2527:
- Updated parameter names for better understanding


raon0211 (approved):
Good catch!

PR #638, opened 2024-10-02
refactor(range): Enhance type safety in the implementation

tooooo1:
`range` was returning `any[]`. Updated it to explicitly return `number[]` by using `new Array<number>(length)` for better type safety. `range` should always return an array of numbers.

#### AS-IS
스크린샷 2024-10-02 오후 1 58 46

#### TO-BE
스크린샷 2024-10-02 오후 1 59 07


raon0211 (approved):
Thanks!

PR #641, opened 2024-10-02
feat(defaults): add `defaults` in compat layer

### LINKED ISSUE #639: Support for `defaults`
```javascript
defaults({ 'a': 1 }, { 'b': 2 }, { 'a': 3 });
// => { 'a': 1, 'b': 2 }
```

I implemented it in compat layer at first.

I wondering that it mutates a first argument (object), but i think that we should not mutate it in `es-toolkit` (excluded compat layer).

dayongkr:
# Description

I added `defaults` in compat layer.

## Benchmark
Screenshot 2024-10-02 at 3 55 05 PM

## Additional

I am wondering if it is also necessary for `es-toolkit/object`. If so, do you think it is appropriate to implement it in a way that does not mutate `object`? I would like to hear some opinions.

related #639


src/compat/object/defaults.ts
```diff
@@ -0,0 +1,150 @@
+import { eq } from '../_internal/eq';
```
raon0211, line 1:
```suggestion
import { eq } from '../util/eq.ts';
```

src/compat/object/defaults.ts
```diff
@@ -0,0 +1,150 @@
...
+ * @param {S[]} sources - The source objects.
+ * @returns {object} Returns the merged object, ensuring that the destination object is not `null` or `undefined`.
+ *
+ * @example
+ * defaults({ a: 1 }, { a: 2, b: 2 }, { c: 3 }); // { a: 1, b: 2, c: 3 }
+ * defaults({ a: 1, b: 2 }, { b: 3 }, { c: 3 }); // { a: 1, b: 2, c: 3 }
+ * defaults({ a: null }, { a: 1 }); // { a: null }
+ * defaults({ a: undefined }, { a: 1 }); // { a: 1 }
+ */
+
```
raon0211, line 127:
```suggestion
```

raon0211 (approved):

kennyjwilli:
Curious - how come this was implemented in compat layer & not the main package?

dayongkr:
Once a function is added to the main package, it cannot be easily removed due to backward compatibility concerns, requiring more extensive discussion.

Therefore, for faster compatibility support, I have initially added it only to the compat layer.

PR #642, opened 2024-10-02
feat(eq): add `eq` in compat layer

dayongkr:
# Description

I added `eq` in compat layer. It performs a `SameValueZero` comparison.

I think it also needs in `es-tookit/util`, but we don't have a util folder in `es-toolkit`. So I added it in compat layer at first.

## Benchmark

Screenshot 2024-10-02 at 7 58 15 PM

It is a tiny function. So this performance is equal with `lodash`.


raon0211 (approved):
Thanks!

PR #644, opened 2024-10-02
feat(compat) : implement `isArrayBuffer` compat

chhw130:
I added `isArrayBuffer` in compat.

### DIFF
--- a/src/predicate/isArrayBuffer.ts
+++ b/src/predicate/isArrayBuffer.ts
@@ -16,6 +16,6 @@
  * console.log(isArrayBuffer(value3)); // false
  */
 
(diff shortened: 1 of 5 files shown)


src/predicate/isArrayBuffer.ts
```diff
@@ -16,6 +16,6 @@
  * console.log(isArrayBuffer(value3)); // false
  */
 
-export function isArrayBuffer(value: unknown): value is ArrayBuffer {
+export function isArrayBuffer(value?: unknown): value is ArrayBuffer {
```
raon0211, line 19:
```suggestion
export function isArrayBuffer(value: unknown): value is ArrayBuffer {
```

raon0211 (approved):
Thanks!

PR #646, opened 2024-10-02
feat(rangeRight): implement rangeRight

D-Sketon:
benchmark
{14299058-84C2-4350-AAAC-D3138BA71623}


src/math/rangeRight.ts
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  const length = Math.max(Math.ceil((end - start) / step), 0);
+  const result = new Array(length);
+
+  for (let i = 0; i < length; i++) {
+    result[i] = start + (length - i - 1) * step;
+  }
+
+  return result;
```
raon0211, line 78: What do you think about the following implementation? We should check the performance, though

```tsx
return range(start, end, step).reverse();
```

raon0211: Yeah, I guess the `reverse` makes our function twice slower in small arrays, so we might go with the current implementation...

raon0211 (approved):
Thanks so much!

PR #647, opened 2024-10-03
fix(before,once): allow accepting arguments for `once` and fix types of `before` and `once`

dayongkr:
# Description

- `before`
  - I corrected the part where the return type was defined as the assigned function, even though it returns a new function.
- `once`:
  - Previously, the function was designed to only accept functions without arguments, but we needs to accept functions with arguments. So I added this feature.
  - I simplified the code a bit.

### DIFF
--- a/docs/ja/reference/function/before.md
+++ b/docs/ja/reference/function/before.md
@@ -5,7 +5,7 @@
 ## インターフェース
 
 ```typescript
-function before<F extends (...args: any[]) => any>(n: number, func: F): F;
+function before<F extends (...args: unkown[]) => any>(n: number, func: F): (...args: Parameters<F>) => ReturnType<F> | undefined;
 ```
 
 ### パラメータ
@@ -17,7 +17,7 @@ function before<F extends (...args: any[]) => any>(n: number, func: F): F;
 
 ### 戻り値
 
-(`F`): 新しい関数を返します。この関数は以下の機能を持ちます：
+(`(...args: Parameters<F>) => ReturnType<F> | undefined`): 新しい関数を返します。この関数は以下の機能を持ちます：
 
 - 呼び出し回数を追跡します。
 - `n-1` 回目の呼び出しまで `func` を呼び出します。
--- a/docs/ja/reference/function/before.md
+++ b/docs/ja/reference/function/before.md
@@ -5,7 +5,7 @@
 ## インターフェース
 
 ```typescript
-function before<F extends (...args: unkown[]) => any>(
+function before<F extends (...args: any[]) => any>(
   n: number,
   func: F
 ): (...args: Parameters<F>) => ReturnType<F> | undefined;
--- a/docs/ja/reference/function/before.md
+++ b/docs/ja/reference/function/before.md
@@ -5,7 +5,10 @@
 ## インターフェース
 
 ```typescript
-function before<F extends (...args: unkown[]) => any>(n: number, func: F): (...args: Parameters<F>) => ReturnType<F> | undefined;
+function before<F extends (...args: unkown[]) => any>(
+  n: number,
+  func: F
+): (...args: Parameters<F>) => ReturnType<F> | undefined;
 ```
 
 ### パラメータ
--- a/docs/ja/reference/function/once.md
+++ b/docs/ja/reference/function/once.md
@@ -6,16 +6,16 @@
 ## インターフェース
 
 ```typescript
-function once<F extends () => any>(func: F): F;
+function once<F extends (...args: unknown[]) => any>(func: F): () => ReturnType<F>;
 ```
 
 ### パラメータ
 
-- `func` (`F extends () => any`): 一度だけ呼び出すように制限する関数です。
+- `func` (`F extends (...args: unknown[]) => any`): 一度だけ呼び出すように制限する関数です。
 
 ### 戻り値
 
-(`F`): `func` が一度呼び出されると結果をキャッシュして返す新しい関数です。
+(`() => ReturnType<F>`): `func` が一度呼び出されると結果をキャッシュして返す新しい関数です。
 
 ## 例
 
--- a/docs/ja/reference/function/once.md
+++ b/docs/ja/reference/function/once.md
@@ -6,16 +6,17 @@
 ## インターフェース
 
 ```typescript
-function once<F extends (...args: any[]) => any>(func: F): (...args: Parameters<F>) => ReturnType<F>;
+function once<F extends () => any>(func: F): F;
+function once<F extends (...args: any[]) => void>(func: F): F;
 ```
 
 ### パラメータ
 
-- `func` (`F extends (...args: any[]) => any`): 一度だけ呼び出すように制限する関数です。
+- `func` (`F extends (() => any) | ((...args: any[]) => void)`): 一度だけ呼び出すように制限する関数です。
 
 ### 戻り値
 
-(`(...args: Parameters<F>) => ReturnType<F>`): `func` が一度呼び出されると結果をキャッシュして返す新しい関数です。
+(`F`): `func` が一度呼び出されると結果をキャッシュして返す新しい関数です。
 
 ## 例
 
--- a/docs/ja/reference/function/once.md
+++ b/docs/ja/reference/function/once.md
@@ -6,12 +6,12 @@
 ## インターフェース
 
 ```typescript
-function once<F extends (...args: unknown[]) => any>(func: F): () => ReturnType<F>;
+function once<F extends (...args: any[]) => any>(func: F): () => ReturnType<F>;
 ```
 
 ### パラメータ
 
-- `func` (`F extends (...args: unknown[]) => any`): 一度だけ呼び出すように制限する関数です。
+- `func` (`F extends (...args: any[]) => any`): 一度だけ呼び出すように制限する関数です。
 
 ### 戻り値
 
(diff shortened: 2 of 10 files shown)


docs/ja/reference/function/before.md
```diff
@@ -5,7 +5,10 @@
 ## インターフェース
 
 ```typescript
-function before<F extends (...args: any[]) => any>(n: number, func: F): F;
+function before<F extends (...args: unkown[]) => any>(
```
raon0211, line 8:
```suggestion
function before<F extends (...args: any[]) => any>(
```

Actually, since parameters are contravariant, we have to specify `never` or `any` in the parameters.

docs/ja/reference/function/once.md
```diff
@@ -6,16 +6,16 @@
 ## インターフェース
 
 ```typescript
-function once<F extends () => any>(func: F): F;
+function once<F extends (...args: unknown[]) => any>(func: F): () => ReturnType<F>;
```
raon0211, line 9:
```suggestion
function once<F extends (...args: any[]) => any>(func: F): () => ReturnType<F>;
```

docs/ja/reference/function/once.md
```diff
@@ -6,16 +6,16 @@
...
 ```
 
 ### パラメータ
 
-- `func` (`F extends () => any`): 一度だけ呼び出すように制限する関数です。
+- `func` (`F extends (...args: unknown[]) => any`): 一度だけ呼び出すように制限する関数です。
```
raon0211, line 14:
```suggestion
- `func` (`F extends (...args: any[]) => any`): 一度だけ呼び出すように制限する関数です。
```

src/function/once.ts
```diff
@@ -15,20 +15,16 @@
  * initialize(); // Logs: 'Initialized!' and returns true
  * initialize(); // Returns true without logging
  */
-export function once<F extends () => any>(func: F): F {
+export function once<F extends (...args: unknown[]) => any>(func: F): () => ReturnType<F> {
```
raon0211, line 18:
```suggestion
export function once<F extends () => any>(func: F): () => ReturnType<F> {
```
I think `func` should be a unary function.

src/function/before.ts
```diff
@@ -27,16 +27,21 @@
  * beforeFn();
  */
 
-export function before<F extends (...args: any[]) => any>(n: number, func: F): F {
+export function before<F extends (...args: unknown[]) => any>(
```
raon0211, line 30:
```suggestion
export function before<F extends (...args: any[]) => any>(
```

docs/zh_hans/reference/function/once.md
```diff
@@ -7,16 +7,16 @@
 ## 签名
 
 ```typescript
-function once<F extends () => any>(func: F): F;
+function once<F extends (...args: unknown[]) => any>(func: F): () => ReturnType<F>;
```
raon0211, line 10:
```suggestion
function once<F extends () => any>(func: F): () => ReturnType<F>;
```

docs/zh_hans/reference/function/once.md
```diff
@@ -7,16 +7,16 @@
...
 ```
 
 ### 参数
 
-- `func` (`F extends () => any`): 要限制的函数。
+- `func` (`F extends (...args: unknown[]) => any`): 要限制的函数。
```
raon0211, line 15:
```suggestion
- `func` (`F extends () => any`): 要限制的函数。
```

docs/reference/function/once.md
```diff
@@ -6,16 +6,16 @@ Repeated calls to the function will return the value from the first invocation.
 ## Signature
 
 ```typescript
-function once<F extends () => any>(func: F): F;
+function once<F extends (...args: unknown[]) => any>(func: F): () => ReturnType<F>;
```
raon0211, line 9:
```suggestion
function once<F extends () => any>(func: F): () => ReturnType<F>;
```

docs/reference/function/once.md
```diff
@@ -6,16 +6,16 @@ Repeated calls to the function will return the value from the first invocation.
...
 ```
 
 ### Parameters
 
-- `func` (`F extends () => any`): The function to restrict.
+- `func` (`F extends (...args:unknown[]) => any`): The function to restrict.
```
raon0211, line 14:
```suggestion
- `func` (`F extends () => any`): The function to restrict.
```

src/function/once.ts
raon0211: If `func` is unary, why can't `once` just return `F` here?

dayongkr: > I think `func` should be a function accepting no arguments.
``` javascript
function say(str) {
  console.log(str)
}

const onced = once(say)
onced("hello")
```

In this case, the say function requires an argument, so func should not be restricted to functions that accept no arguments.

docs/ja/reference/function/before.md
dayongkr: Thanks to review!

docs/zh_hans/reference/function/before.md
```diff
@@ -5,7 +5,10 @@
 ## 签名
 
 ```typescript
-function before<F extends (...args: any[]) => any>(n: number, func: F): F;
+function before<F extends (...args: unkown[]) => any>(
```
dayongkr, line 8:
```suggestion
function before<F extends (...args: any[]) => any>(
```

docs/reference/function/before.md
```diff
@@ -5,7 +5,10 @@ Creates a new function that limits the number of times the given function (`func
 ## Signature
 
 ```typescript
-function before<F extends (...args: any[]) => any>(n: number, func: F): F;
+function before<F extends (...args: unkown[]) => any>(
```
dayongkr, line 8:
```suggestion
function before<F extends (...args: any[]) => any>(
```

docs/ko/reference/function/before.md
```diff
@@ -5,7 +5,10 @@
 ## 인터페이스
 
 ```typescript
-function before<F extends (...args: any[]) => any>(n: number, func: F): F;
+function before<F extends (...args: unknown[]) => any>(
```
dayongkr, line 8:
```suggestion
function before<F extends (...args: any[]) => any>(
```

docs/ko/reference/function/once.md
```diff
@@ -6,16 +6,16 @@
 ## 인터페이스
 
 ```typescript
-function once<F extends () => any>(func: F): F;
+function once<F extends (...args: unknown[]) => any>(func: F): () => ReturnType<F>;
```
dayongkr, line 9:
```suggestion
function once<F extends (...args: any[]) => any>(func: F): () => ReturnType<F>;
```

src/function/once.ts
dayongkr: Also, even if the interface is updated, it shouldn’t cause any issues with backward compatibility.

dayongkr: > If `func` is a function accepting no arguments, why can't `once` just return `F` here?

I’ve applied this to the code. Thank you!

raon0211: If we allowed functions accepting arguments, we might end up with this code:

```tsx
function sum(x, y) {
  return x + y;
}

const onced = once(sum);
sum(1, 2) // 3
sum(2, 4) // 3
```

which is very awkward. So I thought we should restrict our function to functions which accept no arguments...

dayongkr: > If we allowed functions accepting arguments, we might end up with this code:
> 
> ```tsx
> function sum(x, y) {
>   return x + y;
> }
> 
> const onced = once(sum);
> sum(1, 2) // 3
> sum(2, 4) // 3
> ```
> 
> which is very awkward. So I thought we should restrict our function to functions which accept no arguments...

This function returns a new function, so a situation like the one above is unlikely to occur..?

raon0211: When searching through the use cases of `once`, I think there are two usecases:

**[1] Ensuring that we initialize certain objects only once, like:**

```tsx
const createClient = once(() => new Client());

createClient();
createClient(); // client is created only once
```

**[2] Ensuring that we make a side effect only once, like:**

```tsx
const log = once(console.log);

log('hello')
log('world') // this call is ignored
```

But I guess [1] is covered with [memoize](https://es-toolkit.slash.page/reference/function/memoize.html#memoize), so `once` might focus on solving the problem [2].

raon0211: If we were to support [2], we might restrict the function to return `void` when it receives parameters, like:

```tsx
export function once<F extends () => any>(func: F): F;
export function once<F extends (...args: any) => void>(func: F): F;
```

dayongkr: > If we were to support [2], we might restrict the function to return `void` when it receives parameters, like:
> 
> ```tsx
> export function once<F extends () => any>(func: F): F;
> export function once<F extends (...args: any) => void>(func: F): F;
> ```

Oh, if we change it like this, the function’s role will become much clearer!

raon0211: But since we need backward compatibility, the best option we can choose might be supporting both:

```tsx
export function once<F extends () => any>(func: F): F;
export function once<F extends (...args: any) => void>(func: F): F;
```

dayongkr: > But since we need backward compatibility, the best option we can choose might be supporting both:
> 
> ```tsx
> export function once<F extends () => any>(func: F): F;
> export function once<F extends (...args: any) => void>(func: F): F;
> ```

I added these suggestions to the code.

raon0211 (approved):

PR #648, opened 2024-10-03
feat(before): add `before` in compat layer

dayongkr:
# Descritpiton

Our implementation of `before` returns `undefined` after the `n-1` call, while Lodash's implementation returns the result of the `n-1` call. Therefore, I couldn't use our function, so I re-implemented.

## Benchmarks

Screenshot 2024-10-03 at 2 28 53 PM

The issue arises due to the overhead involved when importing the module, which is causing this result. If we resolve the part where `toInteger` is imported, the output will be the same as Lodash.

### DIFF
--- a/benchmarks/performance/before.bench.ts
+++ b/benchmarks/performance/before.bench.ts
@@ -1,28 +1,57 @@
 import { bench, describe } from 'vitest';
 import { before as beforeToolkit_ } from 'es-toolkit';
+import { before as beforeToolkitCompat_ } from 'es-toolkit/compat';
 import { before as beforeLodash_ } from 'lodash';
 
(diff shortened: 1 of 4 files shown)


raon0211:
Can you guess why `before` in our compat library is slower than lodash?

dayongkr:
> Can you guess why `before` in our compat library is slower than lodash?

I answered in our discord channel!

dayongkr:
> Can you guess why `before` in our compat library is slower than lodash?

The `toInteger` function used internally is much faster in our code, so calling this function shouldn’t be an issue.

Since there are no additional internal operations, it seems like the import overhead had a significant impact.

raon0211 (approved):
Thanks!

PR #652, opened 2024-10-03
fix(compat): add missing export

D-Sketon:


raon0211 (approved):
I guess we might need a linter here.. 😢

PR #656, opened 2024-10-03
feat(times): add `times` in compat layer

dayongkr:
# Description

I added `times` in compat layer. 

I prioritized its implementation because it frequently appeared in compatibility tests.

## Benchmark

Screenshot 2024-10-04 at 12 20 49 AM


raon0211 (approved):
Thanks!

PR #657, opened 2024-10-03
feat(constant): add `constant` in compat layer

dayongkr:
# Description

It is very simple function, but I implemented this as it frequently appears in compatibility test code.

## Benchmark

Screenshot 2024-10-04 at 12 37 16 AM


raon0211 (approved):
Thanks!

PR #658, opened 2024-10-03
chore(deps): remove a circular dependency and update rollup plugin and some of types package

dayongkr:
I encountered a circular dependency while analyzing the bundling process and resolved it by adjusting the `eq` import in `isEqualWith`. 

Additionally, I updated three packages that had significantly diverged from their recent versions and performed the `dedupe` command.


raon0211 (approved):
thanks!

PR #659, opened 2024-10-03
feat(slice): add `slice` in compat layer

dayongkr:
# Description

I added `slice` that does not return a dense array for sparse arrays unlike the native `Array.prototype.slice`.

So I couldn't use the native API.

## Benchmark

Screenshot 2024-10-04 at 2 39 43 AM


src/compat/_internal/isIndex.ts
```diff
@@ -1,9 +1,11 @@
 const IS_UNSIGNED_INTEGER = /^(?:0|[1-9]\d*)$/;
 
-export function isIndex(value: PropertyKey) {
+export function isIndex(value: PropertyKey, length?: number): boolean {
```
raon0211, line 3: I understand that this function is already defined in `lodash`, but I think receiving the `length` as an argument is very confusing. For example, I can't predict what will happen for the following code:

```tsx
isIndex(value, 3);
```

So I wish we move the clamping logic outside of `isIndex`. What do you think?

src/compat/array/slice.ts
```diff
@@ -0,0 +1,51 @@
...
+ * @param {number} [start=0] - The start position.
+ * @param {number} [end=array.length] - The end position.
+ * @returns {T[]} - Returns the slice of `array`.
+ *
+ * @example
+ * slice([1, 2, 3], 1, 2); // => [2]
+ * slice(new Array(3)); // => [undefined, undefined, undefined]
+ */
+export function slice<T>(array: readonly T[], start?: number, end?: number): T[] {
+  const length = array == null ? 0 : array.length;
```
raon0211, line 20: If `array` is null, why not we return an empty array first? This will help reduce the complexity of the function.

```suggestion
  if (array == null) {
    return [];
  }
  
  const length = array.length;
```

src/compat/array/slice.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  if (end && typeof end !== 'number' && isIterateeCall(array, start, end)) {
+    // support for expression like `_.map(slice)`
+    start = 0;
+    end = length;
+  } else {
+    start = start == null ? 0 : toInteger(start);
+    end = end === undefined ? length : toInteger(end);
+  }
```
raon0211, line 29: I think the code here is too complex. We have three conditions here: (1) The `if` statement, (2) The first ternary operator, and (3) the second ternary operator. So we have to think about `2^3 === 8` possibilities.

raon0211: What about we reduce the code complexity as follows:

```tsx
// We get rid of null checks by default values
export function slice<T>(array: readonly T[], start = 0, end = array.length): T[] {
  if (typeof end !== 'number') {
    end = length;
  }

  start = toInteger(start);
  end = toInteger(end);

  /* ... */
```

dayongkr: I made modifications, but I left some parts in for compatibility and performance.

src/compat/_internal/isIndex.ts
dayongkr: That seems like a good idea, so I applied it to the code as well!

dayongkr:
And lint error will be fixed in #674

src/compat/array/slice.ts
```diff
@@ -0,0 +1,57 @@
+import { isIterateeCall } from '../_internal/isIterateeCall';
+import { toInteger } from '../util/toInteger';
```
raon0211, line 2:
```suggestion
import { isIterateeCall } from '../_internal/isIterateeCall.ts';
import { toInteger } from '../util/toInteger.ts';
```

raon0211 (approved):
Thanks!

PR #660, opened 2024-10-04
feat(defaultTo) : add `defaultTo` util

chhw130:
I added `defaultTo` in compat.
And I am wondering about the absence of es-toolkit/util. 👀

defaultTo


src/compat/util/defaultTo.ts
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * @example
+ * defaultTo(null, 'default') // returns 'default'
+ * defaultTo(undefined, 42) // returns 42
+ * defaultTo(NaN, 0) // returns 0
+ * defaultTo('actual', 'default') // returns 'actual'
+ * defaultTo(123, 0) // returns 123
+ */
+export function defaultTo(value?: unknown, defaultValue?: unknown): any {
+  return value == null || Number.isNaN(value) ? defaultValue : value;
```
raon0211, line 18:
```suggestion
  if (value == null || Number.isNaN(value)) {
    return defaultValue;
  }
  
  return value;
```

I guess `if` statements clearly shows the branches, so it might be more readable..

raon0211:
Yeah, we might need `es-toolkit/util`, but I didn't find specific functions that need to be added to `es-toolkit` for utility purposes.

raon0211 (approved):
Thanks for your help!

PR #661, opened 2024-10-04
bench: pre-build es-toolkit when bench

D-Sketon:
ref:  https://discord.com/channels/1281071127052943361/1281085859608924282/1291310699351638087

now we have two bundles when benchmarking
{E6190918-9555-4B3F-B176-18296F4F0AA9}

PR #663, opened 2024-10-04
bench: import `lodash` as inlined modules

dayongkr:
# Description 

| Before |  After |
| - | - |
|Screenshot 2024-10-04 at 4 59 49 PM|Screenshot 2024-10-04 at 4 59 24 PM|

I thought there was overhead when importing external packages, so I switched to an inline approach, and I was able to achieve the same performance without needing to reassign.

If it is merged with  [PR](https://github.com/toss/es-toolkit/pull/661 ), it seems that reassignment will no longer be necessary.

### DIFF
--- a/vitest.config.mts
+++ b/vitest.config.mts
@@ -10,6 +10,11 @@ export default defineConfig({
       include: ['src/**/*'],
       exclude: ['src/compat/_internal/**/*'],
     },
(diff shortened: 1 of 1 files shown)


dayongkr:
It was not working with the other funtions.

PR #664, opened 2024-10-04
fix(cloneDeep): cloneDeep should clone own enumerable symbol property

mass2527:
- Fixed `cloneDeep` to properly clone own enumerable symbol properties

### DIFF
--- a/src/object/cloneDeep.ts
+++ b/src/object/cloneDeep.ts
@@ -197,7 +197,7 @@ function cloneDeepImpl<T>(obj: T, stack = new Map<any, any>()): T {
 
 // eslint-disable-next-line
 export function copyProperties(target: any, source: any, stack?: Map<any, any>): void {
-  const keys = [...Object.keys(source), ...getSymbols(source)];
+  const keys = Reflect.ownKeys(source);
 
   for (let i = 0; i < keys.length; i++) {
     const key = keys[i];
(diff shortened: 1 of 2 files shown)


src/object/cloneDeep.ts
```diff
@@ -196,7 +197,7 @@ function cloneDeepImpl<T>(obj: T, stack = new Map<any, any>()): T {
 
 // eslint-disable-next-line
 export function copyProperties(target: any, source: any, stack?: Map<any, any>): void {
-  const keys = Object.keys(source);
+  const keys = [...Object.keys(source), ...getSymbols(source)];
```
raon0211, line 200:
```suggestion
  const keys = Reflect.ownKeys(source);
```

mass2527: I've used `Object.keys` and `getSymbols` to get enumerable own string and symbol properties. I think replacing them with [`Reflect.ownKeys`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Enumerability_and_ownership_of_properties#traversing_object_properties) might change the behavior, as it also includes **non-enumerable** own string and symbol properties.

When I tested Lodash's `cloneDeep`, it appeared not to clone non-enumerable properties.

```typescript
const obj = {};
Object.defineProperty(obj, 'property', {
   value: 'hello',
   enumerable: false,
});

console.info(obj.property); // hello
const cloned = cloneDeepLodash(obj);
console.info(cloned.property); // undefined
```

mass2527: [previous attempt](https://github.com/toss/es-toolkit/pull/420) to fix this issue using `Reflect.ownKeys` failed to pass the test case.

raon0211 (approved):
Thanks for your work!

PR #667, opened 2024-10-04
fix(clone): `clone` should clone function correctly

mass2527:
Unlike Lodash, the `clone` function in 'es-toolkit' returns functions as they are. So I fixed it to behave the same as Lodash.
I’m curious if this was intentional?

### DIFF
--- a/src/object/clone.spec.ts
+++ b/src/object/clone.spec.ts
@@ -119,11 +119,24 @@ describe('clone', () => {
     expect(clonedNestedObj.a[2]).toEqual(nestedObj.a[2]);
   });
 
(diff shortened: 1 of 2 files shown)


raon0211:
I'm not really sure what the behavior of `clone` should be, and I don't think `clone`ing a function should return an object in `es-toolkit`. (This behavior is very hard to predict.)

Meanwhile, we might add the behavior in `es-toolkit/compat` to match lodash's behavior.

mass2527:
I agree, cloning functions into objects seems unpredictable. This isn't a compat function, so let's keep it as is.

PR #670, opened 2024-10-06
style(format): add `prettier-plugin-sort-re-exports` prettier plugin

dayongkr:
# Description

I created a Prettier plugin that sorts re-export statements based on their source, and I applied it to `es-toolkit`.

I implemented it so that only connected re-export statements, which are not separated by line breaks, are sorted, allowing the user to sort only the desired sections.

## Additional

Screenshot 2024-10-06 at 3 59 22 PM

> And also you can check its description in [here](https://www.npmjs.com/package/prettier-plugin-sort-re-exports?activeTab=readme)


dayongkr:
I will soon add support for sorting by chunks as well.

raon0211 (approved):
Thanks!

PR #671, opened 2024-10-06
perf(omitBy): improve omitBy performance

ssi02014:
There is a performance improvement when utilizing `for...in` statements instead of `Object.entries`.

In our testing, we saw about a `1.6x` performance improvement.

스크린샷 2024-10-06 오후 4 26 45


benchmarks/performance/omitBy.bench.ts
```diff
@@ -6,15 +6,14 @@ const omitByToolkit = omitByToolkit_;
 const omitByLodash = omitByLodash_;
 
 describe('omitBy', () => {
+  const obj = { a: 1, b: 'omit', c: 3, d: 'test', e: 0 };
+  const shouldOmit = (value: number | string) => typeof value === 'string';
```
ssi02014, line 10: The code was used repeatedly, so i moved it.

src/compat/object/omit.ts
```diff
@@ -30,7 +30,7 @@ export function omit<T extends Record<string, any>, K extends keyof T>(obj: T, k
  *
  * @template T - The type of object.
  * @param {T} obj - The object to omit keys from.
- * @param {...(PropertyKey | PropertyKey[] | PropertyKey[][]} keys - A variable number of keys to be omitted from the object.
+ * @param {...(PropertyKey | PropertyKey[] | PropertyKey[][])} keys - A variable number of keys to be omitted from the object.
```
ssi02014, line 33: The `compact/omit` comment was missing parentheses, so I added them.

src/object/omitBy.ts
```diff
@@ -22,12 +22,10 @@ export function omitBy<T extends Record<string, any>>(
...
   const result: Partial<T> = {};
-
-  const objEntries = Object.entries(obj);
-  for (let i = 0; i < objEntries.length; i++) {
-    const [key, value] = objEntries[i];
+  for (const key in obj) {
+    const value = obj[key];
     if (!shouldOmit(value, key)) {
-      (result as any)[key] = value;
+      result[key] = value;
```
ssi02014, line 28: Change `Object.entries` to `for...in`. 
This also improves the `any` type assertion.

D-Sketon:
for ... in also iterates over inherited enumerable properties, does this matter?

dayongkr:
> for ... in also iterates over inherited enumerable properties, does this matter?

+ It’s important to note that inherited values will also be iterated over.

ssi02014:
@D-Sketon @dayongkr @raon0211 

Thanks for pointing out the inheritance thing.
When I tested it, I found that there was a performance benefit to proceeding after checking with `Object.has`. I can add `Object.has`.

```ts
// origin
function omitBy1<T extends Record<string, any>>(
  obj: T,
  shouldOmit: (value: T[keyof T], key: keyof T) => boolean
): Partial<T> {

  const objEntries = Object.entries(obj);
  for (let i = 0; i < objEntries.length; i++) {
    const [key, value] = objEntries[i];
    if (!shouldOmit(value, key)) {
      (result as any)[key] = value;
    }
  }

  return result;
}

// new
function omitBy2<T extends Record<string, any>>(
  obj: T,
  shouldOmit: (value: T[keyof T], key: keyof T) => boolean
): Partial<T> {
  for (const key in obj) {
    // or Object.prototype.hasOwnProperty.call
    if (Object.hasOwn(obj, key)) {
      const value = obj[key];
      if (!shouldOmit(value, key)) {
        result[key] = value;
      }
    }
  }

  return result;
}

describe('omitBy', () => {
  const obj = { a: 1, b: 'omit', c: 3 };
  const shouldOmit = (value: number | string) => typeof value === 'string';

  bench('es-toolkit/origin omitBy', () => {
    omitBy1(obj, shouldOmit);
  });

  bench('es-toolkit/new omitBy', () => {
    omitBy2(obj, shouldOmit);
  });
});
```

For accurate testing, I defined and tested all functions in a bench file.

스크린샷 2024-10-06 오후 5 11 12

raon0211:
Is it different from the version of `Object.keys(...)`, rather than `Object.entries(...)`? Based on what I know, `Object.entries(...)` is twice slower that `Object.keys(...)`.

raon0211:
The reason I'm asking is that simplicity in implementation is also a key value for es-toolkit. It makes the code more maintainable and helps reduce the bundle size.

dayongkr:
Screenshot 2024-10-06 at 5 21 18 PM

> ref. https://tc39.es/ecma262/multipage/abstract-operations.html#sec-enumerableownproperties

I think the overhead comes from `Object.entries` because it creates an array. I believe using `Object.keys()`, as suggested by the maintainer, would result in similar performance improvements!

ssi02014:
@raon0211 
There were also performance improvements when using `Object.keys`.

```ts
// origin
function omitBy1<T extends Record<string, any>>(
  obj: T,
  shouldOmit: (value: T[keyof T], key: keyof T) => boolean
): Partial<T> {

  const objEntries = Object.entries(obj);
  for (let i = 0; i < objEntries.length; i++) {
    const [key, value] = objEntries[i];
    if (!shouldOmit(value, key)) {
      (result as any)[key] = value;
    }
  }

  return result;
}

// new
function omitBy2<T extends Record<string, any>>(
  obj: T,
  shouldOmit: (value: T[keyof T], key: keyof T) => boolean
): Partial<T> {

  const objectKeys = Object.keys(obj);
  for (let i = 0; i < objectKeys.length; i++) {
    const key = objectKeys[i];
    const value = obj[key];
    if (!shouldOmit(obj[key], key)) {
      (result as any)[key] = value;
    }
  }

  return result;
}

describe('omitBy', () => {
  const obj = { a: 1, b: 'omit', c: 3 };
  const shouldOmit = (value: number | string) => typeof value === 'string';

  bench('es-toolkit/origin omitBy', () => {
    omitBy1(obj, shouldOmit);
  });

  bench('es-toolkit/new omitBy', () => {
    omitBy2(obj, shouldOmit);
  });
});
```

스크린샷 2024-10-06 오후 5 21 41

ssi02014:
@raon0211 @dayongkr @D-Sketon 
I've seen performance improvements in `Object.keys` as well, so would you agree to change to `Object.keys` instead of `for...in` statements? I can work on that.

raon0211:
The `Object.keys(...)` approach looks good to me.

ssi02014:
@raon0211 Thanks for the feedback, I changed it to `Object.keys`.

src/object/omitBy.ts
```diff
@@ -23,9 +23,10 @@ export function omitBy<T extends Record<string, any>>(
 ): Partial<T> {
   const result: Partial<T> = {};
 
-  const objEntries = Object.entries(obj);
-  for (let i = 0; i < objEntries.length; i++) {
-    const [key, value] = objEntries[i];
+  const objectKeys = Object.keys(obj);
+  for (let i = 0; i < objectKeys.length; i++) {
+    const key = objectKeys[i];
+    const value = obj[key];
```
raon0211, line 29:
```suggestion
  const keys = Object.keys(obj);
  
  for (let i = 0; i < keys.length; i++) {
    const key = keys[i];
    const value = obj[key];
```

raon0211 (approved):
Thanks for your help!

PR #673, opened 2024-10-06
perf(pickBy): improve pickBy performance

ssi02014:
https://github.com/toss/es-toolkit/pull/671

This is the same as the PR above.
Change from `Object.Entries` to `Object.keys`.

### DIFF
--- a/src/object/omitBy.ts
+++ b/src/object/omitBy.ts
@@ -24,7 +24,6 @@ export function omitBy<T extends Record<string, any>>(
   const result: Partial<T> = {};
 
   const keys = Object.keys(obj);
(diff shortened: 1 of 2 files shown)


src/object/omitBy.ts
```diff
@@ -24,7 +24,6 @@ export function omitBy<T extends Record<string, any>>(
   const result: Partial<T> = {};
 
   const keys = Object.keys(obj);
-
```
ssi02014, line 27: A lint error occurs and you fix it. 😭

raon0211 (approved):
Thanks! 😭

PR #675, opened 2024-10-07
feat(compat): Add `isDate` in compat

chhw130:
I added `isDate` in compat layer.

### DIFF
--- a/benchmarks/performance/isDate.bench.ts
+++ b/benchmarks/performance/isDate.bench.ts
@@ -1,8 +1,10 @@
 import { bench, describe } from 'vitest';
 import { isDate as isDateToolkit_ } from 'es-toolkit';
+import { isDate as isDateToolkitCompat_ } from 'es-toolkit/compat';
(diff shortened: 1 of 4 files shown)


ssi02014:
👋 Hi! Why not separate the issue of resolving the `isEqualWith.spec.ts` error into a different pull request?

chhw130:
> 👋 @chhw130 Hi! Why not separate the issue of resolving the `isEqualWith.spec.ts` error into a different pull request?

You're right, that makes sense. I thought it was a simple issue so I included it in this PR.
I'll create a separate PR for the `isEqualWith.spec.ts` issue. Thanks for the suggestion!

ssi02014:
@chhw130 Cool! 👍 I suggested this because I was concerned that bundling it with the `isDate` task would delay the resolution of the `isEqualWith.spec.ts` error issue for review time. Thank you for your positive response.

raon0211 (approved):
Thanks!

PR #676, opened 2024-10-07
refac(unionBy): simplify unionBy implementation logic

ssi02014:
By utilizing `uniqBy`, we can implement unionBy a little more simply.

In the case of `union`, it is implemented as follows 

```ts
export function union<T>(arr1: readonly T[], arr2: readonly T[]): T[] {
  return uniq(arr1.concat(arr2));
}
```

### DIFF
--- a/src/array/unionBy.ts
+++ b/src/array/unionBy.ts
@@ -1,3 +1,5 @@
+import { uniqBy } from './uniqBy';
+
 /**
  * Creates an array of unique values, in order, from all given arrays using a provided mapping function to determine equality.
  *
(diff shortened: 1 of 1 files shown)


ssi02014:
typecheck error related: https://github.com/toss/es-toolkit/pull/675/files

src/array/unionBy.ts
```diff
@@ -1,3 +1,5 @@
+import { uniqBy } from './uniqBy';
```
D-Sketon, line 1:
```suggestion
import { uniqBy } from './uniqBy.ts';
```

ssi02014: @D-Sketon Thanks 😊

PR #679, opened 2024-10-07
docs(difference): add Lodash Compatibility

D-Sketon:


raon0211 (approved):
Thanks for your help!

PR #680, opened 2024-10-07
fix(difference): support array-like object

D-Sketon:
the document will be updated in https://github.com/toss/es-toolkit/pull/679


D-Sketon:
looks like most methods in lodash support array-like object

PR #681, opened 2024-10-07
feat(template): add `template` in compat layer

### LINKED ISSUE #672: Add supports for `template`
At first, i'll add only in compat layer

related #352

dayongkr:
# Description

- The source of the function compiled using the template function has been updated to use ES6+ syntax for greater clarity. For example,  `+\n((__t = (" + interpolateValue + ")) == null ? '' : __t` was replaced with  `+ (${interpolateValue} ?? '')`.
  - However, with the use of ES6+ syntax, runtime environments that do not support these features may encounter errors.
- We ensure that modules remain unmodifiable when bundled as both CJS and ESM. However, there were some test cases that modified these modules. Therefore, those parts have been commented out.

## Benchmarks

Screenshot 2024-10-08 at 2 29 05 PM

### Template Functionality:

* [`src/compat/string/template.ts`](diffhunk://#diff-f1684529e1c9c19a9b0f045ce193bdf791a691df011ad737ec5a1ea9bf32f5d9R1-R166): Implemented the `template` function to create compiled template functions that can interpolate, evaluate, and escape data properties based on provided options.
* [`src/compat/string/templateSettings.ts`](diffhunk://#diff-98b26e6365abfa4e4a3d72e2c11d00510a875686beac67d4394ffbeb9cb7a073R1-R11): Defined default settings for the `template` function, including escape, evaluate, and interpolate delimiters, and imported functions.

### Miscellaneous:

* [`src/compat/_internal/numberTag.ts`](diffhunk://#diff-4fc4d02693eef8e3c9f24ba7e35f23b7bdca36d53365a34a67a696b176f72cc2R1): Added a constant `numberTag` for internal use.
* [`src/compat/index.ts`](diffhunk://#diff-919544eb15e4bb20ce2371a9a8837aa523de3213cb81589129ecf00414f9e8cbR144-R145): Exported the `template` function and `templateSettings` from the string utilities module.

close #672


dayongkr:
I refactored `String.prototype.replace` to use `String.prototype.matchAll`, making the code more aligned with its intended purpose.

raon0211:
I think changing `+\n((__t = (" + interpolateValue + ")) == null ? '' : __t` into  `+ (${interpolateValue} ?? '')` would break in legacy environments, and we cannot transpile or polyfill the behavior. So I would like to stick with the original source code string; what are your thoughts on this?

dayongkr:
> I think changing `+\n((__t = (" + interpolateValue + ")) == null ? '' : __t` into `+ (${interpolateValue} ?? '')` would break in legacy environments, and we cannot transpile or polyfill the behavior. So I would like to stick with the original source code string; what are your thoughts on this?

I implemented it that way under the assumption that legacy environments are not considered. However, if legacy environments are to be taken into account, it would be better to implement it like Lodash. I’ll revise it considering compatibility.

Thank you!

raon0211:
Yeah, we sure support only environments over Node.js 18+, I think even in legacy environments, most things should work when using polyfills or after transpilation. However, since `template` has source code in strings, I think they won't work in legacy environments even after transpilation.

dayongkr:
> Yeah, of course we should support only environments over Node.js 18+.
> 
> However, I most things should work in legacy environments after using polyfills or after transpilation.
> 
> However, since `template` has raw source code as strings, I think they won't work in legacy environments even after transpilation.

I agree with your opinion. I have modified the problematic section and the part that uses the spread operator!

raon0211 (approved):

PR #683, opened 2024-10-07
feat(escape): add `escape` in compat layer

dayongkr:
# Description

I added `escape` in compat layer.

I needed it for implementing `template` method.

unblocks #681 

## Benchmark
Screenshot 2024-10-07 at 10 59 19 PM


raon0211 (approved):
Thanks!

PR #685, opened 2024-10-07
docs(isBlob): Add missing example

mass2527:
- Synchronized example of `isBlob`


raon0211 (approved):
Thanks for finding this!

PR #686, opened 2024-10-07
fix(drop): support array-like args

D-Sketon:
After this PR, dropWhile and dropRightWhile are still a bit different from lodash:
lodash: 
```ts
function dropWhile<T>(
  arr: ArrayLike<T> | null | undefined,
  canContinueDropping: (item: T, index: number, arr: ArrayLike<T>) => unknown
): T[];
```
es-toolkit: 
```ts
function dropWhile<T>(
  arr: ArrayLike<T> | null | undefined,
  canContinueDropping: (item: T, index: number, arr: T[]) => unknown
): T[];
```
The third argument to `canContinueDropping` in lodash is the same as `arr`, but the third argument to `canContinueDropping` in es-toolkit is `Array.from(arr)`.
If we want to modify it, we can't use `dropWhileToolkit`/`dropRightWhileToolkit` in the compat layer, because it doesn't support ArrayLike.

### DIFF
--- a/benchmarks/performance/dropRight.bench.ts
+++ b/benchmarks/performance/dropRight.bench.ts
@@ -1,13 +1,19 @@
 import { bench, describe } from 'vitest';
 import { dropRight as dropRightToolkit_ } from 'es-toolkit';
+import { dropRight as dropRightToolkitCompat_ } from 'es-toolkit/compat';
 import { dropRight as dropRightLodash_ } from 'lodash';
 
 const dropRightToolkit = dropRightToolkit_;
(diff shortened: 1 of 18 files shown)


raon0211:
Great work! Meanwhile, isn't Array<T> compatible with ArrayLike<T> (since it only has additional methods, etc)? 

The whole point of es-toolkit/compat is to ensure that the existing code does not break when we replace lodash with es-toolkit; I think providing additional features is acceptable.

raon0211 (approved):
Thanks! Modified the implementation a little bit to reduce type assertions.

Issue #668, opened 2024-10-04
Confusing example for throttle function with { trailing: true } option


wondonghwi:
The documentation example for the throttle function when using the { trailing: true } option contains an incorrect comment. The comment states that the function is "logged immediately," which is inaccurate. When using trailing: true, the function should not be executed immediately but rather after the throttleMs delay period has passed since the last call.

![image](https://github.com/user-attachments/assets/112fc9ef-1dcd-4d3f-a5d1-4d936f09ac6e)

**Problematic Example:**
```
const trailingFn = throttle(
  () => {
    console.log('Trailing function executed');
  },
  1000,
  { trailing: true }
);

// The comment below is incorrect:
console.log('Trailing function executed');

// Logs 'Trailing function executed' immediately
// Continuously calls will log 'Trailing function executed' every 1 second
trailingFn();
```

**Correction:**
```
const trailingFn = throttle(
  () => {
    console.log('Trailing function executed');
  },
  1000,
  { trailing: true }
);

// Should log after the throttle period, not immediately
// Logs 'Trailing function executed' after 1 second since the last call
trailingFn();
```

**If this comment is not intended, may I modify the comment in both the English and Korean documentation to reflect the correct behavior?**

raon0211:
Actually for `throttle`, the function should be invoked immediately after the first call. Is it not working in that way? Can you provide an example code?

end-jarry:
@raon0211 
Thank you for your response.

Based on the documentation description of the edges option for throttle, I believe there may be some misunderstanding regarding the behavior of { trailing: true }.

As stated in the documentation:

If leading is included, the throttled function is executed immediately when first called.
If trailing is included, the function is executed after the throttle delay (throttleMs) has passed since the last function call.
Given this explanation, I believe the example comment may cause confusion. The current comment suggests that the function will be "logged immediately," but this is true only when { leading: true } is set. In the case of { trailing: true }, the function should only be executed after the delay, not immediately.

Could you please clarify if this is the intended behavior? If so, it would be helpful to update the comment in both the English and Korean documentation to prevent any further misunderstanding.

I appreciate your help in addressing this!

<img width="705" alt="image" src="https://github.com/user-attachments/assets/af65fd77-c061-4ca2-97c9-17ed4053ea85">

wondonghwi:
@raon0211 
Thank you for your response.

Based on the documentation description of the edges option for throttle, I believe there may be some misunderstanding regarding the behavior of { trailing: true }.

[...]

PR #688, opened 2024-10-08
docs(throttle): Update docs

wondonghwi:
#668 

I have added documentation for the { leading: false, trailing: true } option in throttle, as it could cause confusion based on how Lodash handles it.


raon0211 (approved):
Thanks!

PR #692, opened 2024-10-08
fix(isPlainObject): Fix isPlainObject and perf 3x+

belgattitude:
Hello, I'm really impressed by estoolkit. Great !

I'm porting few more tests and fixes to the isPlainObject implementation. Mostly taken from my own isPlainObject implementation: see https://github.com/belgattitude/httpx/tree/main/packages/plain-object#readme

This fix few bugs that I found. So I've completed the test suite. 

## Fixes

- #694 
- #693 
- #695 

### "Realworld" scenario

Should also bring a >3x speedup based on my "real-world" scenarios

image

https://github.com/belgattitude/httpx/tree/main/packages/plain-object#benchmarks

### estoolkit bench

After the changes
```
 ✓ benchmarks/performance/isPlainObject.bench.ts > isPlainObject 3920ms
     name                                       hz     min     max    mean     p75     p99    p995    p999     rme  samples
   · es-toolkit/isPlainObject         2,519,342.19  0.0004  0.0325  0.0004  0.0004  0.0007  0.0009  0.0017  ±0.06%  1259672
   · es-toolkit/compat/isPlainObject  3,528,357.44  0.0003  0.2412  0.0003  0.0003  0.0005  0.0007  0.0014  ±0.12%  1764179   fastest
   · lodash/isPlainObject               692,780.40  0.0014  0.3287  0.0014  0.0014  0.0021  0.0026  0.0044
```

Before
```
✓ benchmarks/performance/isPlainObject.bench.ts > isPlainObject 3842ms
     name                                       hz     min     max    mean     p75     p99    p995    p999     rme  samples
   · es-toolkit/isPlainObject         2,496,468.28  0.0004  0.0388  0.0004  0.0004  0.0008  0.0010  0.0018  ±0.07%  1248235
   · es-toolkit/compat/isPlainObject  3,137,159.94  0.0003  0.3087  0.0003  0.0003  0.0008  0.0010  0.0019  ±0.20%  1568580   fastest
   · lodash/isPlainObject               702,054.71  0.0013  0.0356  0.0014  0.0014  0.0019  0.0025  0.0044  ±0.09%   351028   slowest
```

## Tasks

- [x] Improve test suite in https://github.com/toss/es-toolkit/pull/692/commits/ab76ab3b71b023e41dcb5a41b63492e1a8350641
- [x] Fix bugs: iterator symbol, anonymous classes, runInNewContext (see fixes below)
- [x] Refactor isPlainObject function

## Notes

> See the updated test suite in: https://github.com/toss/es-toolkit/pull/692/commits/ab76ab3b71b023e41dcb5a41b63492e1a8350641

- [x] If an object has an iterator symbol, it shouldn't be a plain-object (open to discuss ?)

image

- [x]  When passing an anonymous class it should not fail with TypeError 

image

- [x] Should support cross realms (node:vm runInNewContext)

image

### DIFF
--- a/src/predicate/isPlainObject.spec.ts
+++ b/src/predicate/isPlainObject.spec.ts
@@ -1,38 +1,49 @@
-import { describe, expect, it } from "vitest";
-import { isPlainObject } from "./isPlainObject";
-import { runInNewContext } from "node:vm";
+import { describe, expect, it } from 'vitest';
+import { runInNewContext } from 'node:vm';
+import { isPlainObject } from './isPlainObject';
 
-describe("isPlainObject", () => {
-  it("should return true for plain objects", () => {
-    const str = "key";
+describe('isPlainObject', () => {
+  it('should return true for plain objects', () => {
+    const str = 'key';
 
     expect(isPlainObject({})).toBe(true);
     expect(isPlainObject(Object.create(null))).toBe(true);
-    expect(isPlainObject(new Object({ key: "new_object" }))).toBe(true);
+    expect(isPlainObject(new Object({ key: 'new_object' }))).toBe(true);
     expect(isPlainObject(new Object({ key: new Date() }))).toBe(true);
-    expect(isPlainObject({ 1: "integer_key" })).toBe(true);
-    expect(isPlainObject({ name: "string_key" })).toBe(true);
-    expect(isPlainObject({ [str]: "dynamic_string_key" })).toBe(true);
-    expect(isPlainObject({ [Symbol("tag")]: "value" })).toBe(true);
+    expect(isPlainObject({ 1: 'integer_key' })).toBe(true);
+    expect(isPlainObject({ name: 'string_key' })).toBe(true);
+    expect(isPlainObject({ [str]: 'dynamic_string_key' })).toBe(true);
+    expect(isPlainObject({ [Symbol('tag')]: 'value' })).toBe(true);
     expect(
       isPlainObject({
-        children: [{ key: "deep-children" }],
-        name: "deep-plain",
-      }),
+        children: [{ key: 'deep-children' }],
+        name: 'deep-plain',
+      })
     ).toBe(true);
     expect(
       isPlainObject({
         children: [{ key: new Date() }],
-        name: "deep-with-regular-object",
-      }),
+        name: 'deep-with-regular-object',
+      })
     ).toBe(true);
-    expect(isPlainObject({ constructor: { name: "Object2" } })).toBe(true);
-    expect(isPlainObject(JSON.parse("{}"))).toBe(true);
+    expect(isPlainObject({ constructor: { name: 'Object2' } })).toBe(true);
+    expect(isPlainObject(JSON.parse('{}'))).toBe(true);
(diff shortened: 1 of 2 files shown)


belgattitude:
Tests passes on my laptop... node 20.18.0, seems the CI doesn't

image

belgattitude:
This PR should make it compatible (drop in replacement) with

- https://github.com/sindresorhus/is-plain-obj
- https://github.com/belgattitude/httpx/tree/main/packages/plain-object#readme

raon0211:
Hey, thanks for your contribution! I really appreciate your implementation—it’s much more modern and concise.

I made a few minor edits:

- I noticed that we check for `Symbol.toStringTag` to handle static objects like `JSON`, `Math`, and `Atomics`. I think we should consider them plain objects since they’re not instances of any class; they just aggregate some functions.
- Similarly, when checking for `Symbol.iterator` to handle `arguments`, I believe it should also be seen as a plain object. It is not an array, and mainly has plain properties like `0`, `1`, and `Symbol.iterator`.

Lastly, while these additional checks add complexity, they also increase the bundle size and may impact performance, even though they address some edge cases.

raon0211 (approved):
Thanks for your contribution!

belgattitude:
> I noticed that we check for Symbol.toStringTag to handle static objects like JSON, Math, and Atomics. I think we should consider them plain objects since they’re not instances of any class; they just aggregate some functions.

Thx @raon0211, I've tested the change on my repo. Not sure about static objects, imho they shouldn't be considered as plain objects.

Without the checks for static object, we get rid off 30 bytes (before - 101b -> 75b brotli/webpack)... The performance won't change a lot as the checks are done last. (I've benchmarked on my repo, and the impact isn't measurable). 

I'm happy with both ways, but I'd recommend to keep support for Atomics in general for "correctness" (if it actually exist :))

belgattitude:
> Similarly, when checking for Symbol.iterator to handle arguments, I believe it should also be seen as a plain object. It is not an array, and mainly has plain properties like 0, 1, and Symbol.iterator.

That one I tend to agree (without this test the bundle would go from 101b -> 83b: in the scenario we keep toStringTag for atomics)

raon0211:
Our team has been discussing the definition of `isPlainObject`. It seems the common understanding of plain objects in JavaScript includes the following points:

- They behave like typical object literals, such as `{ name: 'John' }`.
- Their prototype is either `Object.prototype` or `null`.
- When stringified using `toString`, they return `[object Object]`.

In light of this, we've updated our [documentation on plain objects](https://es-toolkit.slash.page/reference/predicate/isPlainObject.html) and revised the implementation accordingly. You can check out the changes [here](https://github.com/toss/es-toolkit/commit/755d9db62c3435ce6f2b85cfbf92ca96bde572e5).

We chose not to check for `Symbol.toStringTag` or `Symbol.iterator` for the following reasons:

- Not all internal objects have a `Symbol.toStringTag`. For instance, `Symbol.toStringTag in []` evaluates to false, even though `[].toString() === '[object Array]'`.
- There are valid plain objects that can include properties like `{ [Symbol.iterator]: function *() {} }`.

PR #700, opened 2024-10-09
docs(isPlainObject): Update example of korean documentation

mass2527:
Synchronized Korean documentation example with other languages.


raon0211 (approved):
Thanks!

PR #701, opened 2024-10-09
fix(chunk/compact/head/initial/last/tail/take/takeRight/uniq/without): fix lodash compatibility

D-Sketon:
Make the type definitions of the compat layer fully compatible with Lodash


raon0211 (approved):
Looks nice. Thanks for your contribution!

PR #704, opened 2024-10-11
feat(isFile): add `isFile`

lukaszkowalik2:
Hey, 
I've added a new `isFile` function. I would appreciate it if someone could help correct the documentation in Korean, Japanese, and Simplified Chinese.


docs/ko/reference/predicate/isFile.md
```diff
@@ -0,0 +1,31 @@
+# isFile
+
+주어진 값이 `File`인지 확인합니다.
```
raon0211, line 3:
```suggestion
주어진 값이 `File`인지 확인해요.
```

raon0211 (approved):
Thanks!

PR #707, opened 2024-10-13
fix(every/fill/filter/find/findIndex/findLastIndex/indexOf/join): fix lodash compatibility

D-Sketon:
add `ArrayLike<T> | null | undefined` for every/fill/filter/find/findIndex/findLastIndex/indexOf/join
add `fromIndex` for findIndex


benchmarks/performance/every.bench.ts
```diff
@@ -1,5 +1,5 @@
 import { bench, describe } from 'vitest';
-import { every as everyEsToolkit_ } from 'es-toolkit';
+import { every as everyEsToolkit_ } from 'es-toolkit/compat';
```
raon0211, line 2:
```suggestion
import { every as everyEsToolkit_ } from 'es-toolkit';
```

docs/reference/compat/array/every.md
```diff
@@ -18,24 +18,30 @@ You can specify the condition in several ways:
 ## Signature
 
 ```typescript
-function every<T>(arr: T[]): boolean;
-function every<T>(arr: T[], doesMatch: (item: T, index: number, arr: T[]) => unknown): boolean;
-function every<T>(arr: T[], doesMatch: Partial<T>): boolean;
-function every<T>(arr: T[], doesMatch: [keyof T, unknown]): boolean;
-function every<T>(arr: T[], doesMatch: string): boolean;
+function every<T>(arr: ArrayLike<T> | null | undefined): boolean;
```
raon0211, line 21: I'm actually not 100% sure if we should explicitly show that we accept `null` or `undefined` for `every` or other functions here in the docs. The function signature looks much verbose, and it is hard to grasp the essentials. 

What about we stick with the original function signature in the docs, and add a special info section like:

```md
 function every<T>(arr: T[]): boolean;
 function every<T>(arr: T[], doesMatch: (item: T, index: number, arr: T[]) => unknown): boolean;
 function every<T>(arr: T[], doesMatch: Partial<T>): boolean;
 function every<T>(arr: T[], doesMatch: [keyof T, unknown]): boolean;
 function every<T>(arr: T[], doesMatch: string): boolean;

 function every<T extends Record<string, unknown>>(
   object: T,
   doesMatch: (value: T[keyof T], key: keyof T, object: T) => unknown
 ): boolean;
 function every<T extends Record<string, unknown>>(object: T, doesMatch: Partial<T[keyof T]>): boolean;
 function every<T extends Record<string, unknown>>(object: T, doesMatch: [keyof T, unknown]): boolean;
 function every<T extends Record<string, unknown>>(object: T, doesMatch: string): boolean;

::: info `arr` can be `ArrayLike<T>`, `null`, or `undefined`

The `every` function is designed for complete compatibility with lodash:

- If arr is an `ArrayLike<T>`, it gets converted into an array using `Array.from(...)`.
- If arr is `null` or `undefined`, it will be treated as an empty array.

:::

::: info `object` can be `null` or `undefined`

To ensure full compatibility with lodash, the every function handles `object` in this way:

- If `object` is `null` or `undefined`, it will be converted into an empty object.

:::
```

D-Sketon:
done

raon0211 (approved):
Great work. Thanks so much!

Issue #699, opened 2024-10-09
Support for `sample`


Gyumong:
Hello, @raon0211 
I’d like to contribute by working on the sample function if it’s available.

Currently, the `sample` function in es-toolkit only supports arrays. To improve compatibility with Lodash and enhance its functionality, I propose the following updates

1. Add support for object sampling
2. Handle empty collections
3. Handle `null` or `undefined` inputs

Gyumong:
I plan to ensure that only compat support objects

D-Sketon:
lodash also support arraylike
for example _.sample('123') is legal

Gyumong:
Thank you for your helpful feedback :)

raon0211:
Great work. Thanks!

PR #708, opened 2024-10-13
feat(sample): Add `sample` to compat layer 

Gyumong:
I added `sample` in compat layer 

issue #699 

benchmark

image


docs/ko/reference/compat/array/sample.md
```diff
@@ -0,0 +1,43 @@
+# sample
```
raon0211, line 1: I think we might not need docs for `sample` here, because we actually already have a one ([link](https://es-toolkit.slash.page/reference/array/sample.html)).

src/compat/array/sample.ts
```diff
@@ -0,0 +1,94 @@
+import { sample as sampleToolkit } from '../../array/sample';
+import { isArrayLike } from '../predicate/isArrayLike';
```
raon0211, line 2:
```suggestion
import { sample as sampleToolkit } from '../../array/sample.ts';
import { isArrayLike } from '../predicate/isArrayLike.ts';
```

src/compat/array/sample.ts
```diff
@@ -0,0 +1,108 @@
...
+ * @param {ArrayLike<T> | Record<string, T>} collection - The collection to sample from.
+ * @returns {T | string | undefined} A random element from the collection, or `undefined` if the collection is empty or invalid.
+ */
+export function sample<T>(collection: ArrayLike<T> | Record<string, T>): T | string | undefined {
+  if (collection == null) {
+    return undefined;
+  }
+
+  if (isArrayLike(collection)) {
+    return sampleToolkit(collection as readonly T[]);
```
raon0211, line 103:
```suggestion
    return sampleToolkit(Array.from(collection));
```

raon0211 (approved):
Thanks!

PR #711, opened 2024-10-14
fix(cloneDeep):  Cannot assign to read only property '0' of object '[object String]'

D-Sketon:


D-Sketon:
```ts
it(`should clone expando properties`, () => {
  const values = [false, true, 1, 'a'].map(value => {
    const object = Object(value);
    object.a = 1;
    return object;
  });

  const expected = values.map(stubTrue);

  const actual = values.map(value => {
    return cloneDeep(value).a === 1;
  });

  expect(actual).toEqual(expected);
});


it('should clone string objects', () => {
  const object = Object('a');

  const actual = cloneDeep(object);

  expect(actual).toEqual(object);
  expect(actual).not.toBe(object);
});
```

D-Sketon:
Existing unit tests fail

raon0211:
We should check for `target`'s writable property, so updated the code accordingly. Thanks for pointing this out!

raon0211 (approved):
Thanks!

PR #714, opened 2024-10-15
feat(uniqueId): Add uniqueId to compat layer

Na-hyunwoo:
Implemented lodash's uniqueId. 

If there is anything that needs supplementation, please give me feedback. 

I will fix it right away. 

Thank you.

uniqueId benchmark

### DIFF
--- a/benchmarks/performance/uniqueId.bench.ts
+++ b/benchmarks/performance/uniqueId.bench.ts
@@ -1,0 +1,31 @@
+import { bench, describe } from 'vitest';
+import { uniqueId as uniqueIdToolkitCompat } from 'es-toolkit/compat';
+import { uniqueId as uniqueIdLodash } from 'lodash';
+
+describe('uniqueId', () => {
+  bench('es-toolkit/compat/uniqueId', () => {
+    uniqueIdToolkitCompat();
+    uniqueIdToolkitCompat();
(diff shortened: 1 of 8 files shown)


dayongkr:
The functions in compat layer should pass the lodash’s test code. Please refer to the #732 PR!

Na-hyunwoo:
Hi! @dayongkr 

When defining utility functions in the compat layer, we should follow lodash's test! Thank you for your good comments. 

The uniqueId you created was more readable, so I changed my code a bit. 

And in my opinion, it is better not to verify lodash's third test code because it does not match es-toolkit's design principles. 

What do you think?

Give you ref: [Design Principles](https://es-toolkit.slash.page/compatibility.html#design-principles)
스크린샷 2024-10-22 오전 8 28 04

dayongkr:
> Design Principles

Thank you for reviewing my opinion.

In my opinion, that seems to refer to cases where JavaScript implicitly converts types at the language level(e.g. `'3' + 3 == '3' + '3'`). However, the test case involves explicitly coercing a `number` into a `string`, so I believe it is necessary.

raon0211:
As @dayongkr mentioned, I think we should handle that test case.

raon0211 (approved):
Thanks for your contribution!

PR #716, opened 2024-10-15
fix(flatten/flattenDeep/flattenDepth/slice/zipObjectDeep): fix lodash compatibility

D-Sketon:


raon0211 (approved):
Thanks for your work!

PR #720, opened 2024-10-16
refactor(cloneDeep): Refactoring cloneDeep

ssi02014:
`Map` in a `for...of` statement basically behaves the same as `entries()`.

`Set` in a `for...of` statement basically behaves the same as `values()`.

Therefore, we can write even simpler. Is there a reason we're writing this explicitly?


raon0211 (approved):
You're right. Thanks for fixing this!

PR #721, opened 2024-10-16
feat(intersectionBy): Add `intersectionBy` to compatibility layer

mass2527:
I added the `intersectionBy` function to the compatibility layer.

스크린샷 2024-10-17 오전 12 29 56


raon0211 (approved):
Thanks for your contribution!

PR #728, opened 2024-10-18
fix(some): fix lodash compatibility

D-Sketon:
some should support object


raon0211 (approved):
Looks great! Thanks so much for your contribution!

PR #729, opened 2024-10-18
fix(compat): fix type error

D-Sketon:


docs/ja/reference/compat/array/every.md
```diff
@@ -29,7 +29,7 @@ function every<T extends Record<string, unknown>>(
   doesMatch: (value: T[keyof T], key: keyof T, object: T) => unknown
 ): boolean;
 function every<T extends Record<string, unknown>>(object: T, doesMatch: Partial<T[keyof T]>): boolean;
-function every<T extends Record<string, unknown>>(object: T, doesMatch: [keyof T, unknown]): boolean;
+function every<T extends Record<string, unknown>>(object: T, doesMatch: [keyof T[keyof T], unknown]): boolean;
```
raon0211, line 32: Isn‘t this `T[keyof T]` instead of `keyof T[keyof T]`?

D-Sketon:
```js
const obj = { alice: { id: 1, name: 'Alice' }, bob: { id: 2, name: 'Bob' } };
const result = every(obj, ['name', 'Alice']);
```
`T[keyof T]` means `{ id: 1, name: 'Alice' } | { id: 2, name: 'Bob' }`
`keyof T[keyof T]` means `'name' | 'id'`
What's the problem?

raon0211: Oh, It's `keyof (T[keyof T])`; I think it's my mistake.

raon0211 (approved):

PR #740, opened 2024-10-23
fix(merge): improved merge function types

ssi02014:
Is there a reason not to define the types of arguments to the `merge` function more narrowly? 

Given the purpose of the `merge` function and most use cases, I think it's appropriate to narrow the type of the source and destination to `Object`.

What do you think? 

Currently, the following cases are allowed 
Do we really need to allow this? 🤔
```ts
// as-is
merge(1, 2); // type pass
```


raon0211 (approved):
Looks good for me. Thanks for catching this!

PR #749, opened 2024-10-23
feat(compat): add `forEach` function

### LINKED ISSUE #335: Support for `forEach`
Hello, @raon0211!
If you need `forEach`, Can I work it? 

If you don't mind working with it, in the case of `forEach`, it only applied to arrays, but I understand that compat(lodash) supports some objects. So, I plan to ensure that only compat support objects.

kyvg:
This pull request implements `forEach` function which behaves similarly to the [_.forEach]( https://lodash.com/docs/4.17.15#forEach) function
close #335

image


raon0211:
Let me review this in this weekend :)

raon0211 (dismissed):
Thanks for your contribution!

raon0211 (approved):

PR #752, opened 2024-10-25
fix(toMerged): improved toMerged function types

ssi02014:
https://github.com/toss/es-toolkit/pull/740
Like `merge`, `toMerged` also changes the type.
It also modifies the documentation.


raon0211 (approved):
Thanks for your contribution!

PR #754, opened 2024-10-25
feat(replace): add `replace` to compat layer

dayongkr:
# Description

I added `replace` to compat layer.

It utilizes `String.prototype.replace`, resulting in minimal performance difference compared to Lodash. Additionally, I decided not to include it in the original es-toolkit as it seemed unnecessary.

Furthermore, using a union type in the replacement part causes an error. To circumvent this, I temporarily asserted it as `any`.

> related [Issue](https://github.com/microsoft/TypeScript/issues/22378)

## Benchmark
Screenshot 2024-10-25 at 5 04 52 PM


raon0211 (approved):
Looks great. Thanks!

Issue #715, opened 2024-10-15
Support for findKey, findLastKey


Na-hyunwoo:
Hi, @raon0211 !
Can I work **findKey** of https://github.com/toss/es-toolkit/issues/91 ??

https://github.com/toss/es-toolkit/issues/441
In the above issue, someone said they were implementing it, but it seems like it hasn't been implemented because it's been about 2 months.

And I have a question about the findKey interface.

https://lodash.com/docs/4.17.15#findKey

```
_.findKey(users, function(o) { return o.age < 40; });
// => 'barney' (iteration order is not guaranteed)
 
// The `_.matches` iteratee shorthand.
_.findKey(users, { 'age': 1, 'active': true });
// => 'pebbles'
 
// The `_.matchesProperty` iteratee shorthand.
_.findKey(users, ['active', false]);
// => 'fred'
 
// The `_.property` iteratee shorthand.
_.findKey(users, 'active');
// => 'barney'
```

Looking at the example in the lodash document, you can get the key by entering ['active', false] or 'active' as the second argument to findKey, but this makes the complexity of implementation very difficult.

https://github.com/toss/es-toolkit/blob/main/.github/CONTRIBUTING.md

However, if you look at es-toolkit's design priniciples, it emphasizes simplicity. 

So I think that implementing only the first and second interfaces in the above example fits es-toolkit's priniciples. 

However, if that happens, the interface will be different from lodash, so I think there might be a problem during the switching process in lodash. 

Of these two conflicting issues **principles vs compatibility with lodash, which issue has higher priority?**

raon0211:
Sure :)

What about supporting only functions in our original `es-toolkit` library, and supporting `matches` and `matchesProperty`, etc. in our compatibility library?

You can refer to how we approached this problem in [dropWhile](https://github.com/toss/es-toolkit/blob/main/src/compat/array/dropWhile.ts) or other similar functions.

Na-hyunwoo:
That would work!! 

Thank you for your good comments. 

Let’s implement it by referring to the `dropWhile` example 🙏

PR #755, opened 2024-10-26
feat(findKey): add findKey

Na-hyunwoo:
Related issues: https://github.com/toss/es-toolkit/issues/715

Before implementing findKey in the compat layer, I preemptively worked on es-toolkit's findKey.

Hi!, @raon0211 

There is something that needs your opinion.

There was a point where I was focused while implementing it, and that was how strictly to keep the type. 

In the case of lodash's findKey, if an object is received as the first argument of findKey, the return value is a simple string. 

However, in the case of findKey that I implemented, the key value of the object is specified as the return value. 

Since this is not a compat layer, I think it is right to follow the principles of es-toolkit rather than lodash, so I need your opinion on that.

I believe that in order for TypeScript to be used meaningfully, it is right to strictly manage types. 

As a maintainer, please tell me your opinion!

I have one more question. 

After this work, I would like to implement the compat layer's findKey, findLastKey, and compat layer's findLastKey in that order. Is that okay?

findKey benchmark

### DIFF
--- a/docs/ja/reference/object/findKey.md
+++ b/docs/ja/reference/object/findKey.md
@@ -13,12 +13,12 @@ function findKey<T extends Record<any, any>>(
 
 ### Parameters
 
-- `obj` (`Record<T, K>`): 検索するオブジェクト。
-- `predicate` (`(value: K, key: T, obj: Record<T, K>) => boolean`): オブジェクト内の各値に対して実行する関数。
+- `obj` (`T extends Record<any, any>`): 検索するオブジェクト。
+- `predicate` (`(value: T[keyof T], key: keyof T, obj: T) => boolean`): オブジェクト内の各値に対して実行する関数。
 
 ### Returns
 
-(`T | undefined`): 指定されたテスト関数を満たすオブジェクト内の最初の要素のキー。テストに合格する要素がない場合は未定義です。
+(`keyof T | undefined`): 指定されたテスト関数を満たすオブジェクト内の最初の要素のキー。テストに合格する要素がない場合は未定義です。
 
 ## Examples
 
--- a/docs/ja/reference/object/findKey.md
+++ b/docs/ja/reference/object/findKey.md
@@ -5,7 +5,10 @@
 ## Signature
 
 ```typescript
-function findKey<T, K>(obj: Record<T, K>, predicate: (value: K, key: T, obj: Record<T, K>) => boolean): T | undefined;
+function findKey<T extends Record<any, any>>(
+  obj: T,
+  predicate: (value: T[keyof T], key: keyof T, obj: T) => boolean
+): keyof T | undefined;
 ```
 
 ### Parameters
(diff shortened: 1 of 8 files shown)


docs/ja/reference/object/findKey.md
```diff
@@ -0,0 +1,31 @@
+# findKey
```
raon0211, line 1: We actually generate documents in an automated way; you don't have to manually create them.

src/object/findKey.ts
```diff
@@ -0,0 +1,24 @@
...
+ *   'barney':  { 'age': 36, 'active': true },
+ *   'fred':    { 'age': 40, 'active': false },
+ *   'pebbles': { 'age': 1,  'active': true }
+ * };
+ * findKey(users, function(o) { return o.age < 40; }); => 'barney'
+ */
+export function findKey<T extends PropertyKey, K>(
+  obj: Record<T, K>,
+  predicate: (value: K, key: T, obj: Record<T, K>) => boolean
+): T | undefined {
```
raon0211, line 22:
```suggestion
export function findKey<T extends Record<any, any>>(
  obj: T,
  predicate: (value: T[keyof T], key: keyof T, obj: T) => boolean
): keyof T | undefined {
```

I guess this type signature is more accurate than the original one. 

---

Note that `foo` is inferred as `number | string` in the first one, but is accurately inferred as `number` in the second one.

### Old interface

```ts
declare function findKey<T extends PropertyKey, K>(
  obj: Record<T, K>,
  predicate: (value: K, key: T, obj: Record<T, K>) => boolean
): T | undefined

findKey({ foo: 1, bar: '1' }, (val, key, obj) => {
  // foo is string | number here
  obj.foo
})
```

image

### New interface

```ts
declare function findKey<T extends Record<any, any>>(
  obj: T,
  predicate: (value: T[keyof T], key: keyof T, obj: T) => boolean
): T | undefined

findKey({ foo: 1, bar: '1' }, (val, key, obj) => {
  // foo is number here
  obj.foo
})
```

image

docs/ja/reference/object/findKey.md
Na-hyunwoo: @raon0211 Oh, I see... I guess I won't have to write docs from next time.

src/object/findKey.ts
Na-hyunwoo: @raon0211 

Wow, thank you for the kind review. 

In the case of the old interface, the type inference of the object's key and value is inferred independently, so accurate type inference is not possible. 

I will commit again, reflecting your review!!

raon0211:
> I would like to implement the compat layer's findKey, findLastKey, and compat layer's findLastKey in that order. Is that okay?

This will help a lot! 😃

raon0211 (approved):
Thanks for your contribution! ☺️

raon0211 (dismissed):

PR #759, opened 2024-10-26
feat(sum, sumBy): add `sum` and `sumBy` to compat layer

dayongkr:
# Overview

I added `sum` and `sumBy` to compat layer.

The type has been further strengthened for `BigInt`.

## Benchmark

Screenshot 2024-10-27 at 12 54 27 PM


raon0211 (approved):
Thanks!

PR #763, opened 2024-10-26
feat(iteratee): add `iteratee` to compat layer

### LINKED ISSUE #760: Support for `iteratee` in compat layer
It seems necessary to implement it since the iteratee function is used across multiple functions in Lodash.

dayongkr:
# Description

I added `iteratee` to compat layer.

```typescript
  it('should use internal `iteratee` if external is unavailable', () => {
    const iteratee = _.iteratee;
    delete _.iteratee;

    expect(map([{ a: 1 }], 'a')).toEqual([1]);

    _.iteratee = iteratee;
  });
```

The above test case has been commented out because it needs to be implemented in the `map`.

## Benchmark

Screenshot 2024-10-26 at 4 45 38 PM

block #759
close #760


raon0211 (approved):
Thanks for your contribution!

PR #766, opened 2024-10-27
feat(sum,sumBy): add a feature handling `bigint` values to original library

### LINKED ISSUE #742: sum should allow bigint
Currently, sum() only accepts number array. Sum with bigint array will throw type error

dayongkr:
# Overview

I added the feature to original library.

If the array is empty, it is impossible to distinguish between `bigint` and `number`, so I implemented it to always return `0` when the array is empty.

Therefore, in the overloading signature of `sumBy`, it is specified that if the return type of `getValue` is `bigint`, the function may return either a `bigint` or a `number`.

close #742

### DIFF
--- a/docs/ja/reference/math/sum.md
+++ b/docs/ja/reference/math/sum.md
@@ -4,24 +4,29 @@
 
 この関数は数値配列を受け取り、配列のすべての要素を足し合わせた合計を返します。
 
+配列が空の場合、この関数は `0` を返します。
+
+配列に `bigint` 値が含まれている場合、関数は `bigint` 値を返します。
+
 ## インターフェース
 
 ```typescript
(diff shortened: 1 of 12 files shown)


dayongkr:
After discussing it in the Discord channel, we decided not to support `BigInt` in our sum and sumBy functions.

Instead, we’ll implement separate functions, such as `sumBigInt` and `sumBigIntBy`.

Here’s why:

We want to maintain properties like `sum([a,b,c]) + sum([d,e,f]) === sum([a,b,c,d,e,f])`.
However, with `BigInt` support, this property would break.
For example, `sum([]) + sum([1n,2n])` would throw an error, while `sum([...[], ...[1n,2n]])` would return `3n`, thus violating the intended property.

PR #776, opened 2024-10-31
fix(replace): Handle cases with less than 3 arguments

mass2527:
I added an [edge case for the argument length of less than 3](https://github.com/lodash/lodash/blob/ddfd9b11a0126db2302cb70ec9973b66baec0975/lodash.js#L14540) to ensure compatibility with Lodash.


raon0211 (approved):
Thanks!

PR #779, opened 2024-11-01
docs: fix es-toolkit cdn version latest

cruelladevil:
## Overview

Some translation for usage.md docs describe cdn serving version for `@^1`.
This PR matches the versions `latest`.

### DIFF
--- a/docs/ja/usage.md
+++ b/docs/ja/usage.md
@@ -65,14 +65,14 @@ bun add es-toolkit
 ::: code-group
 
 ```html [jsdelivr]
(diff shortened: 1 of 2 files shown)


raon0211:
Actually this was intended. `es-toolkit` follows semantic versioning, and we will make a breaking change and release a new version `2.x` some time. If websites use `latest` version, they will break after we make a breaking change.

raon0211:
Thanks! Meanwhile I think we should encode caret(`^`)s to `%5E` in URLs.

cruelladevil:
Yes! Thank you for pointing it out!

raon0211 (approved):
Thanks!

PR #780, opened 2024-11-01
test(math): add test cases related to the associative property for `sum` and `sumBy`

dayongkr:
I have added the parts discussed on Discord to the test cases and documented the relevant content.

I included this as it seemed like helpful information for users of the function.

related #766


raon0211:
I think it’s fine to just add the test cases and skip the documentation, since the goal is to understand how the functions behave. While documentation is important, I prefer to keep it concise.

dayongkr:
> I think it’s fine to just add the test cases and skip the documentation, since the goal is to understand how the functions behave. While documentation is important, I prefer to keep it concise.

I documented it with the intention of providing reassurance to the user.
However, if the purpose of documentation is to clearly demonstrate the behavior, it may be better to include this information only in the test cases in the documentation!

raon0211 (approved):
Thanks!

PR #781, opened 2024-11-01
refactor(range): simplify the range function

ssi02014:
We can simplify the implementation logic a bit more by setting `default parameters` for the `step`! 🤗


D-Sketon:
The same goes for rangeRight.

ssi02014:
@D-Sketon Thanks. I'll fix the `rangeRight` function soon as well.

raon0211 (approved):
Thanks!

PR #783, opened 2024-11-02
feat(isSubsetWith): update isSubset for object array support

### LINKED ISSUE #767: Improve `isSubset` to Handle Object Comparisons Correctly
### Description
[`isSubset`](https://github.com/toss/es-toolkit/blob/main/src/array/isSubset.ts) fails when `superset` and `subset` arrays contain objects. It compares references instead of properties, leading to incorrect results.

#### Example
```ts
const superset = [
  {
    id: 1,
  },
  {
    id: 2,
  },
  {
    id: 3,
  },
];
const subset = [
  {
    id: 1,
  },
  {
    id: 2,
  },
];

console.log(isSubset(superset, subset)); // Expected: true, Actual: false
```

### Suggested Improvement

To resolve this issue, I propose modifying `isSubset` to compare object properties instead of references. The updated implementation could look like this.

```ts
import { isEqual } from '../predicate/isEqual.ts';

export function isSubset<T>(superset: readonly T[], subset: readonly T[]): boolean {
  return subset.every(subsetItem => superset.some(supersetItem => isEqual(subsetItem, supersetItem)));
}
```

This change ensures that the function correctly identifies subsets of objects by comparing their properties.

kaehehehe:
close #767

I've made it so that we can use IsSubset with object arrays.
Here are the benchmark results.

스크린샷 2024-11-02 19 13 07


raon0211:
For this use case, we should create a new function called `isSubsetWith`. Users can then use it like this:

```tsx
isSubsetWith(arr1, arr2, (x, y) => isEqualWith(x, y));
```

This approach aligns well with other functions like `difference` and `union`.

kaehehehe:
I think that makes sense. I went ahead and implemented the isSubsetWith function! :)

raon0211 (approved):
Thanks for your contribution! 😃

PR #785, opened 2024-11-03
refactor(rangeRight): simplify the rangeRight function

ssi02014:
https://github.com/toss/es-toolkit/pull/781 This is the same thing !


raon0211 (approved):
Thanks for refactoring!

PR #787, opened 2024-11-03
refactor(isSubsetWith): improved by reusing the differenceWith in the isSubsetWith

ssi02014:
Related Pull Request: https://github.com/toss/es-toolkit/pull/783
The `isSubset` function is simply implemented by reusing the `difference` function.

```ts
export function isSubset<T>(superset: readonly T[], subset: readonly T[]): boolean {
  return difference(subset, superset).length === 0;
}
```

I think the same is true for the `isSubsetWith` function. It can be implemented simply by reusing `differenceWith`.

cc. @kaehehehe

### DIFF
--- a/src/array/isSubsetWith.ts
+++ b/src/array/isSubsetWith.ts
@@ -1,3 +1,5 @@
+import { differenceWith } from './differenceWith.ts';
+
 /**
  * Checks if the `subset` array is entirely contained within the `superset` array based on a custom equality function.
  *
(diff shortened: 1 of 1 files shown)


raon0211 (approved):
👍

kaehehehe:
Thanks for refactoring! @ssi02014 

I believe using `differenceWith` is a better choice for maintainability. However, since the implementation of `isSubsetWith` isn’t particularly complex, I thought that sticking with native JS methods might provide better performance than using other functions, which is why I went with that approach.

After doing a quick benchmark test, I found that the original implementation performed slightly better.

스크린샷 2024-11-04 11 20 11


What do you think?

ssi02014:
@kaehehehe 
As I understand it, `es-tookit` prioritizes maintenance and simplicity over performance improvements of `0-0.1x`. 🤗
https://github.com/toss/es-toolkit/blob/main/.github/CONTRIBUTING.md#12-simplicity

PR #789, opened 2024-11-04
test(invariant): add test code

ssi02014:
The first argument of the `invariant` function is named `condition`, and given the use case, it would be nice to define it as a `boolean` type, what do you think?

Is there a reason why the type is `unknown`?

I additionally added test code for type narrowing for the `invariant` function.


raon0211:
Actually it is common practice to write code like `if (!number)` or `if (obj)` -- that's the reason why we have `unknown` here. So I would like to stick with the current code.

ssi02014:
@raon0211 
I understand. The naming of `condition` can be confusing, but as you said, it is correct to keep the current code to allow different types of values. 

Perhaps you would be willing to change the naming?

raon0211:
Yeah, the test code will definitely help. In the meantime, I thought using `if (condition)` is pretty common in JavaScript, so I figured `condition` would be a good name. Do you have any thoughts on the name?

ssi02014:
@raon0211 It would be nice to have a name that encompasses both, since it can take a `value` as an argument in addition to the usual `condition`, but unfortunately I don't have a good idea yet. 😭

So, I'm going to keep the existing code and just add the test code.

ssi02014:
@raon0211 I reverted the existing modifications, leaving only the test code. Thanks for the review. 🤗

raon0211 (approved):
Thanks!

PR #790, opened 2024-11-04
fix(differenceBy): Update `differenceBy` function to handle array-like objects and filter out invalid values

mass2527:
Since the `differenceBy(compat)` function needs to pass the test cases in the [difference-methods spec](https://github.com/lodash/lodash/blob/6a2cc1dfcf7634fea70d1bc5bd22db453df67b42/test/difference-methods.spec.js#L1), I added it to our spec.

After adding this test, I found that two tests were failing, so I fixed them:
- It should work with array-like objects.
- It should filter out values that are not arrays or array-like objects.


raon0211 (approved):
Thanks! Revised code a little to be less verbose.

PR #791, opened 2024-11-04
fix(isEmpty): `isEmpty` should return `true` for objects with only enumerable symbol properties

mass2527:
Lodash's `isEmpty` function checks only the enumerable own string properties of an object. It uses [`for...in` to iterate over enumerable own and inherited properties](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Enumerability_and_ownership_of_properties#traversing_object_properties) and [`hasOwnProperty` to filter for enumerable own properties](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Enumerability_and_ownership_of_properties#querying_object_properties). This behavior applies whether the value is a [prototype](https://github.com/lodash/lodash/blob/ddfd9b11a0126db2302cb70ec9973b66baec0975/lodash.js#L3487-L3491) or [not](https://github.com/lodash/lodash/blob/ddfd9b11a0126db2302cb70ec9973b66baec0975/lodash.js#L11495-L11499).


raon0211 (approved):
Thanks for the catch!

PR #792, opened 2024-11-05
docs(iteratee): Update docs

mass2527:
- Added missing link for `identity`
- Added examples for property-value pairs and partial objects


raon0211 (approved):
Thanks!

Issue #784, opened 2024-11-02
Support for `uniqBy` in compat layer


Gyumong:
Hi @raon0211 :)

If necessary, I plan to implement `uniqBy` method for issue https://github.com/toss/es-toolkit/issues/91.

raon0211:
This will be helpful! 😃

PR #793, opened 2024-11-05
feat(uniqBy): add uniqBy to compat layer

Gyumong:
I added uniqBy in compat layer

issue https://github.com/toss/es-toolkit/issues/784

benchmark

image


raon0211 (approved):
Thanks!

Issue #777, opened 2024-10-31
cloneDeep no cloning methods?


SeeThruHead:
i was trying to get rid of cloneDeep from lodash in our app today and ran into the issue of many methods being undefined on some models we have

just wanted to check to make sure that is expected behaviour of cloneDeep 

thanks

raon0211:
Hello :) Does the problem also occur for `cloneDeep` in our compatibility library? 

```tsx
import { cloneDeep } from 'es-toolkit/compat';
```

For `cloneDeep` in our original library, let us check this.

SeeThruHead:
hello
i just started on a new codebase that is using cloneDeep to clone instantiated classes all over

switching to `import { cloneDeep } from 'es-toolkit'` causes our test to fail with missing methods
i just tried to use es-toolkit/compat and have the same issues

i moved us from lodash to lodash-es to reduce the bundle size and neither of those give us issues with cloneDeep-ing our classes

filipsobol:
I encountered the same issue. Here's a minimal reproduction comparing the behavior in `lodash-es`, `es-toolkit`, and `es-toolkit/compat`: https://stackblitz.com/edit/stackblitz-starters-eanrtc?file=index.js.

It looks like class instances are turned into plain objects.

filipsobol:
Fixed in https://github.com/toss/es-toolkit/pull/794

raon0211:
This was fixed in es-toolkit v1.27.0 :)

PR #794, opened 2024-11-05
fix(cloneDeep): maintain prototype when cloning class instances

filipsobol:
Addresses #777 and #801.

Benchmark results

Benchmark results


filipsobol:
When fixing this issue, I noticed another inconsistency compared to lodash, but it's out of scope of this PR — https://github.com/toss/es-toolkit/issues/795.

src/object/cloneDeep.ts
```diff
@@ -184,7 +185,9 @@ function cloneDeepImpl<T>(obj: T, stack = new Map<any, any>()): T {
   }
 
   if (typeof obj === 'object' && obj !== null) {
-    const result = {};
+    const result =
+      typeof obj.constructor === 'function' && !isPrototype(obj) ? Object.create(Object.getPrototypeOf(obj)) : {};
```
raon0211, line 189: I wonder why we check for whether `obj.constructor` is `function` and `obj` is not a prototype here -- isn't just `Object.create(Object.getPrototypeOf(obj))` enough?

src/object/clone.spec.ts
```diff
@@ -238,4 +238,28 @@ describe('clone', () => {
     expect(clonedError.message).toBe(error.message);
     expect(clonedError.name).toBe(error.name);
   });
+
+  it('should clone class instance', () => {
```
raon0211, line 242: I think this test code is added in `clone`, not `cloneDeep`. Let me add a test case for `cloneDeep`, too.

src/object/cloneDeep.ts
filipsobol: I assumed that calling `Object.create()` and `Object.getPrototypeOf()` on every object would be expensive and hurt performance, but it's actually the opposite! Using just the `Object.create(Object.getPrototypeOf(obj))` improved performance by about 15%. I updated the PR.

src/object/clone.spec.ts
filipsobol: There already is a `should clone instance` test in `cloneDeep.spec.ts`, so that's why I only added a single test covering this case.

raon0211 (approved):
Thanks for your contribution!

PR #796, opened 2024-11-05
perf(compat/array): Improve performance of various array functions in `compat`

filipsobol:
Various compatibility array functions are slow when handling large arrays, which is largely caused by unnecessary `Array.from()` calls. This PR adds a check, so `Array.from()` is only called when necessary.

Below are performance improvements for the `head` function. Results for other methods are similar.

Before:
Before

---

After:
After


raon0211 (approved):
Thanks!

PR #797, opened 2024-11-06
imp(mapKeys): improved mapKeys function type

ssi02014:
In my opinion, there is no need to have `3` generic types, it would be simpler to define types with `2`. What do you think?


src/object/mapKeys.ts
```diff
@@ -3,28 +3,27 @@
...
+export function mapKeys<T extends Record<PropertyKey, any>, K extends PropertyKey>(
   object: T,
-  getNewKey: (value: T[K1], key: K1, object: T) => K2
-): Record<K2, T[K1]> {
-  const result = {} as Record<K2, T[K1]>;
-  const keys = Object.keys(object);
+  getNewKey: (value: T[keyof T], key: keyof T, object: T) => K
+): Record<K, T[keyof T]> {
+  const result = {} as Record<K, T[keyof T]>;
+  const keys = Object.keys(object) as (keyof T)[];
```
ssi02014, line 23: I don't know why there is a `K1` type that extends `keyof T`. 

Since there is no reason to extend `keyof T`, we can just use `keyof T`!

src/object/mapKeys.ts
```diff
@@ -3,28 +3,27 @@
...
+ * @returns {Record<K, T[keyof T]>} - Returns the new mapped object.
  *
  * @example
  * // Example usage:
  * const obj = { a: 1, b: 2 };
  * const result = mapKeys(obj, (value, key) => key + value);
  * console.log(result); // { a1: 1, b2: 2 }
  */
-export function mapKeys<T extends object, K1 extends keyof T, K2 extends PropertyKey>(
+export function mapKeys<T extends Record<PropertyKey, any>, K extends PropertyKey>(
```
ssi02014, line 18: I don't know why there is a `K1` type that extends `keyof T`.

Since there is no reason to extend `keyof T`, we can just use `keyof T`

raon0211 (approved):
Thanks! I wonder why this is written like this..

PR #798, opened 2024-11-06
perf(at): Improve performance and compatibility with lodash

filipsobol:
It turns out that the native `.at()` is kinda slow, and this PR improves the performance (especially in big arrays).

---

Performance before:
Before

---

Performance after:
After


src/array/at.ts
```diff
@@ -14,12 +14,17 @@
...
   for (let i = 0; i < indices.length; i++) {
-    const index = indices[i];
+    let index = indices[i];
 
-    result[i] = arr.at(index);
+    if (index < 0) {
+      index += length;
+    }
+
+    result[i] = index >= 0 && index < length && Number.isInteger(index) ? arr[index] : undefined;
```
raon0211, line 27: Since the `at` function we're working with isn't the compatibility version, I think we should follow the behavior of the native `Array#at(...)` method. 

Otherwise, it could be confusing, because `at(arr, indices)` would behave differently from `indices.map(index => arr.at(index))`. I do appreciate the performance benefits, though.

What do you think?

filipsobol: If you prefer to use the native `.at()` for easier maintenance at the cost of performance, then I'm fine with closing this PR.

However, if you want better performance, then see the updated code. I changed the implementation to match that of `Array.prototype.at` and it's around 3.7x to 12x faster than the current version.

Benchmark results

---

**Because the current and this implementation are incompatible with lodash's, should we create `at()` in `compat`?**

raon0211: If performance is this different, I guess we might go with the improvement you made. Thanks!

I guess we should create `at()` in `compat` to address what you said.

raon0211 (approved):
Thanks!

PR #799, opened 2024-11-06
perf: Improve performance of `toPath`, `get`, `set`, `unset`, `has`, `orderBy`.

filipsobol:
The currently used regular expressions in `toPath()` are expensive, so this PR changes the implementation to not rely on them. Because `toPath` is used in `get`, `set`, `unset`, `has`, and `orderBy`, performance of these functions is also improved.

Performance improvements:
* `toPath()`
  * `toPath: super simple` — 840%
  * `toPath: dots` — 830%
  * `toPath: brackets` — 1270%
  * `toPath: complex` — 1340%
* `get()`
  * `get with string` — 420%
  * `get with string array` — 15%
* `has()`
  * `has with string` — 450%
* `orderBy()`
  * `orderBy (property path)` — 36%
* `set()`
  * `set - dot` — 91%
  * `set - array` — 119%
* `unset()`
  * `unset` — 155%

After these changes, performance of these functions is in most cases on par with lodash.

---

Before:
Before

---

After:
After

---

### DIFF
--- a/src/compat/util/toPath.ts
+++ b/src/compat/util/toPath.ts
@@ -1,24 +1,3 @@
-const DOTS_KEY = /^[\w.]+$/g;
-
-const ESCAPE_REGEXP = /\\(\\)?/g;
-const PROPERTY_REGEXP = RegExp(
-  // Match anything that isn't a dot or bracket.
-  '[^.[\\]]+' +
-    '|' +
-    // Or match property names within brackets.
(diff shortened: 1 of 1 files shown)


raon0211 (approved):
This is awesome! Just a heads-up: lodash's toPath is super fast thanks to memoization, which is fantastic!

PR #814, opened 2024-11-12
feat(keys): Add `keys` to compatibility layer

mass2527:
Added lodash [`keys`](https://lodash.com/docs/4.17.15#keys) to compatibility layer


raon0211 (dismissed):
Thanks!

raon0211 (approved):

PR #817, opened 2024-11-13
fix(flattenObject): flatten for nested objects

### LINKED ISSUE #816: flattenObject doesn't handle array of objects
Given this code

```ts
    const result = flattenObject({
      a: [1, { b: 2 }, 3],
    });

    expect(result).toEqual({
      'a.0': 1,
      'a.1.b': 2,
      'a.2': 3,
    });
```

nested object is not flattened 

Screenshot 2024-11-12 at 19 47 16

nnnnoel:
Fixed #816

### DIFF
--- a/src/object/flattenObject.ts
+++ b/src/object/flattenObject.ts
@@ -1,4 +1,4 @@
-import { isPlainObject } from '../predicate/isPlainObject.ts';
+import { isPlainObject } from "../predicate/isPlainObject.ts";
 
 /**
  * Flattens a nested object into a single level object with dot-separated keys.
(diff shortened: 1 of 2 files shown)


src/object/flattenObject.ts
```diff
@@ -46,7 +46,12 @@ function flattenObjectImpl(object: object, prefix = ''): Record<string, any> {
 
     if (Array.isArray(value)) {
       for (let index = 0; index < value.length; index++) {
-        result[`${prefixedKey}.${index}`] = value[index];
+        const valueKey = `${prefixedKey}.${index}`;
+        if (typeof value[index] === 'object' && value[index] !== null) {
```
nnnnoel, line 50: I think non-object filter needed on top of flattenObjectImpl function, but not sure to code convention of es-toolkit.

raon0211: Actually I think we have a simpler logic here; I made a little modification.

raon0211 (approved):
Thanks!

PR #821, opened 2024-11-13
fix(compat/trim): trim should support character arrays

D-Sketon:


src/compat/string/trim.ts
```diff
@@ -27,10 +28,7 @@ export function trim(str: string, chars?: string | string[], guard?: unknown): s
     }
     case 'object': {
       if (Array.isArray(chars)) {
-        return trimToolkit(
-          str,
-          chars.map(x => x.toString())
-        );
+        return trimToolkit(str, flatten(chars.map(x => x.toString().split(''))));
```
raon0211, line 31:
```suggestion
        return trimToolkit(str, chars.flatMap(x => x.toString().split('')));
```

src/compat/string/trim.ts
```diff
@@ -1,4 +1,5 @@
 import { trim as trimToolkit } from '../../string/trim.ts';
+import { flatten } from '../array/flatten.ts';
```
raon0211, line 2:
```suggestion
```

raon0211 (approved):
Thanks!

PR #823, opened 2024-11-14
fix(intersectionWith): first array and second array's types can different

### LINKED ISSUE #819: Type definitions are same for intersectionWith, differenceWith, etc
I have noticed that the type definitions of `intersectionWith`, `intersectionBy`, `differenceWith`, and `differenceBy` are the same, and they access the same type of arrays.

```ts
import { intersectionWith } from "es-toolkit";

type CSV = { id: number; csv: number };
type JSON = { id: number; json: number };

const array1: CSV[] = [
  { id: 1, csv: 1 },
  { id: 2, csv: 1 },
  { id: 3, csv: 1 },
];
const array2: JSON[] = [
  { id: 2, json: 2 },
  { id: 4, json: 2 },
];

const result = intersectionWith(array1, array2, (a, b) => a.id === b.id);
                                          ^
                                          |
                                          |
                                          |
// Argument of type 'JSON[]' is not assignable to parameter of type 'readonly CSV[]'.
// Property 'csv' is missing in type 'JSON' but required in type 'CSV'.
```

https://github.com/toss/es-toolkit/blob/aac6974e35961fb21ff5bc133e7d3f7b8199ea89/src/array/intersectionWith.ts#L23-L33

This occurs primarily because `secondArr` accepts the same type as `firstArr`. Is this intentional? It feels like a bug, particularly because we only return the values of `firstArr` if `areItemsEqual` returns true.

nnnnoel:
Fixed #819

### DIFF
--- a/src/array/intersectionWith.spec.ts
+++ b/src/array/intersectionWith.spec.ts
@@ -8,4 +8,22 @@ describe('intersectionWith', () => {
       intersectionWith([{ foo: 1 }, { foo: 2 }], [{ foo: 1 }, { foo: 3 }], (x, y) => x.foo === y.foo)
     ).toStrictEqual([{ foo: 1 }]);
   });
+
+  it('should return the intersection of two arrays have different type', () => {
+    type CSV = { id: number; csv: number };
+    type JSON = { id: number; json: number };
+
+    const array1: CSV[] = [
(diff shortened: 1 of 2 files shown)


raon0211 (approved):
Thanks!

PR #825, opened 2024-11-14
feat(add): Add `add` function

kim-dongho:
Add `add` function

image


raon0211:
Hello! :)

Thanks for adding the `add` function to our library! That said, I'm not entirely sure if this function is commonly used in actual JavaScript projects. We aim to include only major functions in our main library, if possible. 

Would you be open to adding this function only to our compatibility library instead, for compatibility purposes?

kim-dongho:
@raon0211 

Hello!

I agree that this function is not commonly used in standard JavaScript. I also agree with adding it to the compatibility library for compatibility purposes.

Thank you :)

raon0211 (approved):
Thanks!

PR #828, opened 2024-11-14
feat(findKey): add findKey to compat layer

Na-hyunwoo:
Implemented lodash's findKey. 

If there is anything that needs to be supplemented, please give me feedback. 

I will supplement it. 

Thank you 🙏🙏

findKey(compat) benchmark


raon0211 (approved):

PR #830, opened 2024-11-15
fix(compat): support PropertyKey

D-Sketon:


D-Sketon:
BTW, should we change `Record<string, unknown>` to `Record<PropertyKey, unknown>` ?

raon0211 (approved):
Thanks for your contribution!

PR #832, opened 2024-11-17
fix(partial): return type error

### LINKED ISSUE #831: `partial`'s return type still requires bound arguments
When using `partial`, the returned function will requires all arguments to be provided. At least, TS thinks so.

Reproduction sandbox here: https://codesandbox.io/p/devbox/t7tdvn.

D-Sketon:
fix #831


raon0211 (approved):
Thanks! Let me fix a little bit of the types.

PR #836, opened 2024-11-21
fix/(intersectionBy, differenceBy, differenceWith): should allow array with different element type.

mass2527:
I think #819 hasn’t been fixed yet.
`intersectionBy`, `differenceWith`, and `differenceBy` should also support arrays with different element types.


raon0211 (approved):
Thanks!

PR #839, opened 2024-11-23
feat(isElement): Implement `isElement`

filipsobol:


src/compat/predicate/isElement.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+ */
+
+describe('isElement', () => {
+  /*
+  it( 'should return `true` for elements', () => {
+    if ( document ) {
+      expect( isElement( body ) ).toBe( true );
+    }
+  } );
+   */
```
filipsobol, line 19: Unfortunately, this test and the one at the bottom of the file require either access to DOM or at least a DOM mocking library like `happy-dom` or `jsdom`.

Consider using Vitest's [Browser Mode](https://main.vitest.dev/guide/browser/), which without isolation is still really fast.

raon0211 (approved):
Thanks for your work!

Issue #837, opened 2024-11-21
Support for `toUpper` in compat layer


DONG-8:
Hi @raon0211  :)

If necessary, I plan to implement `toUpper` method for issue #91.

raon0211:
Sure :)

injae-kim:
PR: https://github.com/toss/es-toolkit/pull/843

PR #843, opened 2024-11-24
feat(toUpper) : add toUpper function

DONG-8:
I added toUpper in compat layer

issue #837 

benchmark
image
image


raon0211 (approved):
Great work! Thanks!

PR #853, opened 2024-11-27
feat(toArray) : Add toArray to compat layer

chhw130:
I added toArray to compat layer

benchmarks
toArray bench

### DIFF
--- a/src/compat/util/toArray.ts
+++ b/src/compat/util/toArray.ts
@@ -1,0 +1,35 @@
+import { isArrayLike } from '../predicate/isArrayLike';
+import { isMap } from '../predicate/isMap';
+
+/**
+ * Converts a value to an array.
+ *
+ * @param {unknown} value - The value to convert.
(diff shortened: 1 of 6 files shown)


src/compat/util/toArray.ts
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * @example
+ * toArray({ 'a': 1, 'b': 2 }) // => returns [1,2]
+ * toArray('abc') // => returns ['a', 'b', 'c']
+ * toArray(1) // => returns []
+ * toArray(null) // => returns []
+ */
+
+export function toArray(value?: unknown): any[] {
+  if (value == null || typeof value === 'number' || typeof value === 'boolean') {
```
raon0211, line 18:
```suggestion
  if (value == null) {
```

We might simplify the condition here.

raon0211 (approved):
Thanks for your contribution!

PR #856, opened 2024-12-01
feat(compat): Add `lastIndexOf` function to `es-toolkit/compat`

evan-moon:
This PR introduces the `lastIndexOf` function to `es-toolkit/compat`. #91 

This function similarly to `Array.prototype.lastIndexOf` but includes support for `NaN` handling, aligning its behavior with [Lodash's implementation](https://github.com/lodash/lodash/blob/v5-wip/src/.internal/strictLastIndexOf.ts).

The implementation has been tested using the test cases from the Lodash repository's `lastIndexOf`. However, in the context of `es-toolkit/compat`, the `fromIndex` parameter only accepts `number` or `undefined` types. 
As a result, test cases related to falsy values for fromIndex have been omitted.

### Benchmark
Test with an array of length 1,000,000.

스크린샷 2024-12-01 오후 2 10 47

### Coverage

These test cases were written with reference to the test cases of [Lodash's lastIndexOf function](https://github.com/lodash/lodash/blob/v5-wip/test/findLastIndex-and-lastIndexOf.spec.js).

스크린샷 2024-12-01 오후 2 11 24

and Instanbul's coverage report says all green.

스크린샷 2024-12-01 오후 2 11 30


raon0211 (dismissed):
Thanks!

raon0211 (approved):

raon0211 (approved):

PR #865, opened 2024-12-03
feat(nth): add `nth` to compat

mass2527:
I added `nth` to the compatibility layer. While fixing the failed test cases, I discovered that `stubA` and `stubB` were incorrect, so I also fixed them.

스크린샷 2024-12-03 오후 11 18 12


raon0211 (approved):
Thanks!

PR #878, opened 2024-12-08
fix(stubObject): Update return type to `Record<PropertyKey, never>`

mass2527:
I believe the return type of `stubObject` would be more accurately represented as `Record<PropertyKey, never>`, as this clearly indicates an empty object. Using `{}` in TypeScript might confuse since it represents a non-nullable value (equal to `NonNullable<unknown>`) rather than specifically an empty object.

### DIFF
--- a/docs/ja/reference/compat/util/stubObject.md
+++ b/docs/ja/reference/compat/util/stubObject.md
@@ -11,12 +11,12 @@
 ## インターフェース
 
 ```typescript
-function stubObject(): {};
+function stubObject(): Record<PropertyKey, never>;
 ```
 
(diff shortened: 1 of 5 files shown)


raon0211:
Hi, thanks for your pull request! (And sorry for the late review.)

You're right that `Record<PropertyKey, never>` is more accurate than `{}`. However, the main goal of `es-toolkit/compat` is to provide 100% compatibility with Lodash. Since `@types/lodash` returns `any`, we’d like to stick with that type for consistency.

Really appreciate you opening this PR, and we’re looking forward to your next contribution! 😊

PR #880, opened 2024-12-08
fix(compat): predicate should have optional parameter

D-Sketon:


raon0211 (approved):
Thanks for pointing these out!

PR #886, opened 2024-12-08
docs: Remove unnecessary readonly keywords

eunhyulkim:
Although there is 'readonly' in the actual interface, it appears to be the overall writing guideline of es-toolkit to omit 'readonly' from arguments and return types when describing method interfaces in the docs. However, I found and fixed inconsistencies in several functions where this wasn't followed.


raon0211 (approved):
Thanks for your contribution!

PR #891, opened 2024-12-08
docs: Fix korean grammar rules for '에요/예요'

eunhyulkim:
The Korean ending '에요/예요' varies depending on whether the preceding syllable has a final consonant, and when the preceding character is an alphabet, it follows the alphabet pronunciation rules. I corrected it accordingly.


raon0211 (approved):
Thanks!

Issue #899, opened 2024-12-13
es-toolkit/compat hasn no exported member "uniqueId" after latest release


redd97:
After the latest release of es-toolkit `uniqueId` function disappeared from "es-toolkit/compat"

It seems to have been removed accidentally in the commit : https://github.com/toss/es-toolkit/commit/537e8a42a80799978000a841ff3ffcad2953ebf0#diff-919544eb15e4bb20ce2371a9a8837aa523de3213cb81589129ecf00414f9e8cbL197

Example of Error: 
```ts
error TS2305: Module '"es-toolkit/compat"' has no exported member 'uniqueId'.
```

PR #900, opened 2024-12-13
fix(compat): restore removed export of uniqueId Function

redd97:
fix https://github.com/toss/es-toolkit/issues/899


raon0211 (approved):
Thanks for pointing this out! I guess we should add a test case for this..

PR #902, opened 2024-12-14
fix(groupBy): preserve Object prototype method

SaeWooKKang:
## What this PR solves
- Fix the `groupBy` function to properly preserve Object prototype (like `toString`, `valueOf`).

## Motivation
While implementing "[fix(groupBy): Support keys like toString or valueOf in groupBy](https://github.com/toss/es-toolkit/commit/1f336b045692e06f852c9b87064a3ae8c5577c7c#diff-636dea33563199f80586e9579e588a6bd655799bd2e43fbed0d3a13ed9286468R27)", we noticed that prototype methods were removed entirely. This is likely unintended behavior unless the user specifically uses these method names as keys.

For example, even when not using 'toString' as a grouping key, attempting to convert the result object to a string would fail:
```typescript
const data = [{ category: 'fruit', name: 'apple' }];
const result = groupBy(data, item => item.category);

// This throws TypeError: Cannot convert object to primitive value
String(result);
```

### DIFF
--- a/src/array/groupBy.spec.ts
+++ b/src/array/groupBy.spec.ts
@@ -46,6 +46,21 @@ describe('groupBy', () => {
     });
   });
 
+  it('should preserve Object prototype', () => {
+    const array = [
+      { category: 'fruit', name: 'apple' },
+      { category: 'fruit', name: 'banana' },
+    ]
+    const result = groupBy(array, item => item.category)
+    
+    expect('toString' in result).toBeTruthy()
+    expect(typeof result.toString).toBe('function')
+    expect(result.fruit).toEqual([
+      { category: 'fruit', name: 'apple' },
+      { category: 'fruit', name: 'banana' },
+    ])
+  })
+
   it('should handle an empty array', () => {
     const array: Array<{ category: string; name: string }> = [];
 
--- a/src/array/groupBy.spec.ts
+++ b/src/array/groupBy.spec.ts
@@ -50,16 +50,16 @@ describe('groupBy', () => {
     const array = [
       { category: 'fruit', name: 'apple' },
       { category: 'fruit', name: 'banana' },
-    ]
-    const result = groupBy(array, item => item.category)
-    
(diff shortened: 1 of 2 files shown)


raon0211:
Hello, thanks for helping us! 

The design principle for es-toolkit is to support everyday usecase, but skip edge cases. (You can refer to our [contributing docs](https://github.com/toss/es-toolkit/blob/main/.github/CONTRIBUTING.md#12-simplicity)).

Is there an actual usecase for the behavior you described? I'm asking because I thought we don't usually access prototype methods when using groupBy.

SaeWooKKang:
@raon0211 
Thank you for the quick feedback.

As you mentioned, there are no direct uses of prototype methods in my code. However, I encountered an issue in a situation beyond my direct control:

1. Created a React boiler project using npm create vite
2. Started the development server with npm run dev
3. Processed server responses using groupBy
4. Error occurred when trying to verify the results with console.log

While the build output works correctly in production, errors occur in the development environment.

This is clearly a situation that developers cannot directly control, and I believe it can be quite tricky to debug.

From my perspective, it feels unnatural for an object to lack its prototype methods when its type is clearly an object - similar to how it would be strange for an array to lack its prototype methods.

However, I understand that this case could be considered an edge case according to es-toolkit's design principles, and I respect your perspective on this matter. Thank you for your feedback.

raon0211:
Hi, thanks for your help! I see what you pointed out.

While reviewing other functions, I noticed that many of them, like `countBy` and `zipObject`, use plain objects (`{}`) to build their results. To keep things consistent, I updated the code to also use `{}` for the result. 

However, I wasn’t sure if this is something we need to enforce with test cases, so I removed that part.

raon0211 (approved):
Thanks!

SaeWooKKang:
Thank you for understanding.

> However, I wasn't sure if this is something we need to enforce with test cases, so I removed that part.

I agree with this point.

The commit caused existing tests to fail. Another person has submitted a #906 to address this issue.

PR #906, opened 2024-12-15
fix(groupBy): should handle keys like toString or valueOf

D-Sketon:


src/array/groupBy.ts
```diff
@@ -37,7 +37,7 @@ export function groupBy<T, K extends PropertyKey>(arr: readonly T[], getKeyFromI
     const item = arr[i];
     const key = getKeyFromItem(item);
 
-    if (result[key] == null) {
+    if (!Object.hasOwn(result, key) || result[key] == null) {
```
SaeWooKKang, line 40: How about this? Are there any edge cases we should consider?
```suggestion
    if (!Object.hasOwn(result, key)) {
```

raon0211 (approved):

Issue #893, opened 2024-12-10
Support for `subtract`


aken-you:
Hi, @raon0211 

I’d like to implement `subtract` method in the `compat` layer if needed. (issue: #91)
Can I work on this?

raon0211:
Sure!

PR #908, opened 2024-12-17
feat(subtract): Add `subtract` to compat layer

aken-you:
#893

### Description

I added `subtract` to compat layer. 

### Benchmark

스크린샷 2024-12-17 오후 10 27 34


raon0211 (approved):
Thanks so much!

PR #909, opened 2024-12-20
fix(reverse): Export reverse from compat barrel file

willmanduffy:
Upgraded to `v1.30.1` and saw the changelog saying that `reverse` was now available in `es-toolkit/compat`. However, when I went to use it, I got the following error:

```
 "reverse" is not exported by "../../node_modules/es-toolkit/dist/compat/index.mjs", imported by "src/components/..."
```


raon0211 (approved):
Thanks! Will release this in the next version soon.

Issue #863, opened 2024-12-02
Support `windowed` function


T3sT3ro:
Kotlin has [`windowed`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.collections/windowed.html) function which is a huge helper. It creates an interleaved windows from a collection for example:

`[1,2,3,4,5].windowed(2) -> [[1,2], [2,3], [3,4], [4,5]]`.

It has also useful parameters for `step size` determining the shift of the window and `partial sindows` determining if leftover windows from the end which are of different size should also be returned.

It is immensely useful when items have to be compared sequentially in pairs/triplets etc. I miss it very much in lodash, so I hope this library could implement it at least. What do you think of it?

raon0211:
I think it would be very helpful. Could you open a pull request for this feature?

T3sT3ro:
I can try, although it must wait until Friday :+1:

PR #911, opened 2024-12-20
feat(windowed): add windowed array function inspired by Kotlin

T3sT3ro:
Implements #863.

Any improvements are welcome. I am not a performance wizard, so any suggestions would be appreciated. I am not sure if I added everything, feel free to commit required changes.

I omitted updating docs because I am not sure how to approach that properly.

I also don't know if I implemented everything I needed, for example if there is something needed in the `compat`.


raon0211 (approved):
Thanks for your help!

PR #914, opened 2024-12-25
fix(isMatch): simplify condition check

apeltop:
Simplify condition check by returning keys.length === 0 directly.

I hope it helps!


raon0211 (approved):
Thanks!

PR #915, opened 2024-12-26
perf(pull): faster `pull`

tuhm1:
Reduce time complexity of `pull` from O($n^2$) to O(n).

Benchmark with lodash: 
pull-after

Benchmark with old implementation: https://stackblitz.com/edit/vitest-dev-vitest-1rmnz8?file=pull.bench.ts


raon0211 (approved):
Thanks for your help!

PR #860, opened 2024-12-01
feat(map): implement map

D-Sketon:
Replace #514 
close #301

PR #917, opened 2024-12-31
Feat(groupBy): Updated groupBy to accept string as a second argument

weijunyu:
## Change

Instead of a key-generating function, callers can now pass a string as the key that they wish to group items by.

## Rationale

In lodash, `groupBy` [also accepts strings as its second argument](https://lodash.com/docs/4.17.15#groupBy) as a shorthand for getting the key to group items on.

This will allow users to do a 1 for 1 replacement with lodash.groupBy.


raon0211:
Hello, this is actually what we will support in our compatibility library. Would you add this similar to this pull request? 

https://github.com/toss/es-toolkit/pull/860

If not, we might work on this soon.

raon0211:
Closing this pull request since it will be resolved in https://github.com/toss/es-toolkit/issues/1005.

PR #925, opened 2025-01-02
feat(reduce): Implement reduce

dasom-jo:


raon0211:
Hey, sorry for the late reply, and thanks for your pull request!

We actually have a function called [pullAt](https://es-toolkit.slash.page/reference/array/pullAt.html) in es-toolkit that’s pretty similar. I think the name `pullAt` fits the behavior a bit better than `reduce`—since `reduce` could easily be confused with `Array.prototype.reduce`.

Could you take a look at our `pullAt` function and let me know what you think?

dasom-jo:
> Hey, sorry for the late reply, and thanks for your pull request!
> 
> We actually have a function called [pullAt](https://es-toolkit.slash.page/reference/array/pullAt.html) in es-toolkit that’s pretty similar. I think the name `pullAt` fits the behavior a bit better than `reduce`—since `reduce` could easily be confused with `Array.prototype.reduce`.
> 
> Could you take a look at our `pullAt` function and let me know what you think?



> Sorry, there was a mistake. I am correcting it now."

raon0211:
Let us look at the lint issue.

raon0211 (approved):
Since we already have the native `Array.prototype.reduce`, I went ahead and added `reduce` to the compatibility library, along with some code to ensure compatibility with lodash. Thanks for opening the pull request!

raon0211 (approved):

PR #928, opened 2025-01-06
feat(isDefinedArray): add `isDefinedArray`

HANYUNSEONG:
Add isDefinedArray function

### DIFF
--- a/src/predicate/isDefinedArray.ts
+++ b/src/predicate/isDefinedArray.ts
@@ -1,0 +1,26 @@
+/**
+ * Check if all items in the array are defined.
+ *
+ * @param array The array to check.
+ * @returns `true` if all items in the array are defined, `false` otherwise.
+ *
+ * @example
+ * ```ts
+ * isDefinedArray([1, 2, 3]); // true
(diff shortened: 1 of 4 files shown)


src/predicate/isDefinedArray.ts
```diff
@@ -0,0 +1,26 @@
...
+ * ```
+ */
+export function isDefinedArray<T>(array: Array<T | undefined | null>): array is T[] {
+  for (let i = 0; i < array.length; i++) {
+    if (array[i] === undefined || array[i] === null) {
+      return false;
+    }
+  }
+
+  return true;
```
skiende74, line 25:
```suggestion
 return array.every(el=> el !== null && el !== undefined);
```
I think this makes code cleaner.

raon0211:
Hello :) Thanks for opening a pull request! I see what this function is doing, but I think using `arr.every(x => x !== undefined && x !== null)` is a bit more explicit and less likely to cause confusion. There can be different interpretations of what a "defined array" really means, so I feel it might be clearer to stick with the native array methods directly. What do you think?

PR #929, opened 2025-01-07
feat(isFiniteNumber): add `isFiniteNumber`

HoberMin:
# Add isFiniteNumber utility function

## Summary
Added a new utility function `isFiniteNumber` that checks if a given value is a finite number. This function also acts as a TypeScript type predicate, helping with type narrowing in TypeScript code.

## Implementation Details
- Function takes any value (`unknown` type) as input
- Returns `boolean` indicating if the value is a finite number
- Implements type predicate functionality (`value is number`)
- Uses native `typeof` and `isFinite` checks

## Usage Examples
```typescript
import { isFiniteNumber } from '@utils/number';

// Regular numbers
isFiniteNumber(42)        // true
isFiniteNumber(-123.45)   // true

// Special number values
isFiniteNumber(Infinity)  // false
isFiniteNumber(-Infinity) // false
isFiniteNumber(NaN)      // false

// Non-number values
isFiniteNumber('123')     // false
isFiniteNumber(null)      // false
isFiniteNumber(undefined) // false

### DIFF
--- a/docs/ko/reference/predicate/isFiniteNumber.md
+++ b/docs/ko/reference/predicate/isFiniteNumber.md
@@ -1,0 +1,43 @@
+# isFiniteNumber
+
+주어진 값이 유한한 숫자인지 확인해요.
+
+유한한 숫자란, `Infinity`, `-Infinity`, `NaN`이 아닌 숫자를 말해요.
+
+TypeScript의 타입 가드로 사용할 수 있어요. 파라미터로 주어진 값의 타입을 `number`로 좁혀요.
+
+## 인터페이스
+
+```typescript
+function isFiniteNumber(value: unknown): value is number;
+```
+
+### 파라미터
+
+- `value` (`unknown`): 유효한 길이인지 확인할 값.
+
+### 반환 값
+
+(`value is number`): 값이 유한한 숫자이면 `true`, 아니면 `false`.
+
+## 예시
+
+```typescript
+import { isFiniteNumber } from 'es-toolkit/predicate';
+
+const value1 = 42;
+const value2 = -42;
+const value3 = Infinity;
+const value4 = -Infinity;
+const value5 = NaN;
+const value6 = 'string';
+const value7 = {};
+
+console.log(isFiniteNumber(value1)); // true
+console.log(isFiniteNumber(value2)); // true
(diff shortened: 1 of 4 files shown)


raon0211:
Hi, thanks for your contribution! I noticed that JavaScript already has the `isFinite(...)` function, so I wonder if this might not be as useful to many users. What do you think?

HoberMin:
@raon0211 

Let me explain the advantages of our function compared to JavaScript's isFinite:

1. String Handling
JavaScript's isFinite automatically attempts to convert strings like "123" to numbers. In contrast, isFiniteNumber provides stricter handling by explicitly checking for the number type.
2. More Predictable Behavior
In JavaScript's isFinite, null is converted to 0, resulting in `isFinite(null`) returning true, while `isFinite(undefined)` returns false. isFiniteNumber eliminates these unpredictable JavaScript behaviors by providing more consistent results.

raon0211:
I understand that JavaScript's `isFinite` has some disadvantages. 

However, these issues don't arise in TypeScript since it enforces numeric arguments. For added strictness, we could use [Number.isFinite](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isFinite), which is stricter. 

What do you think?

HoberMin:
I understand the explanation provided.
However, in the example below, `Number.isFinite` still seems unable to solve the issue:

```ts
export function example(value: unknown) {
  if (Number.isFinite(value)) {
    // value.toFixed(2); Error
    (value as number).toFixed(2); // Need type assertion to work
  }
  if (isFiniteNumber(value)) {
    value.toFixed(2); // ✅
  }
}
````

Additionally, with Number.isFinite, it might be unclear whether the purpose is type checking or value checking.

raon0211:
Hello, sorry for the late reply!

I understand the issue with types in TypeScript. It seems like the problem might be related to TypeScript itself. I believe the signature for `Number.isFinite` should return `value is number`. We could consider contributing to the TypeScript repository to address this.

The goal of es-toolkit is to provide functions for things that are tricky in native JavaScript or TypeScript. If a native function already exists for something, we prefer to use it instead of adding it to our library. I hope you understand our decision not to add this.

Thanks for your understanding!

PR #930, opened 2025-01-09
feat(multiply): Add `multiply` to `compact` layer

aken-you:
스크린샷 2025-01-08 오후 10 42 41


raon0211:
There were some logic which lodash handled, so I added some code and tests.

raon0211 (dismissed):
Thanks!

raon0211 (approved):

PR #933, opened 2025-01-15
(feature) Add Custom Delimiter Support to flattenObject Utility

anaclumos:
## Overview
This PR adds a new `delimiter` parameter to the `flattenObject` utility function, allowing users to specify a custom separator for flattened object keys. The default delimiter remains '.' for backward compatibility.

## Why is this useful?

At @lunit-io we wanted to flatten an object into a JSONPath compatible version with `flattenObject`. This required the delimeter to be a `/`, but the current impl did not support this use case.

Handling duplicate cases are out-of-scope of this library, thus not included here.

## Key Changes
- Added optional `delimiter` parameter to `flattenObject` function signature
- Updated implementation to use the custom delimiter when joining nested keys
- Added comprehensive test cases for various delimiter scenarios
- Updated documentation in multiple languages (EN, JA, KO, ZH)

## Core Files Modified
1. `src/object/flattenObject.ts`: Added delimiter parameter and updated implementation
2. `src/object/flattenObject.spec.ts`: Added new test suite for delimiter functionality
3. Documentation files in multiple languages:
   - `docs/reference/object/flattenObject.md`
   - `docs/ja/reference/object/flattenObject.md`
   - `docs/ko/reference/object/flattenObject.md`
   - `docs/zh_hans/reference/object/flattenObject.md`

## Test Coverage
New test suite includes validation for:
- Forward slash delimiter ('/')
- Dash delimiter ('-')
- Underscore delimiter ('_')
- Multi-character delimiter ('->')
- Empty string delimiter ('')
- Special characters delimiter ('@#$')

## Breaking Changes
None. The new parameter is optional with a default value that maintains existing behavior.

## Example Usage
```typescript
const nested = {
  a: { b: { c: 1 } },
  d: [2, 3]
};

// Previous behavior (still works)
flattenObject(nested)
// { 'a.b.c': 1, 'd.0': 2, 'd.1': 3 }

// New functionality
flattenObject(nested, '/')
// { 'a/b/c': 1, 'd/0': 2, 'd/1': 3 }
```

### DIFF
--- a/docs/ja/reference/object/flattenObject.md
+++ b/docs/ja/reference/object/flattenObject.md
@@ -8,12 +8,13 @@
 ## インターフェース
 
 ```typescript
-function flattenObject(object: object): Record<string, any>;
+function flattenObject(object: object, delimiter = '.'): Record<string, any>;
 ```
 
 ### パラメータ
 
 - `object` (`object`): 平坦化するオブジェクト。
+- `delimiter` (`string`): ネストされたキーの区切り文字。
 
 ### 戻り値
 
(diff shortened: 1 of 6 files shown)


raon0211 (approved):
I thought it would be more appropriate to accept custom delimiters via option objects, so I’ve updated the code accordingly. Thanks for your contribution!

PR #935, opened 2025-01-16
chore(at): improved readability and maintainability

apeltop:
This PR proposes a refactor of the at function to enhance its readability and maintainability by replacing the explicit loop with Array.prototype.map. This change retains the function’s original behavior, including support for negative indices and non-integer index handling, while reducing the number of variables and simplifying the logic.

I believe this change aligns with the project’s coding style and goals, but I understand that code readability can be subjective. I’m open to any suggestions or alternative perspectives from the team!

### DIFF
--- a/src/array/at.ts
+++ b/src/array/at.ts
@@ -14,20 +14,11 @@
  * console.log(result); // [20, 40, 50]
  */
 export function at<T>(arr: readonly T[], indices: number[]): Array<T | undefined> {
-  const result = new Array<T | undefined>(indices.length);
-  const length = arr.length;
-
-  for (let i = 0; i < indices.length; i++) {
-    let index = indices[i];
-
+  return indices.map(index => {
     index = Number.isInteger(index) ? index : Math.trunc(index) || 0;
-
     if (index < 0) {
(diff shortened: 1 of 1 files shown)


raon0211:
Hello :) Could you [run a benchmark](https://github.com/toss/es-toolkit/blob/main/.github/CONTRIBUTING.md#11-performance) for `at` and check the performance difference between the old and new implementation? 

```shell
yarn vitest bench ./benchmarks/performance/at.bench.ts
```

We actually used the `for` loop for performance.

apeltop:
Thank you for taking the time to review this PR and pointing out the potential performance impact. Upon testing, I found that the refactored implementation is approximately 1.3x slower compared to the original. While the updated code improves readability and reduces complexity, the performance regression is not acceptable for this function, given its potential usage scenarios.

I really appreciate the insightful feedback that brought this to my attention—it’s something I hadn’t fully considered during the initial refactor. For now, I will close this PR to ensure the project maintains its performance standards.

Thanks again for your thoughtful review and guidance! 


Screenshot 2025-01-17 at 10 16 38 AM

PR #936, opened 2025-01-16
chore: add error package build handling and updated package.json

ssi02014:
Adds handling for `error` packages that are missing during the build process. 
Also modifies `package.json` to ensure smooth [SubPath](https://nodejs.org/api/packages.html#subpath-exports) is applied.

I'm wondering if I'm missing handling for the `error` package, or if this is intentional?

I think this should be added since we provide documentation and modules for the `error` package in terms of libraries. 

docs: https://es-toolkit.slash.page/reference/error/AbortError.html

스크린샷 2025-01-17 오전 1 35 42

<br />

## Result

This will create the `src/error.d.ts` file, which will also create an `index` file inside the `error` folder.

스크린샷 2025-01-17 오전 1 24 02

스크린샷 2025-01-17 오전 1 24 14

### DIFF
--- a/.scripts/postbuild.sh
+++ b/.scripts/postbuild.sh
@@ -4,6 +4,7 @@
 # TypeScript projects configured with "moduleResolution": "node10"
 # (which is the default when using "module": "commonjs").
 echo "export * from './dist/array';" > array.d.ts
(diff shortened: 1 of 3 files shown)


raon0211 (approved):
Thanks for finding this.

PR #938, opened 2025-01-17
fix(find): add support for fromIndex parameter

### LINKED ISSUE #937: `find` compatibility
I think `/compat/array/find` should support `fromIndex` optional parameter for compatibility. [lodash find](https://lodash.com/docs/4.17.15#find)

Image

Is there a reason for this? or I'd like to make a PR to resolve the issue.

sanginchun:
Resolves #937


raon0211 (approved):
Thanks for your contribution!

PR #939, opened 2025-01-18
feat(groupBy): update type to return non-empty arrays

itsMapleLeaf:
`groupBy` is guaranteed to return a non-empty array for each key. Typing it this way allows one to avoid errors under `noUncheckedIndexAccess`. [See here.](https://www.typescriptlang.org/play/?noUncheckedIndexedAccess=true#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXwHMYdkAHAIQE8AeAFQBp4BpeEADwxFWAGd4AFYqRAwMlJiEoA+ABQAoePCycAtjwBc8AJKcYUAEYQQdKfQWFiZEZpnKQKzbQCU8ALxTmZp5oBK4HDDA1EyMtADaALpSANxycgDe5mB4PBjwqMhqbhYkFJQyYQCMjABMjADMEYwyqC7u6fAApPAlTrHmiAHwMsmoqTlk8DiI8ADy+gBW4BgAdABuUBDIIDw1mTxOLomKir39C0sgmhkq+iLZRLlhAAwRsYoAvnJPcqCQsAgo6Nh4A3kAcngAKIqUhiOiMFjsTjcPiCHDCUTiSSycx2NSaHQiAxGExmRSXKwwGzoxx1DwhOTeeB+ZKBYKMMIMeAzVnhCJRdrbeB7NInPiuP5UQGoEFg-JFUoVKrdWpuDz4ZqtdqKTowbq8v5DEbjKaYeaLZarfmbeDc3YpNIHZbHTJndWCwmkG53cxPJ5AA)

### DIFF
--- a/src/array/groupBy.ts
+++ b/src/array/groupBy.ts
@@ -9,7 +9,7 @@
  * @template K - The type of keys.
  * @param {T[]} arr - The array to group.
  * @param {(item: T) => K} getKeyFromItem - A function that generates a key from an element.
- * @returns {Record<K, T[]>} An object where each key is associated with an array of elements that
+ * @returns {Record<K, [T, ...T[]]>} An object where each key is associated with a non-empty array of elements that
  * share that key.
  *
  * @example
@@ -30,18 +30,18 @@
  * //   ]
  * // }
  */
-export function groupBy<T, K extends PropertyKey>(arr: readonly T[], getKeyFromItem: (item: T) => K): Record<K, T[]> {
-  const result = {} as Record<K, T[]>;
+export function groupBy<T, K extends PropertyKey>(arr: readonly T[], getKeyFromItem: (item: T) => K): Record<K, [T, ...T[]]> {
+  const result = {} as Record<K, [T, ...T[]]>;
 
   for (let i = 0; i < arr.length; i++) {
     const item = arr[i];
     const key = getKeyFromItem(item);
 
     if (!Object.hasOwn(result, key)) {
-      result[key] = [];
+      result[key] = [item];
+    } else {
+      result[key].push(item);
     }
-
-    result[key].push(item);
   }
 
   return result;
--- a/src/array/groupBy.ts
+++ b/src/array/groupBy.ts
@@ -30,7 +30,10 @@
  * //   ]
  * // }
  */
-export function groupBy<T, K extends PropertyKey>(arr: readonly T[], getKeyFromItem: (item: T) => K): Record<K, [T, ...T[]]> {
+export function groupBy<T, K extends PropertyKey>(
+  arr: readonly T[],
+  getKeyFromItem: (item: T) => K
+): Record<K, [T, ...T[]]> {
   const result = {} as Record<K, [T, ...T[]]>;
 
   for (let i = 0; i < arr.length; i++) {


src/array/groupBy.ts
```diff
@@ -30,18 +30,18 @@
...
 
   for (let i = 0; i < arr.length; i++) {
     const item = arr[i];
     const key = getKeyFromItem(item);
 
     if (!Object.hasOwn(result, key)) {
-      result[key] = [];
+      result[key] = [item];
+    } else {
+      result[key].push(item);
```
itsMapleLeaf, line 43: I figured it wouldn't hurt to update the implementation to match the soundness of the return type

raon0211:
Hey, thanks for the suggestion!

I get that `groupBy` could return more specific types for non-empty arrays. But since non-empty arrays only guarantee that the first element is non-nullable, anything beyond that (like `arr[1]` or `arr[2]`) can still be nullable. Because of that, I wasn’t sure if typing for non-empty arrays would add much value.

Is there a particular use case where returning non-empty arrays helps TypeScript?

itsMapleLeaf:
> Hey, thanks for the suggestion!
> 
> I get that `groupBy` could return more specific types for non-empty arrays. But since non-empty arrays only guarantee that the first element is non-nullable, anything beyond that (like `arr[1]` or `arr[2]`) can still be nullable. Because of that, I wasn’t sure if typing for non-empty arrays would add much value.
> 
> Is there a particular use case where returning non-empty arrays helps TypeScript?

Fair question! It applies in any case where you're indexing the first item off of the value array. See the playground here for an example that a coworker ran into in practice: https://www.typescriptlang.org/play/?noUncheckedIndexedAccess=true#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXwHMYdkAHAIQE8AeAFQBp4BpeEADwxFWAGd4AFYqRAwMlJiEoA+ABQAoePCycAtjwBc8AJKcYUAEYQQdKfQWFiZEZpnKQKzbQCU8ALxTmZp5oBK4HDDA1EyMtADaALpSANxycgDe5mB4PBjwqMhqbhYkFJQyYQCMjABMjADMEYwyqC7u6fAApPAlTrHmiAHwMsmoqTlk8DiI8ADy+gBW4BgAdABuUBDIIDw1mTxOLomKir39C0sgmhkq+iLZRLlhAAwRsYoAvnJPcqCQsAgo6Nh4A3kAcngAKIqUhiOiMFjsTjcPiCHDCUTiSSycx2NSaHQiAxGExmRSXKwwGzoxx1DwhOTeeB+ZKBYKMMIMeAzVnhCJRdrbeB7NInPiuP5UQGoEFg-JFUoVKrdWpuDz4ZqtdqKTowbq8v5DEbjKaYeaLZarfmbeDc3YpNIHZbHTJndWCwmkG53cxPJ5AA

unrevised6419:
Closes https://github.com/toss/es-toolkit/issues/1131

joealden:
@raon0211 any objections to this? As @unrevised6419 mentioned above, this would resolve an issue I submitted without seeing that this PR already existed. Without this type change, the DX for us is worse, as we have 1 or 2 scenarios in our app that are pulling the 1st element explicitly.

raon0211:
Thanks again for the suggestion and the detailed discussion — the point about `noUncheckedIndexedAccess` is valid, and I understand the DX concern when pulling the first element off a group.

That said, after some consideration, we've decided to keep the current signature. Two main reasons:

First, TypeScript's own typing for the standard `Object.groupBy(...)` uses plain `T[]` for group values:

```ts
groupBy<K extends PropertyKey, T>(
    items: Iterable<T>,
    keySelector: (item: T, index: number) => K,
): Partial<Record<K, T[]>>;
```

Since es-toolkit aims to stay close to modern JavaScript and its standard typings, following the same convention as `Object.groupBy` keeps our API predictable for users moving between the two.

Second, per our [design principles](https://github.com/toss/es-toolkit/blob/main/.github/CONTRIBUTING.md#13-simplicity), we prefer the simplest interface for the most common use cases over more precise-but-complex types. Non-empty array types tend to leak into downstream code (e.g. friction when passing groups to functions expecting `T[]`, or when mutating groups), so the added precision comes with its own cost.

I'll close this PR for now, but thank you for the contribution and the thoughtful discussion! 🙏

PR #940, opened 2025-01-20
feat(retry-backoff-jitter): additional options exponential backof, jitter 

kim-sung-jee:
## The following features have been added:
* Exponential Backoff
A new factor option to increase delays exponentially (e.g., 100ms, 200ms, 400ms).
The default value is 1.

* Randomized Delays
Added a simple jitter to exponential backoff increases.
The default value is false.

* Customizable Delay Ranges
Added retryMinDelay and retryMaxDelay options for precise delay control.

### DIFF
--- a/docs/ja/reference/function/retry.md
+++ b/docs/ja/reference/function/retry.md
@@ -12,10 +12,13 @@ function retry<T>(func: () => Promise<T>, { retries, delay, signal }: RetryOptio
 
 ### パラメータ
 
-- `func` (`() => Promise<T>`): `Promise`を返す関数。
-- `retries`: 再試行する回数。デフォルトは `Number.POSITIVE_INFINITY` で、成功するまで再試行します。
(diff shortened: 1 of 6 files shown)


raon0211:
Hi, thanks for the great suggestion!

I agree that adding jitter and exponential backoff is important to prevent retry storms. However, one of es-toolkit’s core design principles is to [keep implementations simple](https://github.com/toss/es-toolkit/blob/main/.github/CONTRIBUTING.md#12-simplicity).

How about letting the `delay` option accept a function that returns the delay in milliseconds?

```ts
retry(() => fn(), {
  delay: (retries) => randomInt(0, 2 ** retries)
})
```

This way, users can implement their own strategies as needed, while we keep the core simple.

Thanks again for the thoughtful input!

kim-sung-jee:
Hi, thank you so much for the thoughtful review

I really like the idea of supporting a delay function to let users define their own strategy—definitely keeps the core simple and flexible.

I'll take it from here and push the implementation forward. Looking forward to your feedback once it's up.

Thanks again

Issue #948, opened 2025-02-01
maxBy: difference with lodash


rejetto:
hi, lodash's `maxBy` supports string as second argument (then converted to `x => x?.[string]`, of course) but maxBy in 'es-toolkit/compat' doesn't

gs18004:
Can I work on this?

gs18004:
I am working on this. Thank you!

gs18004:
I made a PR on #949 .

PR #949, opened 2025-02-02
feat(compat): Implement compat/maxBy

gs18004:
I implemented the `maxBy` function to compat, which behaviors similar to Lodash's `maxBy`.

The function finds the element in an array with the maximum value based on a given iteratee, which can be:
- function returning a numeric value
- string representing a property key
- [key, value] pair for property matching
- partial object for deep property matching

Unit tests covering various edge cases have also been added.

Thanks to: #948


raon0211 (dismissed):
Thanks so much for your help! Simplified the implementation a little bit using our original implementation.

raon0211 (approved):

Issue #946, opened 2025-02-01
pickBy: difference with lodash


rejetto:
hi! i'm switching from lodash, using 'es-toolkit/compat', and i noticed that pickBy has a difference: in lodash the second argument is optional, but that's not true with your lib.
thanks for your work!

raon0211:
This is fixed in our latest version in our library.

PR #950, opened 2025-02-02
feat(compat): Implement compat/pickBy

gs18004:
I added the `pickBy` function to compat, which behaviors similar to Lodash's `pickBy`:
the second param of the function is optional.

Thanks to: #946

### DIFF
--- a/src/compat/object/pickBy.ts
+++ b/src/compat/object/pickBy.ts
@@ -21,7 +21,9 @@ export function pickBy<T extends Record<string, any>>(
   obj: T,
   shouldPick?: (value: T[keyof T], key: keyof T) => boolean
 ): Partial<T> {
-  if (obj === null || obj === undefined) return {};
+  if (obj === null || obj === undefined) {
+    return {};
+  }
   const result: Partial<T> = {};
 
   if (shouldPick === undefined) {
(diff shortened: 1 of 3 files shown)


src/compat/object/pickBy.ts
```diff
@@ -0,0 +1,40 @@
...
+ * if the property should be picked, and `false` otherwise.
+ * @returns {Partial<T>} A new object with the properties that satisfy the predicate function.
+ *
+ * @example
+ * const obj = { a: 1, b: 'pick', c: 3 };
+ * const shouldPick = (value) => typeof value === 'string';
+ * const result = pickBy(obj, shouldPick);
+ * // result will be { b: 'pick' }
+ */
+export function pickBy<T extends Record<string, any>>(
```
IlyaShorin, line 20: Hi!
If you could add small fix here it would be great. 

Function doesn't warn you if you pass null as obj, and if you do so, it will crash :) 

however, lodash pickBy in that case returns {}
you can look here:
https://github.com/lodash/lodash/blob/4.17.15/lodash.js#L13582


Please add same here, or some warning that obj can't be null.

gs18004: Thank you for your suggestion! I fixed this problem. Would you check it?

IlyaShorin: Great! Thank you very much) 

Maybe add some tests for it? I'm not sure. 

And wait for approval from someone from repo)

gs18004: Sure! I will add some tests :)

raon0211 (dismissed):
Thanks so much for your help! Moved some test cases from lodash.

raon0211 (approved):

PR #952, opened 2025-02-02
feat(compat): Implement compat/sortedLastIndex, compat/sortedLastIndexBy

gs18004:
I added the `sortedLastIndex`, `sortedLastIndexBy` function to compat.
The testcases follow Lodash.


gs18004:
`ci/circleci: test` failes on `defer.spec.ts`, even though I did not modify this file.

dayongkr:
There are conflicts. Please resolve them and resubmit for review. Thanks!

src/compat/array/sortedLastIndexBy.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+    sortedLastIndexBy([30, 50], 40, function (value) {
+      args = args || [value];
+    });
+    // @ts-expect-error Variable 'args' is used before being assigned.ts(2454)
+    expect(args).toEqual([40]);
+  });
+
+  it('should work with `_.property` shorthands', () => {
+    const objects = [{ x: 30 }, { x: 50 }];
+    const actual = sortedLastIndexBy(objects, { x: 40 }, v => v.x);
```
dayongkr, line 17: Test case misses Lodash's string shorthand pattern. Current test only uses function iteratee `v => v.x` while Lodash tests with string shorthand `'x'`.

Please add test for string shorthand to ensure full compatibility :prey:

src/compat/array/sortedLastIndex.spec.ts
```diff
@@ -0,0 +1,74 @@
+import { describe, expect, it } from 'vitest';
+import { sortBy } from './sortBy.ts';
+import { sortedLastIndex } from './sortedLastIndex.ts';
+
+describe('sortedIndex', () => {
```
dayongkr, line 5:
```suggestion
describe('sortedLastIndex', () => {
```

src/compat/array/sortedLastIndex.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    expect(actual).toEqual(expected);
+  });
+
+  it(`should align with 'sortBy' for various data types`, () => {
+    const symbol1 = Symbol ? Symbol('a') : null;
+    const symbol2 = Symbol ? Symbol('b') : null;
+    const symbol3 = Symbol ? Symbol('c') : null;
+
+    const expected = [1, '2', {}, symbol1, symbol2, null, undefined, NaN, NaN];
```
dayongkr, line 44: Test uses only one array configuration but Lodash tests with two different arrays.

Please add the second array test case: `['2', null, 1, symbol1, NaN, {}, NaN, symbol2, undefined]` to ensure full test coverage and compatibility :pray:

src/compat/array/sortedLastIndexBy.spec.ts
gs18004: `sortedIndexBy.spec.ts` has the similar code, so I referred to this. one.

```typescript
it('should work with `_.property` shorthands', () => {
  const objects = [{ x: 30 }, { x: 50 }];
  const actual = sortedIndexBy(objects, { x: 40 }, v => v.x);

  expect(actual).toBe(1);
});
```

Is this one also wrong?

dayongkr: That's right, it seems that function also needs to be updated!

dayongkr:
Thank you for updating `sortedIndexBy` as well!

However, since the `iteratee` function already exists in `compat/object` and performs the same behavior, I think it might be better to use that instead. Could you check this part?

gs18004:
> Thank you for updating `sortedIndexBy` as well!
> 
> However, since the `iteratee` function already exists in `compat/object` and performs the same behavior, I think it might be better to use that instead. Could you check this part?

Thanks for your feedback! 
I wrote the commit messages in Korean by mistake, is it okay?

src/compat/array/sortedIndexBy.ts
```diff
@@ -25,7 +27,7 @@ const MAX_ARRAY_INDEX = MAX_ARRAY_LENGTH - 1;
 export function sortedIndexBy<T, R>(
   array: ArrayLike<T> | null | undefined,
   value: T,
-  iteratee?: Iteratee<T, R>,
+  iterateeFunc?: Iteratee<T, R>,
```
dayongkr, line 30:
```suggestion
  iteratee?: Iteratee<T, R>,
```
Could you update the parameter name like this? Also, isn't it fine to write `iterateeToolkit(iteratee)` below without using a ternary operator?

dayongkr:
> Thanks for your feedback!
> I wrote the commit messages in Korean by mistake, is it okay?

Of course, no problem at all.

However, there’s a bit more to review, so I don’t think I can merge it today. I’ll take another look and get it merged sometime next week 🙏

raon0211 (approved):
Thanks for your contribution!

Issue #958, opened 2025-02-10
Different behaviour with lodash assignIn


SebastienElet:
Hello,

We have a different behaviour by replacing `lodash` with `es-toolkit/compat`.

Here is a code sample to reproduce the issue
```typescript
import { assignIn } from 'es-toolkit/compat'
import { assignIn as assignInLodash } from 'lodash'

const values = [{ workId: undefined }, { exerciseId: '1' }]
const assignInLodashResult = assignInLodash({}, ...values)
const assignInEsToolKitResult = assignIn({}, ...values)

console.dir({
  assignInLodashResult,
  assignInEsToolKitResult,
})
```

The output is 
```
{
  assignInLodashResult: { workId: undefined, exerciseId: '1' },
  assignInEsToolKitResult: { exerciseId: '1' }
}
```

I use `es-tookit` 1.32 and `lodash` 4.17.11

PR #959, opened 2025-02-11
fix(assignIn): Correct undefined assignment

hsskey:
#958 

Description:
This pull request addresses an issue in the assignIn implementation where properties with undefined values are not assigned to the target object if the key is absent. To align the behavior with Lodash's assignIn, the condition in assignInImpl has been updated to assign the property when the key does not exist on the target object, even if the value is undefined.

Changes include:
- Updating the condition in assignInImpl to check for key existence.
- Adding unit tests to verify that undefined values are assigned correctly.
- Ensuring that the behavior remains consistent with Lodash's implementation.

This fix resolves the reported issue and improves the consistency of property assignment.


raon0211 (approved):
Thanks!

PR #963, opened 2025-02-16
feat(reverseString): Add the reverseString function

### LINKED ISSUE #957: Support for `reverseString`
@raon0211 
What if we added a reverseString function to reverse a string?

### Suggesting
I suggest the `spread + reverse + join` approach.
```ts
export function reverseString(value: string): string {
  return [...value].reverse().join(''); // or Array.from(value).reverse().join('');
}
```
```ts
// result
reverseString("foo 𝌆 bar"); // "rab 𝌆 oof"
reverseString("foo 🤗 bar"); // "rab 🤗 oof"
```

This is because the `for statement`, `split + reverse + join` approach runs into problems with invalid Unicode, as shown below.

```ts
export function reverseString(value: string): string {
  let result = '';

  for (let i = value.length - 1; i >= 0; i--) {
    const char = value[i];

    result += char;
  }

  return result;
}
```

```ts
export function reverseString(value: string): string {
  return value.split("").reverse().join("");
}
```

For example, we can expect the result to look like this
```ts
reverseString("foo 𝌆 bar"); // "rab 𝌆 oof"
reverseString("foo 🤗 bar"); // "rab 🤗 oof"
```

However, the `for statement`, `split + reverse + join` method breaks special characters.
```ts
reverseString('foo 𝌆 bar'); // "rab �� oof"
reverseString('foo 🤗 bar'); // "rab �� oof"
```

Why not add a `reverseString` function by implementing it in a `spread + reverse + join` fashion to simplify the code and handle invalid Unicode issues? 

I personally think this would be a useful module.

### Reference
https://stackoverflow.com/questions/958908/how-do-you-reverse-a-string-in-place

ssi02014:
closed: https://github.com/toss/es-toolkit/issues/957

I worked on adding a new function, reversetString. I implemented it as suggested in the issue above. 🤗


raon0211 (approved):
Thanks so much!

PR #964, opened 2025-02-18
feat(truncate): Add the truncate function 

mreiden:
Attempt implementing lodash's truncate function with tests and benchmarks.

vitest-truncate

image


raon0211 (approved):
Thanks for your contribution!

PR #966, opened 2025-02-24
fix: improved trimStart and trimEnd function logic and added test code

ssi02014:
Why does the `trimStart/trimEnd` function only process `one character` if it is a string as the second argument? 

The current code does not achieve the results you expect from the code below.

```ts
// es-toolkit/string/trimStart or trimEnd

trimStart('*-abc', '*-'); //  '*-abc', We expect 'abc' to be returned here. 😱
trimEnd('abc*-', '*-'); // 'abc*-', We expect 'abc' to be returned here. 😱

trimStart('-_-abc', '-_-'); // '-_-abc', We expect 'abc' to be returned here. 😱
trimEnd('abc-_-', '-_-'); // 'abc-_-', We expect 'abc' to be returned here. 😱
```
```ts
// lodash/trimStart or trimEnd, es-toolit/compat/trimStart or trimEnd

trimStart('-_-abc', '-_-'); // 'abc', Yes! We're getting the results we expect. 🙆‍♂️
trimEnd('abc-_-', '-_-'); // 'abc', Yes! We're getting the results we expect. 🙆‍♂️

// or es-toolkit/trimStart or trimEnd use array argument
trimStart('-_-abc', ['-', '_']); // 'abc', Yes! We're getting the results we expect. 🙆‍♂️
trimEnd('abc-_-', ['-', '_']); // 'abc', Yes! We're getting the results we expect. 🙆‍♂️
```

How you handle `strings longer than 2` is up to you, but why not treat them the same as `trimstart/trimEnd` in `lodash` and `es-toolkit/compat`? 🤔

Alternatively, if you only want to process a single character, you can error out if the string is longer than 2. But personally, it feels awkward to handle only one character. 

I added the expected test code along with the existing ones, all of which pass, and I believe the code is less complex after the modifications. 🤗

## Result Example
```ts
trimStart('-_-abc', '_-'); // abc
trimStart('-_-abc', ['-', '_']); // abc
trimStart('-_-abc', ['_-']); // abc
```
```ts
trimEnd('abc-_-', '_-'); // abc
trimEnd('abc-_-', ['-', '_']); // abc
trimEnd('abc-_-', ['_-']); // abc
```

## benchmark
Performance is still better than lodash. 👍

<br />

스크린샷 2025-02-25 오전 2 33 16


benchmarks/performance/trimStart.bench.ts
```diff
@@ -8,11 +8,11 @@ const trimStartLodash = trimStartLodash_;
 describe('trimStart', () => {
   bench('es-toolkit/trimStart', () => {
     const str = 'kebab-case';
-    trimStartToolkit(str, 'se');
+    trimStartToolkit(str, 'ke');
```
ssi02014, line 11: Remove because it contains strings that are not appropriate for the test.

src/string/trimEnd.ts
```diff
@@ -16,20 +18,11 @@ export function trimEnd(str: string, chars?: string | string[]): string {
     return str.trimEnd();
   }
 
+  const charsToUse = Array.isArray(chars) ? flatMap(chars, char => char.toString().split('')) : chars;
```
ssi02014, line 21: Handle `chars` inside trimStart/trimEnd; by consolidating them, `switch statements` can also be removed.

src/string/trimEnd.ts
```diff
@@ -16,20 +18,11 @@ export function trimEnd(str: string, chars?: string | string[]): string {
     return str.trimEnd();
   }
 
+  const charsToUse = Array.isArray(chars) ? flatMap(chars, char => char.split('')) : chars;
```
ssi02014, line 21: Handle `chars` inside `trimStart/trimEnd` by consolidating them, `switch statements` can also be removed.

benchmarks/performance/trimStart.bench.ts
dayongkr: Oh, good catch! Thank you!

dayongkr:
The behavior does seem a bit strange. And there’s more to discuss—

How do you think `trimEnd('abccc', 'cc')` would behave? I expected it to return `'abc'`, but Lodash returns `'ab'`, which I also find odd.

There seems to be quite a bit to go over. How about discussing this on Discord?

PR #967, opened 2025-02-25
feat(predicate): Add isJSON utility function

HoberMin:
# Add isJSON utility function

## Summary
Added a new utility function `isJSON` that checks if a given value is a valid JSON string. This function also acts as a TypeScript type predicate, helping with type narrowing in TypeScript code.

## Implementation Details
- Function takes any value (`unknown` type) as input
- Returns `boolean` indicating if the string can be parsed as valid JSON
- Implements type predicate functionality (`value is string`)
- Uses try/catch with native `JSON.parse()` for validation
- Handles edge cases like JSON primitives ('null', 'true', etc.)

## Usage Examples
```typescript
import { isJSON } from 'es-toolkit/predicate';

// Valid JSON strings
isJSON('{"name":"John","age":30}')  // true
isJSON('[1,2,3]')                  // true
isJSON('true')                     // true (JSON primitive)
isJSON('null')                     // true (JSON primitive)
isJSON('42')                       // true (valid number string)

// Invalid JSON or non-strings
isJSON('invalid json')             // false
isJSON({ name: 'John' })           // false (object, not string)
isJSON(null)                       // false (not a string)


raon0211 (approved):
Thanks for your contribution!

PR #970, opened 2025-02-28
feat(isURL): Add isURL predicate function

yeophs:
Checks if a value is a valid URL string.

Uses JavaScript's built-in URL constructor to validate the URL.
Valid URL format must include a protocol (http, https, etc.).


dayongkr:
Sorry for the long wait.

Our priorities have shifted and new functions aren't high on the list right now. See the [contributing guide](https://github.com/toss/es-toolkit/blob/main/.github/CONTRIBUTING.md#41-what-we-accept) for what we accept.

Also, as @johnpyp noted, `URL.canParse(value)` covers this in a single call, which puts `isURL` close to the built-in replacements [§1.1](https://github.com/toss/es-toolkit/blob/main/.github/CONTRIBUTING.md#11-development-scope) leaves out.

Closing for now. If you'd like to make the case, an [Ideas](https://github.com/toss/es-toolkit/discussions/new?category=ideas) discussion is the place.

Thanks for the contribution.

Issue #846, opened 2024-11-25
Support `sortedIndex`, `sortedIndexBy`, `sortedIndexOf`


1eeminhyeong:
Hi, @raon0211 
I would like to implement a `sortedIndex`, `sortedIndexBy`, `sortedIndexOf` method if needed, is that okay?
If so, can I implement it in `compat` layer?

Related Issues: #91

raon0211:
Sure!

PR #971, opened 2025-03-02
feat(sortedIndexOf): Implement compat/sortedIndexOf

Kyujenius:
I added `sortedIndexOf` in compat layer using by sortedIndex which is 

Resolves: #846 

```tsx
import { sortedIndex } from './sortedIndex';

/**
 * This method is like `indexOf` but performs a binary search on a sorted array.
 * @param {ArrayLike<T> | null | undefined} array The sorted array to inspect.
 * @param {T} value The value to search for.
 * @returns {number} Returns the index of the matched value, else -1.
 * @example
 *
 * sortedIndexOf([4,5,5,5,6], 5) => 1
 * sortedIndexOf([1.1, 2.2, 3.3], 2.2) => 1
 *
 */
export function sortedIndexOf<T>(array: ArrayLike<T> | null | undefined, value: T): number {
  if (!array?.length) return -1;

  const index = sortedIndex(array, value);
  if (index < array.length && areValuesEqual(array[index], value)) return index;
  return -1;
}

const areValuesEqual = <T>(a: T, b: T): boolean => Object.is(a, b);

```

I conducted more extensive testing, comparing all results with the existing Lodash functions. My tests covered a wider range of scenarios, and in all cases, my implementation produced identical results to the original Lodash functions.

Test

|Test List|Test Coverage|
|---|---|
|스크린샷 2025-03-03 오전 3 02 29|스크린샷 2025-03-03 오전 3 02 57|

|**Bench Test (It is 1.7x~ 1.8x faster than lodash )**|
|--|
|benchTest|


dayongkr:
Thank you for your contribution! I will review it within the weekend.

Kyujenius:
@dayongkr Of course! If there's anything incorrect or that needs to be fixed, please feel free to let me know! If possible, I'd love to contribute even more!

src/compat/array/sortedIndexOf.spec.ts
```diff
@@ -0,0 +1,114 @@
+import { describe, expect, it } from 'vitest';
+import { sortedIndexOf } from './sortedIndexOf';
+
+describe('sortedIndexOf', () => {
+  // Basic functionality
+  it('should find first occurrence in numeric array', () => {
```
dayongkr, line 6: To ensure compatibility with Lodash, we are adding its test cases. Could you add the Lodash test case and remove the others for maintenance?

[Test Case #1](https://github.com/lodash/lodash/blob/v5-wip/test/sortedIndexOf-methods.spec.js)
[Test Case #2](https://github.com/lodash/lodash/blob/v5-wip/test/indexOf-methods.spec.js)

dayongkr: If you feel that the Lodash test case is insufficient, you’re welcome to add additional test cases.

src/compat/array/sortedIndexOf.ts
```diff
@@ -0,0 +1,26 @@
...
+ * sortedIndexOf([1.1, 2.2, 3.3], 2.2) => 1
+ *
+ */
+export function sortedIndexOf<T>(array: ArrayLike<T> | null | undefined, value: T): number {
+  if (!array?.length) {
+    return -1;
+  }
+
+  const index = sortedIndex(array, value);
+  if (index < array.length && areValuesEqual(array[index], value)) {
```
dayongkr, line 20:
```suggestion
  if (index < array.length && eq(array[index], value)) {
```

Lodash uses`eq` for comparisons, and since we have implemented our own `eq` function, it would be better to use it as well!
> In fact, the two behave differently—for example, `Object.is(0, -0)` is `false`, whereas `eq(0, -0)` is `true`.

src/compat/array/sortedIndexOf.ts
```diff
@@ -0,0 +1,26 @@
+import { sortedIndex } from './sortedIndex';
+
+/**
+ * This method is like `indexOf` but performs a binary search on a sorted array.
```
dayongkr, line 4:
```suggestion
 * This method is like `indexOf` but performs a binary search on a sorted `array`.
```

src/compat/array/sortedIndexOf.ts
```diff
@@ -0,0 +1,26 @@
...
+ * @param {ArrayLike<T> | null | undefined} array The sorted array to inspect.
+ * @param {T} value The value to search for.
+ * @returns {number} Returns the index of the matched value, else -1.
+ * @example
+ *
```
dayongkr, line 9:
```suggestion
 *
 * @example
```

src/compat/array/sortedIndexOf.spec.ts
Kyujenius: Thank you so much for helping me despite my shortcomings! I will make the corrections according to your comments as soon as possible!

dayongkr:
This seems to be your first contribution, and it's also my first review 😄 I can see that you put a lot of thought into your work!

If you plan to contribute more in the future, I recommend joining our [Discord](https://discord.gg/swSfnNQ9), where we actively communicate.

Kyujenius:
Thanks for inviting me ! 😄 It gives me strength!

Kyujenius:
@dayongkr 
Looking at the test case you provided, I see that it combines multiple methods in a unified and consistent way. Should I also follow this approach? Or would it be better for me to organize the tests by separating them internally by individual methods?

src/compat/array/sortedIndexOf.spec.ts
Kyujenius: @dayongkr
Looking at the test case you provided, I see that it combines multiple methods in a unified and consistent way. Should I also follow this approach? Or would it be better for me to organize the tests by separating them internally by individual methods?

dayongkr: Since we manage each function separately, it is better to separate the test cases by function!

Kyujenius: Thank you for the quick reply 😄

Kyujenius: @dayongkr 
I'm sorry to keep bothering you.

I made the commit by reflecting all the comments you mentioned. However, I am in a dilemma.

What should I do if I write in Typescript, when I specify any[], and test case when I have more parameters of the function, I get an lint error?

Everything else is considered and I put it up!

dayongkr:
```
// eslint-disable-next-line @typescript-eslint/ban-ts-comment
//@ts-expect-error
```

Although it's not a necessary test case, I think it's better to show that we ran the test cases from Lodash as they are. So, I think it would be fine to just add the above comment to each line for ignoring type check.

dayongkr: And feel free to ask any questions at any time! We need to share our thoughts with each other to achieve consistent results. Thanks for your contributions!

src/compat/array/sortedIndexOf.ts
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * This method is like `indexOf` but performs a binary search on a sorted `array`.
+ * @param {ArrayLike<T> | null | undefined} array The sorted array to inspect.
+ * @param {T} value The value to search for.
+ * @returns {number} Returns the index of the matched value, else -1.
+ *
+ * @example
+ * sortedIndexOf([4,5,5,5,6], 5) => 1
+ * sortedIndexOf([1.1, 2.2, 3.3], 2.2) => 1
```
dayongkr, line 12:
```typescript
import { sortedIndexOf } from 'es-toolkit/compat';

const numbers = [11, 22, 33, 44, 55];
const unSortedNumbers = [55, 33, 22, 11, 44];

// Normal case
sortedIndexOf(numbers, 11);
// Return value: 0
// Explanation: The position of the value 11 in the numbers array is 0.

// Non-existent value
sortedIndexOf(numbers, 30);
// Return value: -1
// Explanation: 30 doesn't exist in the array, so it returns -1.

// Empty array
sortedIndexOf([], 30);
// Return value: -1
// Explanation: When searching for a value in an empty array, it returns -1.

// Unsorted array
sortedIndexOf(unSortedNumbers, 11);
// Return value: -1
// Explanation: When using an unsorted array, it returns -1.
```
Could you provide a more detailed example as mentioned in the docs? There are people who only refer to the JS docs, and we're trying to sync both.

Kyujenius: Sure! 👍

src/compat/array/sortedIndexOf.spec.ts
```diff
@@ -0,0 +1,148 @@
+import { describe, expect, it } from 'vitest';
+import { sortedIndexOf } from './sortedIndexOf';
+
+describe('sortedIndexOf', () => {
+  //--------------- Lodash Test Case #1 -------------------
+  it('should find first occurrence in numeric array', () => {
```
dayongkr, line 6:
```suggestion
  it('should perform a binary search', () => {
```
I think it would be better to match the test title with Lodash. 
```ts
it(`\`_.${methodName}\` should accept a falsey \`array\``, () => {
                const expected = lodashStable.map(falsey, lodashStable.constant(-1));

                const actual = lodashStable.map(falsey, (array, index) => {
                    try {
                        return index ? func(array) : func();
                    } catch (e) {}
                });

                expect(actual).toEqual(expected);
            });
```
Also, it seems like the above test case is missing. Could you check on that?

src/compat/array/sortedIndexOf.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+  it(`should return \`-1\` for an unmatched value`, () => {
+    const array = [1, 2, 3];
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-expect-error
+    const empty = [];
+
+    expect(sortedIndexOf(array, 4)).toBe(-1);
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-expect-error
+    expect(sortedIndexOf(array, 4, true)).toBe(-1);
```
dayongkr, line 21: I meant to add comments only in cases where arguments are passed incorrectly like this. Comments shouldn't be added elsewhere.

It would be better to use as any or find another solution.

src/compat/array/sortedIndexOf.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+    const array = [4, 4, 5, 5, 5, 6, 6];
+    expect(sortedIndexOf(array, 5)).toBe(2);
+  });
+
+  //--------------- Lodash Test Case #2 -------------------
+  it(`should return \`-1\` for an unmatched value`, () => {
+    const array = [1, 2, 3];
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-expect-error
+    const empty = [];
```
dayongkr, line 16:
```suggestion
    const empty: unknown[] = [];
```

You can do like this!

src/compat/array/sortedIndexOf.spec.ts
Kyujenius: I think I was in a hurry. I'll reflect it meticulously.

src/compat/array/sortedIndexOf.spec.ts
Kyujenius: It looks great..! 👍

src/compat/array/sortedIndexOf.spec.ts
dayongkr: Only the small things are left! There's no need to rush 👍

src/compat/array/sortedIndexOf.ts
Kyujenius: I did it! May I ask if this is what you want?

dayongkr:
```ts
import { sortedIndex } from './sortedIndex';
import { eq } from '../util/eq';

const unSortedArray = [55, 33, 22, 11, 44];
const duplicateArray = [1, 2, 2, 3, 3, 3, 4];
const emptyArray = [];
const zeroMinusArray = [-0];
const zeroPlusArray = [0];
const floatingArray = [1.1, 2.2, 3.3];
const arrayLike = { length: 3, 0: 10, 1: 20, 2: 30 };
/**
 * This method is like `indexOf` but performs a binary search on a sorted `array`.
 *
 * @param {ArrayLike<T> | null | undefined} array The sorted array to inspect.
 * @param {T} value The value to search for.
 * @returns {number} Returns the index of the matched value, else -1.
 *
 * @example
 * const numbers = [1, 2, 3, 4, 5];
 * sortedIndexOf(numbers, 11); // Return value: 0
 * sortedIndexOf(numbers, 30); // Return value: -1
 *
 * // If the value is duplicated, it returns the first index of the value.
 * const duplicateNumbers = [1, 2, 2, 3, 3, 3, 4];
 * sortedIndexOf(duplicateNumbers, 3); // Return value: 3
 *
 * // If the array is unsorted, it can return the wrong index.
 * const unSortedArray = [55, 33, 22, 11, 44];
 * sortedIndexOf(unSortedArray, 11); // Return value: -1
 *
 * // -0 and 0 are treated the same
 * const mixedZeroArray = [-0, 0];
 * sortedIndexOf(mixedZeroArray, 0); // Return value: 0
 * sortedIndexOf(mixedZeroArray, -0); // Return value: 0
 *
 * // It works with array-like objects
 * const arrayLike = { length: 3, 0: 10, 1: 20, 2: 30 };
 * sortedIndexOf(arrayLike, 20); // Return value: 1
 */
export function sortedIndexOf<T>(array: ArrayLike<T> | null | undefined, value: T): number {
  if (!array?.length) {
    return -1;
  }

  const index = sortedIndex(array, value);
  if (index < array.length && eq(array[index], value)) {
    return index;
  }
  return -1;
}
```

I have reorganized the examples to make them more concise. If you find this satisfactory, please consider applying these changes to the documentation. Thank you!

dayongkr (approved):
The examples modifications are minor, so I will make the adjustments myself!

Welcome to your first contribution, and huge thanks!!

Kyujenius:
@dayongkr 
Thank you very much for letting me know one by one, and I was able to finish it. Thank you. 😢 💌 💯

PR #972, opened 2025-03-05
fix(reverseString): fix comment and docs

ssi02014:
I accidentally wrote the comment Example incorrectly, so I'll fix it.


raon0211 (approved):
Thanks!

PR #1056, opened 2025-03-16
feat(compat): add functions to compat layer

kimpuro:
I added `functions` in compat layer.

Resolves: #1036 

Most of the code is identical to lodash.
The test code is also the same, with some additional code added for test coverage.
스크린샷 2025-03-16 오후 2 58 19

These are the benchmark results.
스크린샷 2025-03-16 오후 2 55 59

If there is anything missing, please let me know!
Thank you.


dayongkr:
It would be better to use `keys` from the compat package instead of `Object.keys`, and minor documentation updates are needed.

Since these are very minor issues, I'll modify them directly after merging.

Thank you for your contribution!!

dayongkr (approved):

PR #1058, opened 2025-03-16
docs: edit link when clicking reference tab in nav

seongminn:
ES Toolkit Docs has the navigation links that allow users to access each category. Currently, clicking on the Reference tab redirects to the [Chunk](https://es-toolkit.slash.page/reference/array/chunk.html) page, which is the third item in the Reference category. This behavior is unexpected and may confuse users.

To improve usability, I’ve updated the links to direct users to the first subpage in the category, which is the [at](https://es-toolkit.slash.page/reference/array/at.html) page. This ensures a more intuitive navigation experience.


dayongkr (approved):
Thanks for your contribution!

PR #1062, opened 2025-03-18
feat(takeWhile): Implement compat/takeWhile

### LINKED ISSUE #990: Add `takeWhile` to compat package
## Description  
We have to add `takeWhile` to `compat/array` package.  

If there are no assignees, feel free to assign yourself and contribute. Thank you! 😊  

## References  

- [Lodash Document](https://lodash.com/docs/4.17.15#takeWhile)  
- [Test Case](https://github.com/lodash/lodash/blob/v5-wip/test/takeWhile.spec.js)

L2HYUNN:
close #990 

### Implement

```ts
import { toArray } from '../_internal/toArray';
import { negate } from '../function/negate';
import { isArrayLikeObject } from '../predicate/isArrayLikeObject';
import { iteratee as createIteratee } from '../util/iteratee';

export function takeWhile<T>(
  _array: ArrayLike<T> | null | undefined,
  predicate?:
    | ((value: T, index: number, array: ArrayLike<T>) => unknown)
    | Partial<T>
    | [keyof T, unknown]
    | PropertyKey
): T[] {
  if (!isArrayLikeObject(_array)) {
    return [];
  }

  const array = toArray(_array);
  const index = array.findIndex(negate(createIteratee(predicate)));

  return index === -1 ? array : array.slice(0, index);
}
```

### Test
스크린샷 2025-03-18 오후 6 16 29
스크린샷 2025-03-18 오후 6 16 53
스크린샷 2025-03-18 오후 6 22 21


src/compat/array/takeWhile.ts
```diff
@@ -0,0 +1,157 @@
...
+    | ((value: T, index: number, array: ArrayLike<T>) => unknown)
+    | Partial<T>
+    | [keyof T, unknown]
+    | PropertyKey
+): T[] {
+  if (!isArrayLikeObject(_array)) {
+    return [];
+  }
+
+  const array = toArray(_array);
```
dayongkr, line 153:
```suggestion
  array: ArrayLike<T> | null | undefined,
  predicate?:
    | ((value: T, index: number, array: ArrayLike<T>) => unknown)
    | Partial<T>
    | [keyof T, unknown]
    | PropertyKey
): T[] {
  if (!isArrayLikeObject(_array)) {
    return [];
  }

  array = toArray(array);
```

Since it's a function parameter and exposed externally, how about writing it like this instead of including an underscore?

src/compat/array/takeWhile.ts
```diff
@@ -0,0 +1,157 @@
+import { toArray } from '../_internal/toArray';
+import { negate } from '../function/negate';
+import { isArrayLikeObject } from '../predicate/isArrayLikeObject';
+import { iteratee as createIteratee } from '../util/iteratee';
```
dayongkr, line 4: Why is aliasing used here??

src/compat/array/takeWhile.ts
```diff
@@ -0,0 +1,157 @@
...
+    | PropertyKey
+): T[] {
+  if (!isArrayLikeObject(_array)) {
+    return [];
+  }
+
+  const array = toArray(_array);
+  const index = array.findIndex(negate(createIteratee(predicate)));
+
+  return index === -1 ? array : array.slice(0, index);
```
dayongkr, line 156: If the index is `0`, it returns an empty array, which might not be the intended behavior.

src/compat/array/takeWhile.ts
L2HYUNN: I also like your approach.

However, I wrote it this way to maintain consistency with the existing code.
Since `compat/takeRightWhile` uses underscore, I followed the same convention here for consistency.

Do you think it would be better to make the change as suggested?

src/compat/array/takeWhile.ts
L2HYUNN: For the same reason as explained above

src/compat/array/takeWhile.ts
L2HYUNN: As far as I understand, returning an empty array when `index === 0` is consistent with Lodash's takeWhile. 

Could you share a counterexample, if any?

dayongkr: Ah, I misunderstood the implementation. Since it's using negate, it should be fine. 🙏

src/compat/array/takeWhile.ts
dayongkr: Only this function is written that way, while most of the other functions don't use underscores.

There are occasionally functions that don't follow the convention, so when implementing a compat function, it's a good idea to also refer to Lodash's interface!

src/compat/array/takeWhile.ts
dayongkr: It's fine to add an alias if there's a name conflict, but doing so unnecessarily can be confusing.

So don't hesitate to improve things instead of blindly following existing patterns! 😄

src/compat/array/takeWhile.ts
dayongkr: Thank you so much for carefully submitting the PR while referring to other code!

L2HYUNN: I removed underscores from the parameter name and refactored the implementation to use internal variables instead.

Please let me know if there's a cleaner or more idiomatic way to approach this.

L2HYUNN:
@dayongkr 
Thanks for the review! 
I've reflected your feedback. When you have a moment, I'd appreciate it if you could review again.

L2HYUNN:
Rebased and force pushed to fix a linting issue.

docs/ko/reference/array/takeWhile.md
```diff
@@ -27,3 +27,60 @@ takeWhile([1, 2, 3, 4], x => x < 3);
...
+
+### 인터페이스
+
+```typescript
+function takeWhile<T>(array: ArrayLike<T> | null | undefined): T[];
+function takeWhile<T>(
+  array: ArrayLike<T> | null | undefined,
+  predicate: (item: T, index: number, array: T[]) => unknown
+): T[];
+function takeWhile<T작(array: ArrayLike<T> | null | undefined, matches: Partial<T>): T[];
```
raon0211, line 50:
```suggestion
function takeWhile<T>(array: ArrayLike<T> | null | undefined, matches: Partial<T>): T[];
```

raon0211:
Your contribution was perfect 👍

PR #1085, opened 2025-03-26
feat(predicate): Add `isBrowser` and `isNode`

dayongkr:
Instead of checking with typeof window or typeof process, I add a predicate utility using ES's `globalThis` to detect browser (`window.document`) or Node.js (`process.node.version`) environments.

However, testing these functions in multiple environments with Vitest was difficult, so I tested it only in environments like jsdom. 

Please let me know if additional tests are needed!


raon0211 (approved):
Great work!

src/predicate/isBrowser.ts
```diff
@@ -0,0 +1,17 @@
...
+ * @returns {boolean} `true` if the current environment is a browser, otherwise `false`.
+ *
+ * @example
+ * if (isBrowser()) {
+ *   console.log("This is running in a browser");
+ *   document.getElementById('app').innerHTML = 'Hello World';
+ * }
+ */
+export function isBrowser(): boolean {
+  return globalThis.window?.document != null;
```
Smrtnyk, line 16: what if it is in worker or service worker, will this still work?
that is still browser and not node environment

PR #1086, opened 2025-03-26
chore(eslint): Restrict console usage in source code

dayongkr:
Sometimes, `console` statements are accidentally left in source files, so we should prevent this using ESLint.


raon0211 (approved):
:+1:

raon0211 (approved):

PR #1091, opened 2025-03-27
docs(trimStart, trimEnd): Add docs that users should pass a string with a single character

raon0211:


src/string/trimEnd.spec.ts
```diff
@@ -45,4 +45,20 @@ describe('trimEnd', () => {
...
+  it('should remove trailing characters from the string when multiple characters are provided in an array', () => {
+    expect(trimEnd('123000', ['0', '3'])).toEqual('12');
+  });
+
+  it('should return the string unchanged when none of the trailing characters in the array match', () => {
+    expect(trimEnd('hello', ['x', 'y', 'z'])).toEqual('hello');
+  });
+
+  it('should handle cases where multiple trailing characters in the array need removal', () => {
+    expect(trimEnd('abcabcabc', ['c', 'b'])).toEqual('a');
```
raon0211, line 62:
```suggestion
    expect(trimEnd('abcabcabc', ['c', 'b'])).toEqual('abcabca');
```

PR #1093, opened 2025-03-27
feat(camelizeKeys, snakeizeKeys): Add `camelizeKeys`, `snakeizeKeys` to object

HyeokjaeLee:
This PR adds two utility functions for recursively converting object keys between formats:
- camelizeKeys: Converts all object keys to camelCase format
- snakeizeKeys: Converts all object keys to snake_case format

Features:
- Deeply recursive conversion for nested objects and arrays
- Type-safety with TypeScript type transformations

camelizeKeys-spec
snakeizeKeys-spec
57f75f9e)

amelizeKeys-bench
snakeizeKeys-bench


dayongkr:
The compat package is for lodash compatibility, but it looks like these functions aren’t part of lodash.

Please double-check, and if confirmed, kindly close the PR and issue.

Thanks for your contribution!

HyeokjaeLee:
Hello, thank you for the review!

Yes, the `camelizeKeys` and `snakeizeKeys` functions are indeed not part of lodash. However, these functions are useful utility functions for recursively transforming object keys, and I submitted the PR to add them to the es-toolkit.

These functions are already located in the `src/object/` directory and are not part of the compat package. Therefore, they are unrelated to lodash compatibility.

Could you please clarify whether adding new utility functions to the es-toolkit project is allowed, or if we should strictly implement only functions that are identical to those in lodash?

dayongkr:
Adding the function to`es-toolkit` (instead of `es-toolkit/compat`) would require more thorough consideration. First, we need to validate whether there's enough demand for this function.

Would you be able to share more about why this function is needed, or whether similar functions in other packages have a high number of downloads on npm?

Thank you again for your contribution!

dayongkr:
Additionally, would there be any benefits to supporting the `SnakeCase` type?

HyeokjaeLee:
Thank you for your thoughtful review and questions!

### Why these functions are needed:

1. **API Integration Use Case**: Many APIs use snake_case in their responses (Stripe, Twitter, Eventbrite), while JavaScript/TypeScript codebases typically use camelCase. These utilities enable seamless conversion between the two formats without writing custom code for each API integration.

2. **High Demand**: Similar packages have significant download numbers:
   - `camelcase-keys` has 16.4M+ weekly downloads
   - `snakecase-keys` has 800K+ weekly downloads

3. **Front-end to Back-end Communication**: When front-end (JavaScript, camelCase) communicates with back-end (Python/Ruby, snake_case), these functions simplify data transformation and improve developer experience.

4. **Type Safety**: Our implementation includes proper TypeScript support with the `Camelized<T>` and `Snakified<T>` type utilities, which provide type-safe transformations that preserve the structure of complex nested objects.

raon0211:

I think this use case is important, and I really appreciate your work on it. I'd like to suggest a couple of minor changes to the pull request:

- I believe `toCamelCaseKeys` and `toSnakeCaseKeys` are more natural names than the original `camelizeKeys` and `snakeizeKeys`, so let’s consider renaming them.
- It might be good to refactor the code slightly for clarity.

Let’s tackle these improvements after merging this pull request. Thanks again for the great contribution! 🙂

raon0211 (approved):

Issue #1095, opened 2025-03-27
orderBy typing should support string[] and number[] like lodash


lanshunfang:
The `T extends object` in 

```typescript
declare function orderBy<T extends object>(arr: readonly T[], criteria: Array<((item: T) => unknown) | keyof T>, orders: Array<'asc' | 'desc'>): T[];
```

should be changed to 

`T extends Object`

```typescript
declare function orderBy<T extends object | string | number>(arr: readonly T[], criteria: Array<((item: T) => unknown) | keyof T>, orders: Array<'asc' | 'desc'>): T[];
```

lanshunfang:
Before change:

![Image](https://github.com/user-attachments/assets/c8acd3e8-ec0c-4fb2-90ca-d27c26158ba7)

After change:

![Image](https://github.com/user-attachments/assets/6d540105-f3d2-43b6-baec-331d68787896)

lanshunfang:
Hmm, it's used in many places.

lanshunfang:
https://github.com/toss/es-toolkit/pull/1096

dayongkr:
If you use the `orderBy` function from the `es-toolkit/compat` package, which is compatible with Lodash, it can handle this case. I recommend using that function!

To manage issues, I’ll go ahead and close this for now, but feel free to reopen it anytime if further discussion is needed!

PR #1096, opened 2025-03-27
fix(orderBy): issue-1095: Support ordering string and number array

lanshunfang:
https://github.com/toss/es-toolkit/issues/1095

image


dayongkr:
Also, if you use the `orderBy` function from the `es-toolkit/compat` package, which is compatible with Lodash, it can handle this case. I recommend using that function!

To manage PRs, I’ll go ahead and close this for now, but feel free to reopen it anytime if further discussion is needed!

PR #1101, opened 2025-03-30
fix(pickBy): fix lodash compatibility

D-Sketon:


src/compat/object/pickBy.spec.ts
```diff
@@ -119,23 +189,6 @@ describe('pickBy', () => {
     expect(argsList).toEqual(expected);
   });
 
-  it(`iterates over own string keyed properties of objects`, () => {
```
raon0211, line 122: Oh, isn't this test case in lodash?

D-Sketon: https://github.com/lodash/lodash/blob/v5-wip/test%2Fiteration-methods.spec.js#L237-L237

D-Sketon: pickby was excluded (forInMethods)

raon0211 (approved):
Thanks!

PR #1109, opened 2025-04-03
ci: Add test step in Deno

raon0211:


src/predicate/isNode.ts
```diff
@@ -21,5 +21,5 @@ declare let process:
  * }
  */
 export function isNode(): boolean {
-  return typeof process !== 'undefined' && process?.versions?.node != null;
+  return process?.versions?.node != null;
```
dayongkr, line 24: Wouldn't this cause a reference error if the global process object is unavailable?

dayongkr (approved):

PR #1115, opened 2025-04-04
test: Fix flaky test dependent on wrongly assumed `Date.now()` accuracy

wojtekmaj:
`Date.now()` has millisecond precision but not microsecond precision, so the result might be 99ms due to rounding.
`Date.now()` is also *not* guaranteed to increment every millisecond - on some systems, it may tick every ~4ms.

You can see the flakiness e.g. here: https://github.com/toss/es-toolkit/actions/runs/14255327713

### DIFF
--- a/src/function/retry.spec.ts
+++ b/src/function/retry.spec.ts
@@ -28,7 +28,9 @@ describe('retry', () => {
     const end = Date.now();
     expect(result).toBe('success');
     expect(func).toHaveBeenCalledTimes(2);
-    expect(end - start).toBeGreaterThanOrEqual(delay);
+    // Date.now() has millisecond precision but not microsecond precision, so the result might be 99ms due to rounding.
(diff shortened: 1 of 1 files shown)


dayongkr:
Thank you for identifying this issue!

If millisecond precision is causing problems, what do you think about using the Performance API, which offers sub-millisecond precision?

wojtekmaj:
I think it just moves the issue to after the comma :D You'd still need to round the result before asserting it. I personally don't think it's worth it.

dayongkr:
​Since `Date.now()` has millisecond precision, using a timing function with microsecond precision should resolve the issue. If not, we could consider using `process.hrtime()`, which offers nanosecond precision.​

Although You've added comments explaining this, explicitly lowering the test case threshold from 100 to 99 feels a bit cautious.. 🙏

raon0211 (approved):

Issue #994, opened 2025-03-15
Add `unzipWith` to compat package


dayongkr:
## Description  
We have to add `unzipWith` to `compat/array` package.  

If there are no assignees, feel free to assign yourself and contribute. Thank you! 😊  

## References  

- [Lodash Document](https://lodash.com/docs/4.17.15#unzipWith)  
- [Test Case](https://github.com/lodash/lodash/blob/v5-wip/test/unzipWith.spec.js)

PR #1118, opened 2025-04-05
feat(unzipWith): Implement compat/unzipWith

L2HYUNN:
close: #994 

### Implement

I’m considering two different ways to implement the `unzipWith` function, and I’m debating whether to use a for loop or the map method.

Would love to hear thoughts on which approach is more suitable here.

**for-loop**
```ts
export function unzipWith<T>(
  array: T[][] | ArrayLike<ArrayLike<T>> | null | undefined,
  iteratee?: ((...args: any[]) => unknown) | null
): any[] {
  if (!isArrayLikeObject(array) || !array.length) {
    return [];
  }

  const unziped = isArray(array) ? unzipToolkit(array) : unzipToolkit(Array.from(array, value => Array.from(value)));

  if (!iteratee) {
    return unziped;
  }

  const result: any[] = new Array(unziped.length);

  for (let i = 0; i < unziped.length; i++) {
    const value = unziped[i];

    result[i] = iteratee(...value);
  }

  return result;
}
```

**map**
```ts
export function unzipWith<T>(
  array: T[][] | ArrayLike<ArrayLike<T>> | null | undefined,
  iteratee?: ((...args: any[]) => unknown) | null
): any[] {
  if (!isArrayLikeObject(array) || !array.length) {
    return [];
  }

  const unziped = isArray(array) ? unzipToolkit(array) : unzipToolkit(Array.from(array, value => Array.from(value)));

  if (!iteratee) {
    return unziped;
  }

  return unziped.map(item => iteratee(...item));
}
```

### Test 
image

### Bench
스크린샷 2025-04-05 오후 12 35 05


raon0211 (approved):
Perfect work!

PR #1119, opened 2025-04-05
fix(compat): add guard for take/takeRight/defaults/repeat/words

D-Sketon:


dayongkr (approved):
Thanks for your contribution

wo-o29:
@D-Sketon  

Hi! Could you please explain why the `takeRight` and `take` functions require the guard argument even though they are not from the compat module?

D-Sketon:
@wo-o29 I don't remember; I might have been confused about the compat and non-compat packages at the time.

wo-o29:
> @wo-o29 I don't remember; I might have been confused about the compat and non-compat packages at the time.

Thank you for your reply.

I noticed that in the documentation for takeRight, there is no explanation for the third parameter (guard), yet it exists in the internal implementation. I was curious about the intent behind this, which is why I asked the above question.

wo-o29:
@dayongkr 

Hello! I think `take` and `takeRight` should be removed because they are not compat modules and there is no explanation for the third argument (guard) in their respective documentation. I'd like to hear the maintainer's opinion🙏🏻

PR #1132, opened 2025-04-15
perf(toPairs): improve performance by pre-allocating array and using iterator values

ssi02014:
For `Map` and `Set`, I improved the performance of `toPairs/toPairsIn` with better logic than converting to `entries` via `Array.from(obj.entries())`.

This logic pre-allocates an array and uses Iterator values.

### AS-IS
oldNewPairs

### TO-BE
스크린샷 2025-04-15 오후 4 08 44

### DIFF
--- a/src/compat/_internal/mapToEntries.ts
+++ b/src/compat/_internal/mapToEntries.ts
@@ -1,0 +1,10 @@
+export function mapToEntries(map: Map<any, any>) {
+  const arr = new Array(map.size);
+  const keys = map.keys();
+  const values = map.values();
+
+  for (let i = 0; i < arr.length; i++) {
+    arr[i] = [keys.next().value, values.next().value];
+  }
+  return arr;
+}
--- a/src/compat/_internal/setToEntries.ts
+++ b/src/compat/_internal/setToEntries.ts
@@ -1,0 +1,10 @@
+export function setToEntries(set: Set<any>) {
+  const arr = new Array(set.size);
+  const values = set.values();
+
+  for (let i = 0; i < arr.length; i++) {
+    const value = values.next().value;
+    arr[i] = [value, value];
+  }
+  return arr;
+}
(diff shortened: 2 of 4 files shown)


src/compat/_internal/mapToEntries.ts
```diff
@@ -0,0 +1,10 @@
+export function mapToEntries(map: Map<any, any>) {
+  const arr = new Array(map.size);
+  const keys = map.keys();
+  const values = map.values();
+
+  for (let i = 0; i < arr.length; i++) {
+    arr[i] = [keys.next().value, values.next().value];
+  }
+  return arr;
+}
```
ssi02014, line 10: I ran performance tests with the three methods below.
We found that using an iterator was the best way to go.
```ts
const mapToEntries1 = (map: Map<any, any>) => {
  const arr = new Array(map.size);
  let i = 0;
  for (const [key, value] of map) {
    arr[i++] = [key, value];
  }

  return arr;
};

const mapToEntries2 = (map: Map<any, any>) => {
  const arr = new Array(map.size);
  const keys = map.keys();
  const values = map.values();

  for (let i = 0; i < map.size; i++) {
    arr[i] = [keys.next().value, values.next().value];
  }
  return arr;
};

const mapToEntries3 = (map: Map<any, any>) => {
  return Array.from(map.entries());
};
```

## Performance
map

src/compat/_internal/setToEntries.ts
```diff
@@ -0,0 +1,10 @@
+export function setToEntries(set: Set<any>) {
+  const arr = new Array(set.size);
+  const values = set.values();
+
+  for (let i = 0; i < arr.length; i++) {
+    const value = values.next().value;
+    arr[i] = [value, value];
+  }
+  return arr;
+}
```
ssi02014, line 10: I ran performance tests with the three methods below.
We found that using an iterator was the best way to go.

```ts
const setToEntries1 = (set: Set<any>) => {
  const arr = new Array(set.size);
  let i = 0;
  for (const value of set) {
    arr[i++] = [value, value];
  }

  return arr;
};

const setToEntries2 = (set: Set<any>) => {
  const arr = new Array(set.size);
  const values = set.values();

  for (let i = 0; i < set.size; i++) {
    const value = values.next().value;
    arr[i] = [value, value];
  }
  return arr;
};

const setToEntries3 = (set: Set<any>) => {
  return Array.from(set.entries());
};
```

## Performance
set

raon0211 (approved):
Thanks!

Issue #1016, opened 2025-03-15
Add `wrap` to compat package


dayongkr:
## Description  
We have to add `wrap` to `compat/function` package.  

If there are no assignees, feel free to assign yourself and contribute. Thank you! 😊  

## References  

- [Lodash Document](https://lodash.com/docs/4.17.15#wrap)  
- [Test Case](https://github.com/lodash/lodash/blob/v5-wip/test/wrap.spec.js)

PR #1071, opened 2025-03-23
feat(countBy): Implement compact/countBy

Jeong-Rae:
Resolves:  #1000

Implemented compact/countBy as described in #1000

I included the lodash test cases provided in the issue, along with a few additional ones.

PR #1135, opened 2025-04-17
feat(wrap): Implement `compact/wrap`

Jeong-Rae:
Resolves:  #1016

Implemented compact/wrap described in #1016

I included the lodash test cases provided in the issue, along with a few additional ones.


hsskey:
I noticed your commit messages are in Korean, but the project's contributing guidelines state "All communications in this repository will be in English." Would you mind using English for commit messages going forward?

src/compat/function/wrap.ts
```diff
@@ -0,0 +1,61 @@
...
+export function wrap(value: unknown, wrapper: (...args: unknown[]) => unknown): (...args: unknown[]) => unknown {
+  return function (this: unknown, ...args: unknown[]): unknown {
+    const wrapFn = isFunction(wrapper)
+      ? wrapper
+      : // eslint-disable-next-line @typescript-eslint/no-unused-vars
+        (fn: unknown, ..._args: unknown[]) => fn;
+
+    return wrapFn.apply(this, [value, ...args]);
+  };
+}
```
hsskey, line 61: Your type signatures define wrapper as always being a function, making this conditional check inconsistent with the type system. This creates an unreachable code path from TypeScript's perspective.

Consider:

1. Removing the check (preferred for strongly typed code)
2. Throwing an error instead of providing a fallback
3. Updating types to reflect that wrapper might not be a function

This would better align your runtime code with the TypeScript contract.

Jeong-Rae: Thank you for your valuable feedback. I aimed to preserve both Lodash’s TypeScript typing (treating wrapper as a function) and its JS compatibility (falling back to identity when it isn’t). That led me to enforce wrapper as a function in the implementation, but you’re right—it’s better not to. I’ll update it as follows
```suggestion
export function wrap(value: unknown, wrapper: unknown): (...args: unknown[]) => unknown {
  return function (this: unknown, ...args: unknown[]): unknown {
    const wrapFn = isFunction(wrapper)
      ? (wrapper as (value: unknown, ...args: unknown[]) => unknown)
      : identity;

    return wrapFn.apply(this, [value, ...args]);
  };
}
```
Thanks again for the great review!

Jeong-Rae:
> I noticed your commit messages are in Korean, but the project's contributing guidelines state "All communications in this repository will be in English." Would you mind using English for commit messages going forward? Same issue in #1071 as well.

You’re absolutely right. I assumed these commits would be squashed and overlooked it. Writing commit messages in English is indeed more appropriate, and I’ll make sure to follow that going forward.

raon0211 (approved):

PR #1136, opened 2025-04-17
fix(maxBy, minBy): return undefined for empty array inputs

umsungjun:
- Added an early return undefined when the input array is empty in both maxBy and minBy functions. This prevents the function from returning the initial element incorrectly when there are no items to evaluate.

- Updated the return type of maxBy to explicitly allow undefined. This reflects the correct behavior when the input array has no elements.

**Benchmarks**
스크린샷 2025-04-17 오후 9 09 47

### DIFF
--- a/src/array/maxBy.ts
+++ b/src/array/maxBy.ts
@@ -60,7 +60,11 @@ export function maxBy<T>(items: readonly T[], getValue: (element: T) => number):
  *   x => x.age
  * ); // Returns: { name: 'john', age: 30 }
  */
-export function maxBy<T>(items: readonly T[], getValue: (element: T) => number): T {
+export function maxBy<T>(items: readonly T[], getValue: (element: T) => number): T | undefined {
+  if (items.length === 0) {
+    return undefined;
+  }
+
   let maxElement = items[0];
   let max = -Infinity;
 
--- a/src/array/minBy.ts
+++ b/src/array/minBy.ts
@@ -61,6 +61,10 @@ export function minBy<T>(items: readonly T[], getValue: (element: T) => number):
  * ); // Returns: { name: 'joe', age: 26 }
  */
 export function minBy<T>(items: readonly T[], getValue: (element: T) => number): T | undefined {
+  if (items.length === 0) {
+    return undefined;
+  }
+
   let minElement = items[0];
   let min = Infinity;
 


hsskey:
Nice fix! Just a small suggestion: since this PR changes the behavior for empty arrays, can you add a test case for that? It'll help prevent regressions in the future and document the expected behavior more clearly. Thanks 🙌

umsungjun:
To. @hsskey 

Thanks for the suggestion! 🙌
The test case for the empty array scenario was actually already present in minBy.spec.ts and maxBy.spec.ts.
Let me know if there’s anything else I should adjust!

스크린샷 2025-04-18 오후 2 53 33

hsskey:
Thanks for showing me! I see the tests were already there.

hsskey:
> To. @hsskey
> 
> Thanks for the suggestion! 🙌 The test case for the empty array scenario was actually already present in minBy.spec.ts and maxBy.spec.ts. Let me know if there’s anything else I should adjust!
> 
> 스크린샷 2025-04-18 오후 2 53 33

Thanks for showing me! I see the tests were already there.

raon0211:
Hey! Thanks for opening the pull request. I appreciate the effort, but I believe the current behavior is actually correct.

In JavaScript, `Math.min()` returns `Infinity` when called with no arguments, and `Math.max()` returns `-Infinity`. This might seem odd at first, but it makes sense when you think about how these functions work.

For example, `Math.min(1, 2, 3)` will always be greater than or equal to `Math.min(1, 2, 3, x)` for any number `x`. As you add more values, the result can only stay the same or get smaller.

Now, take that idea to the extreme: if you call `Math.min()` with no values at all, the result should be greater than or equal to `Math.min(x)` for any `x`. That means it must return the largest possible value — which is `Infinity`.

umsungjun:
Thank you for the feedback! @raon0211 

I’ve updated both maxBy and minBy to return undefined when the input array is empty.
The reason for this change is that the test cases expect undefined when the array is empty, so I added the condition to match that expectation.

스크린샷 2025-04-19 오후 10 51 30


I understand your explanation about Math.min() and Math.max() returning Infinity or -Infinity when called without arguments.
However, in this case, since minBy and maxBy are designed to return an element from the array, I believe returning undefined when the array is empty is the correct behavior.

If I’ve misunderstood anything, please let me know!

umsungjun:
Thank you for the feedback!

I understand now.
```tsx
const emptyArray = [];
console.log(emptyArray[Infinity]); // undefined
```
It returns undefined.

raon0211:
Thanks for your understanding :)

Issue #1092, opened 2025-03-27
Support for `isWindow`


ssi02014:
In client development, we sometimes need to `type guard` whether a certain element is a `window` or not. I think this would be a useful utility function for such cases, why not support it in `es-toolkit`? 

## Suggestion
```ts
export function isWindow(element: unknown): element is Window {
  if (typeof window === 'undefined') {
    return false;
  }

  return element === window;
}
```

dayongkr:
We already have `isBrowser`, so in what kind of situation would we need `isWindow` instead?

ssi02014:
@dayongkr 
They serve different purposes. 

While `isBrowser` checks that the current environment is a browser, `isWindow` is a function that validates that the given argument is a `window` object and provides type guards. 

For example, it could be used when writing logic related to scroll events, such as the following 

```ts
const getScorollPosition = (element: HTMLElement | Window) => {
  if (isWindow(element)) {
    // Ensure element is a window
    // element: Window
    return {
      scrollX: element.scrollX,
      scrollY: element.scrollY,
    };
  }

// element: Element
  return {
    scrollX: element.scrollLeft,
    scrollY: element.scrollTop,
  };
};
```

It looks like we could reuse isBrowser. 
```ts
export function isWindow(element: unknown): element is Window {
  if (!isBrowser()) {
    return false;
  }

  return element === window;
}
```

dayongkr:
Got it! I checked the `is-window` package on npm and it seems like there's a solid user base for it, so I’ll bring it up in our regular meeting next Thursday.

Since we don’t have a formal process yet for adding new functions, things are a bit delayed and I apologize for not being able to communicate more clearly.

Really appreciate your contribution!

dayongkr:
I forgot to include the issue in the agenda by mistake. I sincerely apologize for the oversight, and I will make sure to add it to the agenda next week.

dayongkr:
Thank you for the great idea! Could you please open a PR for this?

ssi02014:
@dayongkr Yes~! Let me work on it.

PR #1143, opened 2025-04-19
feat: add isWindow function

ssi02014:
issue: https://github.com/toss/es-toolkit/issues/1092
I worked on adding the isWindow function.


vitest.config.mts
```diff
@@ -4,6 +4,7 @@ import packageJson from './package.json';
 export default defineConfig({
   test: {
     name: packageJson.name,
+    environment: 'jsdom',
```
ssi02014, line 7: Set up jsdom with vitest's environment for accurate testing.

ssi02014: We can do that and have a more accurate test for `isBrowser` as well.

src/predicate/isWindow.spec.ts
```diff
@@ -0,0 +1,23 @@
...
+    expect(isWindow(window.document)).toBeFalsy();
+    expect(isWindow({})).toBeFalsy();
+    expect(isWindow(div)).toBeFalsy();
+    expect(isWindow(1)).toBeFalsy();
+    expect(isWindow('')).toBeFalsy();
+  });
+
+  it('should return "false" if the element is not the `Window` object', () => {
+    const windowSpy = vi.spyOn(global, 'window', 'get');
+    windowSpy.mockImplementation(() => undefined as unknown as typeof globalThis.window);
```
ssi02014, line 19: I set the window object to undefined after mocking it and wrote the test code assuming it was a server environment.

vitest.config.mts
```diff
@@ -4,6 +4,7 @@ import packageJson from './package.json';
 export default defineConfig({
   test: {
     name: packageJson.name,
+    environment: 'happy-dom',
```
ssi02014, line 7: Set up `happy-dom` with vitest's environment for accurate testing. 

We can do that and have a more accurate test for `isBrowser` as well.

ssi02014: We could also use `jsdom`, but we already had `happy-dom` installed, and `jsdom` has issues with the `Blob.size` test.

src/predicate/isWindow.spec.ts
```diff
@@ -0,0 +1,11 @@
...
+import { isWindow } from './isWindow.ts';
+
+describe('isWindow', () => {
+  it('should return "false" if the element is not the `Window` object', () => {
+    expect(isWindow({})).toBeFalsy();
+    expect(isWindow([])).toBeFalsy();
+    expect(isWindow(1)).toBeFalsy();
+    expect(isWindow('')).toBeFalsy();
+  });
+});
```
ssi02014, line 11: When I tested with both environments of `vitest` set to `jsdom` and `happy-dom`, both environments caused issues with `isWindow` and other separate test files. So, first of all, I only wrote the test code to pass.

ssi02014: 1. jsdom: https://github.com/toss/es-toolkit/actions/runs/14547617444
    - tests/check-dist.spec.ts
    - src/object/clone.spec.ts 
  
스크린샷 2025-04-19 오후 6 25 05

2. happy-dom: https://github.com/toss/es-toolkit/actions/runs/14547701146
    - tests/check-dist.spec.ts
 
스크린샷 2025-04-19 오후 6 26 36

ssi02014: If you have any good advice for test code, let us know! 🙏

src/predicate/isWindow.ts
```diff
@@ -0,0 +1,9 @@
+import { isBrowser } from './isBrowser.ts';
+
+export function isWindow(element: unknown): element is Window {
```
hsskey, line 3: I’d suggest slightly enhancing the implementation to support edge cases like **cross-realm windows** (e.g., iframes).  
Using a structural check like `window.window === window` is a safe and well-documented pattern ([MDN](https://developer.mozilla.org/en-US/docs/Web/API/Window/window)) for verifying if a value is a `Window` object, even across different contexts.

```ts
import { isBrowser } from './isBrowser';

export function isWindow(element: unknown): element is Window {
  if (!isBrowser()) {
    return false;
  }

  // Case 1: Same-realm window
  if (element === window) {
    return true;
  }

  // Case 2: Cross-realm iframe window
  try {
    return (
      typeof element === 'object' &&
      element !== null &&
      'window' in element &&
      (element as Window).window === element
    );
  } catch {
    return false;
  }
}

ssi02014: @hsskey 
What you're suggesting requires some thought.

I was thinking of the `isWindow` function as a function that simply checks if the given argument is a window object.

I didn't consider the edge case of checking if it is a `window` object in different contexts.

One of es-toolkit's main values is `simplicity`, so it's a good suggestion, but it's up to the maintainer to decide if edge cases should be handled.
- https://github.com/toss/es-toolkit/blob/main/.github/CONTRIBUTING.md#12-simplicity

hsskey: @ssi02014 
You're totally right that simplicity is a core value, and using `element === window` does keep things very clean and straightforward.

From a technical perspective, there's also the `window.window === window` pattern, which can help cover a few more edge cases like iframes. It's still quite readable, and once understood, it’s a relatively simple way to provide a bit more safety.

That said, I totally get that it might go beyond what the toolkit aims to cover by default. Personally, I lean slightly toward including the structural check since it’s lightweight and could help avoid subtle issues later on — but I'm happy to go with whatever direction aligns best with the project goals.

Appreciate the thoughtful discussion

dayongkr:
There are no further comments, so I'm closing this PR for now.
Feel free to reopen it if you have any additional thoughts or updates.

PR #1149, opened 2025-04-23
feat(compat/object): add forIn and forInRight

dayongkr:


raon0211 (approved):

PR #1152, opened 2025-04-24
fix(throttle): fix throttle with both leading and trailing

seungrodotlee:
Fix throttle with both leading and trailing.

This fixes an issue where the function was being invoked twice — once on leading and once on trailing — even though it should only be called once if triggered within the same `throttleMs` window.


raon0211 (approved):
Thanks for fixing our bug!

PR #1159, opened 2025-05-01
feat(bench): add vitest benchmark for ary function

umsungjun:
Added the missing benchmark for the ary function from es-toolkit in ary.bench.ts

스크린샷 2025-05-01 오후 3 13 53


dayongkr (approved):

Issue #1053, opened 2025-03-15
Different behaviour with lodash words


D-Sketon:
lodash:
```js
_.words('1st-2nd-3rd')  // -> ['1st', '2nd', '3rd']
_.words("I don't+can't-won't") // -> ['I', "don't", "can't", "won't"]
_.words("नमस्ते नमस्ते") // -> ['नमस्ते', 'नमस्ते']
```

es-toolkit:
```js
_.words('1st-2nd-3rd')  // -> ['1', 'st', '2', 'nd', '3', 'rd']
_.words("I don't+can't-won't") // -> ['I', 'don', 't', 'can', 't', 'won', 't']
_.words("नमस्ते नमस्ते") // -> ['नमस', 'त', 'नमस', 'त']
```

raon0211:
😭  I guess we should fix this.

PR #1163, opened 2025-05-03
fix(words): support ordinal number/contractions

D-Sketon:
close https://github.com/toss/es-toolkit/issues/1053


raon0211 (approved):
Thanks!

PR #1164, opened 2025-05-03
chore(at): fix return type in comments

Seung-wan:
- return type of `at` function is `T[]`, but in comments it was a `Array<T | undefined>`.


raon0211 (approved):
Thanks for pointing this out!

PR #1167, opened 2025-05-05
fix(compat): align compat HOF behavior with lodash for sparse arrays

### LINKED ISSUE #1161: The different behaviors of iteration methods between lodash and es-toolkit/compat when handling sparse arrays
lodash
```ts
_.every([1,,3], (i) => {console.log(i);return true;})
// => 1
// => undefined
// => 3
```
es-toolkit/compat
```ts
every([1,,3], (i) => {console.log(i);return true;})
// => 1
// => 3
```

greenheadHQ:
close #1161

I think we only need to fix `filter`, `every`, and `some` methods. The rest of the higher-order functions in `compat` use `for` loops instead of JavaScript’s built-in higher-order functions, which skip empty slots in sparse arrays, so I believe they can handle those empty slots properly.

### DIFF
--- a/src/compat/array/every.spec.ts
+++ b/src/compat/array/every.spec.ts
@@ -158,4 +158,12 @@ describe('every', () => {
     expect(every(objects, [0, 1])).toBe(false);
     expect(every(objects, [Symbol.for('c'), 1])).toBe(true);
   });
+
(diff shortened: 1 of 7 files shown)


raon0211 (approved):
Thanks for fixing this!

Issue #1012, opened 2025-03-15
Add `memoize` to compat package


dayongkr:
## Description  
We have to add `memoize` to `compat/function` package.  

If there are no assignees, feel free to assign yourself and contribute. Thank you! 😊  

## References  

- [Lodash Document](https://lodash.com/docs/4.17.15#memoize)  
- [Test Case](https://github.com/lodash/lodash/blob/v5-wip/test/memoize.spec.js)

jch1223:
@dayongkr 
I'll handle this issue!

PR #1171, opened 2025-05-08
feat(memoize): add memoize to compat

jch1223:
I've tried to modify it using class syntax. Is this the right approach?

closed: #1012 

image

### DIFF
--- a/benchmarks/performance/memoize.bench.ts
+++ b/benchmarks/performance/memoize.bench.ts
@@ -1,8 +1,10 @@
 import { bench, describe } from 'vitest';
 import { memoize as memoizeToolkit_ } from 'es-toolkit';
+import { memoize as memoizeToolkitCompat_ } from 'es-toolkit/compat';
 import { memoize as memoizeLodash_ } from 'lodash';
 
 const memoizeToolkit = memoizeToolkit_;
+const memoizeToolkitCompat = memoizeToolkitCompat_;
 const memoizeLodash = memoizeLodash_;
 
 describe('memoize', () => {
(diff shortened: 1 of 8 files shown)


shren207 (commented):
```
-----------------------|---------|----------|---------|---------|-----------------------
File                   | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s
-----------------------|---------|----------|---------|---------|-----------------------
memoize.ts             |   46.23 |    51.61 |   56.52 |   47.77 | 29-57,102-199,415-429
```

The test coverage of your code is currently too low. While es-toolkit/compat may have a slight performance impact and a larger bundle size compared to the original es-toolkit, this is because its primary goal is not performance, but complete functional parity with lodash. (you might have to check [this reference](https://es-toolkit.slash.page/compatibility.html) about `compat`)  Therefore, the test cases for es-toolkit/compat must include, as much as possible, all test cases used for the equivalent lodash methods. Except for a few specific exceptions, the coverage must reach 100% across all areas.

Sometimes, even after including all lodash test cases, you may still fail to achieve 100% coverage. In such cases, you may need to add additional test cases to increase coverage or simplify the implementation logic. This is currently the case with your code.

dayongkr (approved):
Thanks for the first contribution — welcome aboard! 🎉

As @shren27 mentioned, the test code doesn't fully cover all lines and include unnecessary docs. But, the overall implementation looks solid, so I've gone ahead and merged it. I'll make a few adjustments to the code and tests afterward to tidy things up.

Thanks again!!

> One more minor suggestion — it would be great if you could write your commit messages in English. This helps improve clarity for other contributors and users.

dayongkr:
The original code followed lodash, but lodash separates storage by key type unnecessarily.

Since [Map uses the SameValue](https://tc39.es/ecma262/multipage/keyed-collections.html#sec-map.prototype.get) strategy (e.g., `'1'` and `1` are treated as different keys), and we're targeting modern JS, we don’t need to support environments without Map. So there’s no need to create custom classes or separate cache storage based on key types.

Also, the fact that Lodash uses SameValueZero internally isn’t really an issue either, since [CanonicalizeKeyedCollectionKey](https://tc39.es/ecma262/multipage/keyed-collections.html#sec-canonicalizekeyedcollectionkey) makes the behavior effectively similar in practice — differences in stored key values aren’t a concern for caching.

I’ve simplified the implementation to just use `Map`.

Left this comment for reference — thanks again for your contribution!

PR #1173, opened 2025-05-09
fix[dropRight]: Improve code clarity while preserving behavior

jiji-hoon96:
## Description
This PR improves the dropRight function implementation by making the code more readable and explicit while preserving the existing behavior expected by the tests. The current implementation uses a confusing approach with Math.min(-itemsCount, 0) that makes it hard to understand the function's intent and could lead to maintenance issues.

## Changes Made
Restructured the function to explicitly handle different cases

- NaN values are treated as 0 (returns full array copy)
- Non-integer values are floored to ensure consistent behavior
- Negative or zero values return a full array copy
- Values greater than array length return an empty array

Added additional test cases to ensure the behavior is properly specified and maintained

## Bug Details
The original implementation

```ts
export function dropRight<T>(arr: readonly T[], itemsCount: number): T[] {
  itemsCount = Math.min(-itemsCount, 0);

  if (itemsCount === 0) {
    return arr.slice();
  }

  return arr.slice(0, itemsCount);
}
```

This implementation had several issues

Math.min(-itemsCount, 0) is always less than or equal to 0, making the logic difficult to understand
When itemsCount is positive, arr.slice(0, itemsCount) actually works because -itemsCount becomes negative, and slice() interprets negative indices as counting from the end
This logic works by coincidence rather than being clearly intentional

## Implementation Details
The new implementation preserves all the behaviors of the original function but makes them explicit and easier to understand

```ts
export function dropRight<T>(arr: readonly T[], itemsCount: number): T[] {
  if (Number.isNaN(itemsCount) || itemsCount <= 0) {
    return arr.slice();
  }

  itemsCount = Math.floor(itemsCount);

  if (itemsCount >= arr.length) {
    return [];
  }

  return arr.slice(0, arr.length - itemsCount);
}
```

## Test Results
All existing tests continue to pass with this implementation. Additionally, I've added two test cases

- Edge case with very large numbers
- Explicit handling of NaN input values

These additional tests help ensure the function's behavior is well-defined across all input types.

스크린샷 2025-05-09 오후 1 45 52

### DIFF
--- a/src/array/dropRight.spec.ts
+++ b/src/array/dropRight.spec.ts
@@ -19,4 +19,12 @@ describe('dropRight', () => {
   it('should return empty array if itemsCount >= arr.length', () => {
     expect(dropRight([1.2, 2.3, 3.4], 4)).toEqual([]);
   });
+
+  it('should correctly handle edge cases with very large numbers', () => {
+    expect(dropRight([1, 2, 3], Number.MAX_SAFE_INTEGER)).toEqual([]);
+  });
+
+  it('should handle NaN values as 0', () => {
+    expect(dropRight([1, 2, 3], Number.NaN)).toEqual([1, 2, 3]);
+  });
 });
--- a/src/array/dropRight.ts
+++ b/src/array/dropRight.ts
@@ -15,11 +15,11 @@
  * // result will be [1, 2, 3] since the last two elements are dropped.
  */
 export function dropRight<T>(arr: readonly T[], itemsCount: number): T[] {
-  itemsCount = Math.min(-itemsCount, 0);
+  itemsCount = Math.max(itemsCount, 0);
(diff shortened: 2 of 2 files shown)


raon0211:
Thanks for the pull request! We're going to stick with the current approach because:

- Negative indices are standard in `Array#slice` ([MDN reference](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice#parameters)) and other array methods
- `NaN` handling isn't part of the ECMAScript spec, and we avoid adding complexity for edge cases per our [design principles](https://github.com/toss/es-toolkit/blob/main/.github/CONTRIBUTING.md#13-simplicity)

Appreciate you taking the time to contribute!

PR #1180, opened 2025-05-11
feat(findLastKey): add findLastKey to compat

### LINKED ISSUE #1031: Add `findLastKey` to compat package
## Description  
We have to add `findLastKey` to `compat/object` package.  

If there are no assignees, feel free to assign yourself and contribute. Thank you! 😊  

## References  

- [Lodash Document](https://lodash.com/docs/4.17.15#findLastKey)  
- [Test Case](https://github.com/lodash/lodash/blob/v5-wip/test/find-methods.spec.js)

myungjuice:
close #1031 
added compact/findLastKey described in #1031 

### Test
spec

### Benchmarks
bench


benchmarks/performance/findLastKey.bench.ts
```diff
@@ -0,0 +1,35 @@
+import { bench, describe } from 'vitest';
+import { findLastKey as findLastKeyLodash } from 'lodash';
+// import { findLastKey as findLastKeyCompatToolkit } from 'es-toolkit/compat';
+import { findLastKey as findLastKeyCompatToolkit } from '../../src/compat/object/findLastKey';
```
shren207, line 4:
```suggestion
import { bench, describe } from 'vitest';
import { findLastKey as findLastKeyLodash } from 'lodash';
import { findLastKey as findLastKeyCompatToolkit } from 'es-toolkit/compat';
```

If the reason you used a relative import path instead of the absolute `es-toolkit/compat` was that the benchmark for `es-toolkit/compat` couldn’t be measured—resulting in `NaN`—then this is due to Vitest’s cache. If you **delete** the `node_modules` folder in your project’s root and rerun the benchmark tests, the speed of the module imported via `es-toolkit/compat` will be reported correctly.

Pasted Graphic 3

docs/reference/object/findLastKey.md
shren207, line 1: Documentation for the `es-toolkit/compat` utility functions, rather than for `es-toolkit`, should be located in the `docs/reference/compat/` directory.

myungjuice:
@shren207 
Thank you for the review.
I’ve moved the previously written documentation to the docs/reference/compat directory.
I also updated the benchmark code to import findLastKey using an absolute path (es-toolkit/compat).

Please let me know if everything looks good 🙏

dayongkr (approved):
Thanks for the contribution!!

dayongkr:
It looks like you referred to the type of `compat/findKey`, but since we're using the types directly from `@types/lodash`, I updated the type accordingly. Also, since we use an `iteratee` function, I replaced it with that. If you're curious about the changes, feel free to check out [this commit](https://github.com/toss/es-toolkit/commit/a93b7ed06092b74efe5365d100fe602c967046f6)!

PR #1182, opened 2025-05-11
docs: fix wrong url 

shren207:
Pasted Graphic 2

Fixed a bug where clicking the `frontend-fundamentals` link from the floating banner at the bottom right of the page on https://es-toolkit.slash.page/ would lead to a 404 page when the site locale was set to English.


dayongkr (approved):
Nice catching!!

Thanks for the contribution!

PR #1185, opened 2025-05-15
fix(findKey): correct typo 'Partail' to 'Partial' in type annotation

myungjuice:
Fix a typo in the `findKey` function's type annotation (`Partail` → `Partial`).

### DIFF
--- a/src/compat/object/findKey.ts
+++ b/src/compat/object/findKey.ts
@@ -27,7 +27,7 @@ export function findKey<T extends Record<any, any>>(
  *
  * @template T - The type of the object.
  * @param {T | null | undefined} obj - The object to inspect.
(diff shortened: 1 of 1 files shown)


myungjuice:
Hi team,
I only made a small typo fix in the documentation, but I'm seeing a type-checking error when running `yarn tsc --noEmit`.

It seems the error is coming from `src/compat/util/bindAll.spec.ts`, where the `bindAll` function is called with `args` that has the type `IArguments`. According to the error message, this type isn't assignable to `string | string[]`.

My guess is that changing the call to `bindAll(object, Array.from(args));` might resolve the issue, but since this isn't related to the change I made, I didn't include it in this PR.

Would you mind taking a quick look when you have time? Thank you!

image

raon0211 (approved):
Thanks for finding this!

PR #1188, opened 2025-05-18
test(math): add test cases for min and max with null and undefined in…

umsungjun:
- Added test cases to ensure `min` and `max` return `undefined` when passed `null` or `undefined`.
- Referenced implementation from: https://github.com/toss/es-toolkit/commit/04776a4b067083c92c01515bb8bf7586cdde2447

스크린샷 2025-05-18 오후 2 46 29

### DIFF
--- a/src/compat/math/max.spec.ts
+++ b/src/compat/math/max.spec.ts
@@ -52,4 +52,9 @@ describe('max', () => {
   it('should return undefined when skipping all values', () => {
     expect(max([Symbol('a'), null, NaN])).toBe(undefined);
   });
+
+  it('should return undefined when input is null or undefined', () => {
+    expect(max(null)).toBe(undefined);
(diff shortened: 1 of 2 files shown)


dayongkr:
We're adding test cases based on what the function is supposed to do.

Do you think we still need to test with `null` or `undefined` even if the function isn’t expected to handle them? 🙏

umsungjun:
That makes sense. Since null and undefined are allowed in the type definition but not considered part of the expected or valid use cases, it seems reasonable to omit tests for those cases. 

Thank you for the explanation!

dayongkr:
Seems like we’re thinking the same thing, so I’ll close the PR for now. If you feel like it needs more discussion, feel free to reopen. Thanks!

umsungjun:
Thank you always!

PR #1191, opened 2025-05-19
feat: expose api in the browser console

### LINKED ISSUE #1186: Support global API access in console for easier testing
Hello Toss OSS Team.

I’d like to suggest a fun feature. Popular libraries like lodash and dayjs expose their APIs as global objects in the browser’s developer console, allowing developers to experiment with them directly. (Please refer to the attached video.)

Have you ever considered adding such a feature?
If you find the idea appealing but haven’t had the time to work on it, I’d be more than happy to take on the task myself.

https://github.com/user-attachments/assets/ecc834b4-bf32-4d2e-900f-62c37bc5ea24

shren207:
close #1186

image

the existing bundler configuration was already well-structured, so the implementation went very smoothly.

### 1.  Improving the console guide text
The current console guide text does not tell users that they can call APIs via `_.` or `__.`. I tried adding a hint like:

image

But the result looked little bit verbose, so I didn’t include it in the commit. 😅

Any suggestions for a cleaner, more discoverable console text are welcome! 

### 2. Managing the `BUNDLE_PATH` constant

Since both `es-toolkit` and `es-toolkit/compat` bundle paths needed to be referenced from two separate files, I extracted the paths into a shared file at `/configs/bundlePath.ts` and exported them from there.

I couldn’t find any existing code that clearly shows how global constants are typically managed in this project, so I went with this approach. 
If there’s a preferred or recommended pattern for managing constants in this project, I’d appreciate any guidance!

### 3. Exposing both `es-toolkit` and `es-toolkit/compat` in the console
Is it necessary to expose both es-toolkit and es-toolkit/compat on the global object, or should we expose only one of them? 

### 4. Alias names `_` and `__`
Are `_` and `__` intuitive choices for the global aliases? If there are clearer or more conventional names, I’d love to hear them.


dayongkr:
I haven’t reviewed it in detail yet, but I’d like to decouple it from the package build as much as possible.
Something like:

```html
<script type="module">
import * as es from "https://cdn.jsdelivr.net/npm/es-toolkit@latest/+esm";
window.es = es;
</script>
```

This way, we can load the latest build directly in the browser. What do you think about going in this direction?

shren207:
@dayongkr That sounds like a much better approach! I found that `/docs/.vitepress/shared.mts` already contains code for loading external JS scripts in a way that’s compatible with `VitePress`.
With that as a reference, I implemented your suggested approach — and it worked exactly as intended!
Thank you for the simple yet excellent idea!

dayongkr (approved):
Thanks for the contributions!!

PR #1197, opened 2025-05-28
test(identity): add benchmark comparing es-toolkit and lodash

umsungjun:
Added a benchmark test for the `identity` function to compare performance between `es-toolkit` and `lodash`.

스크린샷 2025-05-28 오후 9 41 00


raon0211 (approved):
Thanks!

PR #1200, opened 2025-06-06
feat: packages/codemod-lodash-to-es-toolkit

hyesungoh:
Hello!

I've added a simple package that transforms lodash import statements to `es-toolkit/compat`.

Please consider this as a PoC-level PR to gather feedback from the maintainers 🙇 

> Feel free to close this if needed 👍 

Currently, it's implemented to transform even subpath imports to `es-toolkit/compat`, but I think this could be made configurable through options in the future.

---

안녕하세요!

lodash import 문을 `es-toolkit/compat`으로 바꾸는 간단한 패키지를 추가했어요

메인테이너분들의 의견을 여쭙는 PoC 정도의 PR이라고 생각해주시면 좋을 거 같아요 🙇 

> 편히 닫으셔도 무방해요 👍 

지금은 subpath 임포트까지 `es-toolkit/compat`으로 변경하도록 구현해두었는데 옵션을 통해 추가할 수 있을 것이라고 생각해요

PR #1204, opened 2025-06-11
feat: packages/gritql-lodash-to-es-toolkit

hyesungoh:
Hi team!

As mentioned in this comment (https://github.com/toss/es-toolkit/pull/1200#issuecomment-2951924569), I thought GritQL would also be a good approach, so I've set up a simple structure for it.

> Please feel free to close this PR when does need anymore! 🙏 

However, the `grit auth login` command doesn't work on my local machine, so I wasn't able to actually test it 😢

---

I couldn't find documentation about testing in the official GritQL docs, so it seems like a separate testing environment would need to be set up to make testing possible.

> I think this could also be considered a tradeoff point 🤔

Personally, I think codemod would be better due to the maturity of the library and community, but GritQL is definitely a tool with clear advantages as well.

I'd appreciate it if the maintainers could make the judgment call. Thank you 👍

### DIFF
--- a/packages/gritql-lodash-to-es-toolkit/README.md
+++ b/packages/gritql-lodash-to-es-toolkit/README.md
@@ -1,0 +1,104 @@
+# @es-toolkit/gritql-lodash-to-es-toolkit
+
+🔧 **GritQL patterns for migrating from lodash to es-toolkit/compat**
+
+This package provides [GritQL](https://docs.grit.io/) patterns to automatically transform lodash import statements to es-toolkit/compat imports. GritQL is a powerful query language for searching and modifying source code, offering superior performance and composability compared to traditional codemods.
+
+## 📖 Step-by-Step Guide
+
+### 1. Install Grit CLI
+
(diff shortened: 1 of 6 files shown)


packages/gritql-lodash-to-es-toolkit/README.md
```diff
@@ -0,0 +1,104 @@
...
+cd your-project
+grit init
+```
+
+### 3. Download patterns from GitHub
+
+```bash
+# Download the latest patterns directly from GitHub
+curl -fsSL https://raw.githubusercontent.com/toss/es-toolkit/main/packages/gritql-lodash-to-es-toolkit/.grit/grit.yaml -o .grit/grit.yaml
+```
```
hyesungoh, line 29: Can this approach achieve the same (or recommended) effect as a published package? 🤔

hyesungoh:
it moves other repo https://github.com/toss/es-toolkit-codemod/pull/1

PR #1230, opened 2025-06-23
fix(debounce): add export for debounceoptions type

eranshcloud:
This PR will expose debounceOptions type, to allow passing leading/trailing when using typescript.


dayongkr:
We're not planning to export it for now. 

Is it possible to extract and use the type with Parameters<typeof debounce>[2]?

dayongkr:
Since there has been no response for a long time, I will close it. If you need it, please reopen it.

PR #1233, opened 2025-06-25
fix(compat/defaults): handle `undefined` and `null` sources in `compat/defaults`

hwibaski:
## Summary

Before, the `sources` parameter had a strict type, so unless a user intentionally hacked around the type system, `null` or `undefined` values were never passed.
Later, while making the type compatible with Lodash, it became possible to pass nil (`null` or `undefined`) values.
When this happened, `compat/defaults` would throw an error, while Lodash’s `defaults` would simply ignore those values without a problem.

This change fixes that issue by making `compat/defaults` handle nil values the same way as Lodash — safely skipping them without errors.

## Before Change

The following code would throw an error when a null or undefined source parameter is provided:

```ts
import { defaults as esDefaults } from "es-toolkit/compat";

function main() {
  esDefaults({ a: 1 }, null);
}

main();
```

image

> Note: In Lodash’s defaults, passing null or undefined as a source does not cause an error — it simply ignores those values.
The current behavior in es-toolkit is inconsistent with that.

## After Change

The defaults function has been updated to safely ignore null or undefined sources.
If a nil (null/undefined) source is passed, it will simply be skipped without throwing an error.

```ts
import { defaults as esDefaults } from "es-toolkit/compat";

function main() {
  esDefaults({ a: 1 }, null); // returns { a: 1 }
}

main();
```

## Testing

image

## Coverage

image

### DIFF
--- a/src/compat/object/defaults.spec.ts
+++ b/src/compat/object/defaults.spec.ts
@@ -75,4 +75,16 @@ describe('defaults', () => {
   it('should match the type of lodash', () => {
     expectTypeOf(defaults).toEqualTypeOf<typeof defaultsLodash>();
   });
+
+  it('should not throw an error when a source is `undefined`', () => {
+    const source = undefined;
+    const actual = defaults({ a: 1 }, source);
+    expect(actual).toEqual({ a: 1 });
+  });
+
+  it('should not throw an error when a source is `null`', () => {
+    const source = null;
+    const actual = defaults({ a: 1 }, source);
+    expect(actual).toEqual({ a: 1 });
+  });
 });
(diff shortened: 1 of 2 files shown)


hwibaski:
@dayongkr
Hi, I noticed this PR hasn't been reviewed for a couple of weeks.
If you have a moment, could you review it?
I just mentioned you directly in case you might have missed it.
If it was a bother, I won’t mention you directly next time — sorry about that!

raon0211 (dismissed):
Sorry for the late review. Really appreciate your work!

raon0211 (approved):

PR #1240, opened 2025-07-01
docs(fill): enhance fill function documentation

ssi02014:
I fix it because the fill function jsdoc type is not written correctly.


src/array/fill.ts
```diff
@@ -5,10 +5,10 @@
  * start index up to the end index (non-inclusive). If the start or end indices are not provided, it defaults to filling the
  * entire array.
  *
- * @template T, U
- * @param {Array<T | U>} array - The array to fill.
- * @param {U} value - The value to fill the array with.
- * @returns {Array<T | U>} The array with the filled values.
+ * @template T - The type of the value to fill the array with.
```
ssi02014, line 8: I added most of the comments because they contain template descriptions.

raon0211 (approved):
Thanks!

PR #1242, opened 2025-07-01
fix(pull): enhance pull function type definition

ssi02014:
I think it's better to set the type of `valuesToRemove` to a narrower type instead of a broader type like `unknwon`.

```ts
interface LoDashStatic {
    /**
     * Removes all provided values from array using SameValueZero for equality comparisons.
     *
     * Note: Unlike _.without, this method mutates array.
     *
     * @param array The array to modify.
     * @param values The values to remove.
     * @return Returns array.
     */
    pull<T>(array: T[], ...values: T[]): T[];
    /**
     * @see _.pull
     */
    pull<T>(array: List<T>, ...values: T[]): List<T>;
}
interface Collection<T> {
    /**
     * @see _.pull
     */
    pull(...values: T[]): Collection<T>;
}
```

Even existing lodash does not allow types like `unknown`.


benchmarks/performance/pull.bench.ts
```diff
@@ -10,7 +10,7 @@ describe('pull array size 100', () => {
   const even = [...Array(50)].map((_, i) => i * 2);
 
   bench('es-toolkit/pull', () => {
-    pullToolkit([...array], even);
+    pullToolkit([...array], ...even);
```
ssi02014, line 13: There was a type error, which I fixed.

PR #1244, opened 2025-07-01
fix(flattenObject): improve flattenObject performance

JeongHwan-dev:
## Summary

Optimizes `flattenObject` performance by replacing `Object.keys()` iteration with direct `for-in` loop.

## Benefits

This change delivers significant performance improvements by eliminating unnecessary array creation through `Object.keys()`. The new `for-in` approach directly iterates over object properties, reducing memory allocation and improving execution speed, with even greater gains for large objects. The V8 engine can better optimize `for-in` loops, resulting in more efficient code execution. The implementation also becomes cleaner and more readable while maintaining identical API behavior with zero breaking changes. The added `Object.hasOwn()` check ensures only own properties are processed, providing better security and predictable behavior.

## Changes

- Replace `Object.keys()` + for loop with `for-in` loop
- Add `Object.hasOwn()` check for proper property filtering

## Test
test

### DIFF
--- a/src/object/flattenObject.ts
+++ b/src/object/flattenObject.ts
@@ -40,25 +40,24 @@ export function flattenObject(object: object, { delimiter = '.' }: FlattenObject
 
 function flattenObjectImpl(object: object, prefix = '', delimiter = '.'): Record<string, any> {
   const result: Record<string, any> = {};
-  const keys = Object.keys(object);
 
-  for (let i = 0; i < keys.length; i++) {
-    const key = keys[i];
-    const value = (object as any)[key];
+  for (const key in object) {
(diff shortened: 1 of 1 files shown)


raon0211:
Can you run a benchmark of two different implementations? You can see the sample benchmark code in `benchmarks` directory.

JeongHwan-dev:
@raon0211 
I ran benchmarks using the sample code in the benchmarks directory for both implementations. The results showed no significant performance difference between the two, and in some cases, the updated implementation actually performed worse. 
I think it makes sense to close this PR now. 

Thanks again for the feedback! 🙇🏻‍♂️

PR #1249, opened 2025-07-02
test(debounce): change delay to advanceTimerByTime in test cases

ssi02014:
Removed the delay function used in the test code and changed it to utilize the advanceTimersByTime function provided by vitest.

We could also change “compat/debounce” to advanceTimerByTime, but the test case below causes infinite calls in a fake timer environment, limiting the use of advanceTimerByTime.

Therefore, I think it would be better to keep the existing test code for “compat/debounce.”

```ts
it('should invoke the trailing call with the correct arguments and `this` binding', async () => {
    let actual: any;
    let callCount = 0;
    const object = {};

    const debounced = debounce(
      // eslint-disable-next-line @typescript-eslint/no-unused-vars
      function (this: any, _: any) {
        actual = [this];
        // eslint-disable-next-line prefer-rest-params
        Array.prototype.push.apply(actual, arguments as any);
        return ++callCount !== 2;
      },
      32,
      { leading: true }
    );
   
    // 😭 Oops.. Currently, the test code cannot escape the while loop in the fake timer environment.
    while (true) {
      if (!debounced.call(object, 'a')) {
        break;
      }
    }

    await delay(32);

    expect(callCount).toBe(2);
    expect(actual).toEqual([object, 'a']);
  });
```

### DIFF
--- a/src/function/debounce.spec.ts
+++ b/src/function/debounce.spec.ts
@@ -1,10 +1,16 @@
-import { describe, expect, it, vi } from 'vitest';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 import { debounce } from './debounce';
-// adjust the import path as necessary
-import { delay } from '../promise';
+
+beforeEach(() => {
+  vi.useFakeTimers({ shouldAdvanceTime: true });
+});
+
+afterEach(() => {
+  vi.useRealTimers();
+});
 
 describe('debounce', () => {
-  it('should debounce function calls', async () => {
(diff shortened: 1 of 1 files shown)


dayongkr:
Thanks for the contribution!

But, we're trying to avoid using Fake Timer.

As you probably saw while editing, we only use delays in milliseconds, so using timer mocks doesn't really give us much benefit.

ssi02014:
@dayongkr Yes I understand your intent. Thanks for the feedback

PR #1263, opened 2025-07-08
test(debounce): improved the test code for the debounce function

ssi02014:
Improved the test code for the debounce function.

- Add test code for the `debounced.flush` function.
- Added test code for the `edges option of the debounce` function.


src/function/debounce.spec.ts
```diff
@@ -13,6 +13,8 @@ describe('debounce', () => {
     debouncedFunc();
     debouncedFunc();
 
+    expect(func).not.toHaveBeenCalled();
```
ssi02014: This code was added to more clearly test the basic debounce behavior.

src/function/debounce.spec.ts
```diff
@@ -3,110 +3,116 @@ import { debounce } from './debounce';
 // adjust the import path as necessary
 import { delay } from '../promise';
 
+const DEBOUNCE_MS = 50;
```
ssi02014: To reduce code duplication, change debounceMs to a constant.

src/function/debounce.spec.ts
```diff
@@ -3,110 +3,116 @@ import { debounce } from './debounce';
...
 
+  it('should immediately invoke the delayed function when flush is called', async () => {
+    const func = vi.fn();
+    const debouncedFunc = debounce(func, DEBOUNCE_MS);
+
+    debouncedFunc();
+    debouncedFunc.flush();
+
+    expect(func).toHaveBeenCalledTimes(1);
+  });
```
ssi02014: Add test code for the `debounced.flush` function.

src/function/debounce.ts
```diff
@@ -50,6 +50,7 @@ export interface DebouncedFunction<F extends (...args: any[]) => void> {
  * @param {number} debounceMs - The number of milliseconds to delay.
  * @param {DebounceOptions} options - The options object
  * @param {AbortSignal} options.signal - An optional AbortSignal to cancel the debounced function.
+ * @param {Array<'leading' | 'trailing'>} options.edges - An optional array specifying whether the function should be invoked on the leading edge, trailing edge, or both.
```
ssi02014: Added missing comments.

src/function/debounce.spec.ts
```diff
@@ -3,110 +3,161 @@ import { debounce } from './debounce';
...
+    debouncedFunc();
+
+    expect(func).toHaveBeenCalledTimes(1);
+
+    debouncedFunc();
+
+    await delay(DEBOUNCE_MS);
+
+    expect(func).toHaveBeenCalledTimes(2);
+  });
```
ssi02014: Added test code for the edges option of the debounce function.

ssi02014:
It passes when tested locally. I don't know why `ci/circleci: test` fails. 😭

src/function/debounce.ts
eranshcloud: there's also an export missing for the DebounceOption type, I have a PR waiting for approval, but nobody reviewed it.

ssi02014:
```ts
type DebounceOptions = Parameters<typeof debounce<() => void>>[2];
```
@codecov-commenter  It's a shame that the review is delayed 🥲
DebounceOptions can be approached in the same way as above

eranshcloud: yes, but this is a ridiculous way to use it... hopefully we get proper responses on our PRs.

ssi02014:
@dayongkr @raon0211 When will the review for this PR take place?

dayongkr (dismissed):
Thanks for improving the debounce tests! I pushed a small follow-up: kept the `flush` and leading-edge cases (they cover the previously-missing branches and bring `debounce.ts` to 100%), and dropped the redundant trailing/both cases plus the shared timeout constant so each test stays self-contained. Approving — will merge once CI is green.

dayongkr (approved):

ssi02014:
@dayongkr Thank you. It took quite a while, but I’m pleased that it’s been approved and merged 😊

PR #1265, opened 2025-07-11
fix(docs): correct typo in findKey.md

fullmetalsheep:
Corrected wrong variable reference in the given example


raon0211 (approved):
Thanks so much!

PR #1284, opened 2025-07-31
fix: simplified check for primitive

Connormiha:
I simplified the primitive check using Object, which always returns the argument itself if it's not a primitive.
But the main point is that this approach even works with `document.all`

### DIFF
--- a/src/predicate/isPrimitive.ts
+++ b/src/predicate/isPrimitive.ts
@@ -27,5 +27,5 @@
  * isPrimitive([1, 2, 3]); // false
  */
 export function isPrimitive(value: unknown): value is null | undefined | string | number | boolean | symbol | bigint {
-  return value == null || (typeof value !== 'object' && typeof value !== 'function');
+  return Object(value) !== value;
 }


raon0211:
Hey, thanks for the PR! 

I appreciate you suggesting this alternative approach. You're right that `Object(value) !== value` is a clever way to check for primitives. However, I ran some benchmarks on my machine and found that the current implementation actually performs better:

image

### Benchmark code
```typescript

function isPrimitive(value: unknown): value is null | undefined | string | number | boolean | symbol | bigint {
  return value == null || (typeof value !== 'object' && typeof value !== 'function');
}

function isPrimitiveAlternative(
  value: unknown
): value is null | undefined | string | number | boolean | symbol | bigint {
  return Object(value) !== value;
}

describe('isPrimitive', () => {
  bench('es-toolkit/isPrimitive', () => {
    isPrimitive({});
    isPrimitive([]);
    isPrimitive(null);
    isPrimitive(1);
    isPrimitive(true);
    isPrimitive(Symbol('a'));
    isPrimitive(123n);
  });

  bench('es-toolkit/isPrimitive alternative', () => {
    isPrimitiveAlternative({});
    isPrimitiveAlternative([]);
    isPrimitiveAlternative(null);
    isPrimitiveAlternative(1);
    isPrimitiveAlternative(true);
    isPrimitiveAlternative(Symbol('a'));
    isPrimitiveAlternative(123n);
  });
});
```

Beyond performance, I also think the current implementation is more readable - it's immediately clear what's being checked, whereas `Object(value) !== value` requires a bit more mental parsing to understand.

Given these factors, I think we should keep the current implementation. Really appreciate you taking the time to explore alternatives though - these kinds of discussions help us make better decisions about the codebase. Looking forward to your future contributions! 🙂

PR #1287, opened 2025-07-31
fix(cloneDeepWith): Handle customizer returning null

### LINKED ISSUE #1286: Returning `null` from `cloneDeepWith` customizer
In Lodash, and according to the es-toolkit docs, if `cloneDeepWith`'s customizer returns `undefined`, it uses the default clone method; it otherwise uses the returned value. Our codebase relies on this behavior to skip cloning certain items by returning `null`. However, this doesn't work in es-toolkit; it treats null and undefined as equivalent.

joshkel:
Fixes #1286


raon0211 (approved):
Thanks for fixing this!

PR #1290, opened 2025-08-03
chore: add Vitest ESLint plugin configuration

manudeli:


eslint.config.mjs
```diff
@@ -38,6 +39,18 @@ export default [
...
+  {
+    files: ['**/*.spec.ts*', '**/*.test.ts*', '**/*.test-d.ts*'],
+    plugins: { vitest },
+    rules: {
+      ...vitest.configs.recommended.rules,
+      'vitest/no-commented-out-tests': 'warn',
+      'vitest/expect-expect': 'warn',
+      'vitest/valid-expect': 'warn',
+      'vitest/no-identical-title': 'warn',
+      'vitest/valid-title': 'warn',
```
manudeli, line 51: I think it would be good to gradually change these rules from 'warn' to 'error' after merge this pr

lumirlumir: Hi, @manudeli 

I came across this from ESLint 😄

We now support Bulk Suppression starting from v9.24.0. Feel free to check it out—I think it could be helpful for your situation, especially for a smoother transition to `"error"` from `'warn'` or `'off'`.

Here are some references:

- https://eslint.org/docs/latest/use/suppressions
- https://eslint.org/blog/2025/04/eslint-v9.24.0-released/#bulk-suppressions

eslint.config.mjs
```diff
@@ -38,6 +39,18 @@ export default [
   },
   pluginJs.configs.recommended,
   ...tseslint.configs.recommended,
+  {
+    files: ['**/*.spec.ts*', '**/*.test.ts*', '**/*.test-d.ts*'],
```
dayongkr:
```suggestion
    files: ['**/*.spec.ts*'],
```

dayongkr (approved):
Looks good! Thank you! 

However, I'm planning to standardize test code to use `spec.ts` format, so I've removed other formats and only included that format in the `files`.

eslint.config.mjs
manudeli: I didn't know that way it. Thanks for sharing good information 👍

PR #1297, opened 2025-08-04
feat(benchmarks): update performance documentation

manudeli:


docs/performance.md
```diff
@@ -4,23 +4,19 @@ description: The performance difference between es-toolkit and alternative libra
 
 # Performance
 
-![Graph showing the difference in performance between es-toolkit and lodash. Using es-toolkit results in a performance gain of up to 11 times.](/assets/performance.png)
-
-Designed with performance in mind, es-toolkit provides an average of 2× performance improvement compared to alternative libraries like lodash. Some functions achieve up to an 11× performance gain by fully utilizing modern JavaScript APIs.
```
manudeli, line 9: Currently, the difference is not up to 11 times, so this part can be considered an exaggeration, so it is removed.

benchmarks/performance/difference.bench.ts
```diff
@@ -2,10 +2,12 @@ import { bench, describe } from 'vitest';
 import { difference as differenceToolkit_ } from 'es-toolkit';
 import { difference as differenceCompatToolkit_ } from 'es-toolkit/compat';
 import { difference as differenceLodash_ } from 'lodash';
+import { difference as differenceLodash_es } from 'lodash-es';
```
dayongkr, line 5: As far as I know, there shouldn't be any performance differences between lodash-es and lodash, so I'm wondering what the reason was for adding it?

manudeli: image

The page https://es-toolkit.dev/performance.html uses lodash-es to demonstrate the performance of es-toolkit. so I used lodash-es too

dayongkr:
Thanks for this, and sorry for the very late revisit. We're moving the performance docs toward dynamic generation (#1578), which builds the tables and chart from `benchmark-results.json` instead of hand-maintained tables, so this approach supersedes the manual updates here. I'll close this one for that reason. I really appreciate the work, especially dropping the exaggerated 11x claim. I'll make sure that correction carries over into the dynamic version.

PR #1301, opened 2025-08-08
fix(cloneDeepWith): fix runtime error in environments without Blob support

Li13:
I'm unable to use cloneDeepWith in my mini-program. It fails in runtime environments that do not support Blob. Although src/predicate/isBlob.ts provides compatibility for environments without Blob support, the cloneDeepWith function itself does not have this compatibility.


dayongkr (approved):
Thanks for your contribution!

PR #1302, opened 2025-08-08
fix(sample): include "undefined" in return union type

screendriver:
`sample()` returns `undefined` when an empty array is given. This should also be reflected in the return type.


dayongkr (approved):
Looks good to me! Thanks for your contribution!

screendriver:
Unfortunately I cannot merge because the `deno` CI job is pending and required.

dayongkr:
Since this involves a change to the interface, I haven’t merged it yet because I’d like to hear other maintainer’ opinion. I’ll leave another comment later this week.

screendriver:
> I’ll leave another comment later this week

Since this is almost 3 months ago I just wanted to ask kindly if we can merge it @dayongkr ?

dayongkr:
Ah, this was put on hold per previous discussion with maintainers. Let me sync up and update.

dayongkr:
After thinking about it more, I think this needs to be addressed in the next major update, so I'll discuss internally about the timing of the next major release.

Sorry for making you wait so long.

dayongkr:
It is technically a breaking change, but since only the types are affected with no runtime code changes, and the current types are incorrect anyway, I'll go ahead and apply the fix.

PR #1308, opened 2025-08-11
fix(tests): Fix test in US time zones

joshkel:
`new Date('2024-01-01')` evaluates to 2024-01-01T00:00:00.000Z, then `getFullYear` resolves that in the local time zone. In time zones that are negative of UTC (such as America/New_York), this results in `getFullYear` giving the previous year, causing a spurious test failure.

Moving the test dates away from January 1 and December 31 fixes this.


terrymun:
@joshkel Have you considered [manually setting a timezone](https://github.com/vitest-dev/vitest/issues/1575#issuecomment-1439286286) in vitest? This approach might be a little more bulletproof in the future, if there are other test files that end up needing to parse dates.

I'm happy to help with this by creating a PR to verify this alternative approach :)

joshkel:
@terrymun I thought about that, but this seemed simpler.

I'm happy to go with whichever approach the developers prefer.

dayongkr (approved):
Thanks for your contribution!

PR #1317, opened 2025-08-13
fix: remove trailing spaces in sortBy function example

wo-o29:
## Summary
This PR updates the `sortBy` function example data in all documentation languages (`en`, `ko`, `zh_hans`, `zh_hant`) to remove unintended trailing spaces in the `user` property.  

The extra spaces caused the actual sort result to differ from the expected example output, which could confuse readers.

## Changes
- Removed trailing spaces from `'bar '` → `'bar'`
- Removed trailing spaces from `'foo '` → `'foo'`
- Applied this change consistently across all translations to ensure identical execution results.


dayongkr (approved):
Thanks!!

PR #1318, opened 2025-08-13
fix: some function example issues in Korean docs and variable name mismatches in all languages

wo-o29:
## Summary
This PR addresses issues in the documentation examples for the `some` function, making them consistent and executable as intended.

## Changes
1. Korean documentation (`ko`):
   - Replaced incorrect `every` calls with `some` in both array and object examples.
   - Updated expected outputs in the object examples to match the actual behavior of `some`.

2. All language documentation (`en`, `ko`, `zh_hans`, `zh_hant`):
   - Fixed variable name mismatches in examples where `obj` was declared but `object` was used in function calls.


dayongkr (approved):
Thanks for your contribution!

PR #1321, opened 2025-08-15
fix: incorrect findLast examples in English and Japanese docs

wo-o29:
## Summary
This PR fixes incorrect examples in the English and Japanese documentation for the `findLast` function.  
The current examples incorrectly import and call `find` instead of `findLast`.

## Changes
- Replaced all `find` imports and calls with `findLast` in English and Japanese `findLast` documentation examples.
- Updated array and object usage examples to correctly demonstrate `findLast` behavior.


dayongkr (approved):

PR #1324, opened 2025-08-15
fix: Fix incorrect lowerFirst examples

wo-o29:
## Summary
This PR fixes incorrect examples in the Japanese documentation for the `lowerFirst` function.  
The current examples mistakenly use the `lowerCase` function instead of `lowerFirst`.

## Changes
- Replaced all `lowerCase` calls with `lowerFirst` in the Japanese `lowerFirst` documentation examples.
- Ensured that the examples now accurately demonstrate how only the first character of a string is lowercased.


dayongkr (approved):

PR #1325, opened 2025-08-15
fix(defaultsDeep): defaultsDeep can now deep merge objects in from arrays

kilimandzharov:
### Greetings!

First of all I wanted to say thank you for such a miraculous project, my colleagues and me really appreciate your work!

During migration from lodash I noticed that `es-toolkit/defaultsDeep` works slightly different from `lodash/defaultsDeep`. To be exactly, objects in arrays are treated in different ways: `es-toolkit/defaultsDeep` just  copies array from the field and `lodash/defaultsDeep` merges objects deeply. Moreover, `lodash/defaultsDeep` append additional elements from the source array to the target array if the source array is longer than the target array. And that's why I suggest you these fixes.

I also attached benchmark screenshot just like your contribution guide says. Thank you for your time!

Снимок экрана 2025-08-15 в 19 27 40

### DIFF
--- a/src/compat/object/defaultsDeep.spec.ts
+++ b/src/compat/object/defaultsDeep.spec.ts
@@ -224,6 +224,24 @@ describe('defaultsDeep', () => {
     });
   });
 
+  it('should work with objects in arrays', () => {
+    const target = { a: [{ foo: 1 }] };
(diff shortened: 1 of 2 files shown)


dayongkr (approved):
Thank you for reporting this issue and for submitting the PR!

PR #1329, opened 2025-08-17
fix: Fix incorrect callback key parameter type in each docs

wo-o29:
## Summary
This PR fixes a documentation error in the `each` function for the English and Korean docs.  
The `callback` parameter type for the key was incorrectly documented as `T` instead of `keyof T`.

## Changes
- Updated the `callback` parameter type in English and Korean docs:
  - Incorrect: `(value: T[keyof T], key: T, object: T)`
  - Corrected: `(value: T[keyof T], key: keyof T, object: T)`
- Adjusted the parameter description so that:
  - `key` is described as the **current element's key (index or property name)**, matching the actual function signature.

This change aligns the documentation with the correct TypeScript definition and clarifies the role of the `key` parameter.

Additionally, updated the callback parameter names in the examples to match the interface.(`ko`, `en`, `ja`)


dayongkr (approved):
Thanks for your contribution!

PR #1330, opened 2025-08-17
fix: Fix incorrect return type description for isFunction docs

wo-o29:
## Summary
This PR fixes incorrect return type descriptions in the Korean and Japanese documentation for the `isFunction` function.  
Both docs incorrectly stated the return type as `value is number`.

## Changes
- Updated the return type in Korean and Japanese docs:
  - Incorrect: (value is number)
  - Corrected: (value is (...args: never[]) => unknown)
- Ensured the description now matches the actual TypeScript interface and type guard behavior.


dayongkr (approved):
Thanks for your contribution!

PR #1331, opened 2025-08-17
fix: Fix incorrect return type of isInteger docs

wo-o29:
## Summary
This PR corrects the return type description of the `isInteger` function in all documentation translations (`en`, `ko`, `zh_hans`, `zh_hant`).

Previously, the docs stated the return type as `boolean`.  
In TypeScript, `isInteger` is a type guard, so the proper return type is `value is number`.

## Changes
- Updated the return type description in all languages:
  - Incorrect: (`boolean`)
  - Correct: (`value is number`)
- Ensured that the narrative explanation still clarifies: returns `true` if the value is an integer, otherwise `false`.

### DIFF
--- a/docs/ja/reference/compat/predicate/isInteger.md
+++ b/docs/ja/reference/compat/predicate/isInteger.md
@@ -22,7 +22,7 @@ function isInteger(value?: unknown): value is number;
 
 ### 戻り値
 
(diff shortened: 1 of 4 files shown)


dayongkr (approved):
Thanks for your contribution!

PR #1332, opened 2025-08-17
fix: Add missing parameter and return descriptions to isNil docs

wo-o29:
## Summary
This PR updates the `isNil` documentation across all translations to include missing parameter and return descriptions.  
Previously, the docs only contained the interface and examples, but did not explicitly describe the function's parameter or return value.

## Changes
- Added a "Parameters" section explaining:
  - `x` (`unknown`): The value to check.
- Added a "Returns" section explaining:
  - `(x is null | undefined)`: Returns `true` if the value is `null` or `undefined`, otherwise `false`.
- Applied the same update consistently across all language versions (`en`, `ko`, `zh_hans`, `zh_hant`).

This ensures the `isNil` documentation is complete, consistent, and aligns with other predicate function docs.

### DIFF
--- a/docs/ja/reference/predicate/isNil.md
+++ b/docs/ja/reference/predicate/isNil.md
@@ -12,6 +12,14 @@ TypeScriptの型ガードとしてよく使用され、パラメータとして�
 function isNil(x: unknown): x is null | undefined;
 ```
 
+### パラメータ
+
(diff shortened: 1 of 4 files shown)


dayongkr (approved):
Thanks for your contribution!

PR #1334, opened 2025-08-17
fix(startsWith): Fix startsWith docs return type from string to boolean

wo-o29:
## Summary
This PR fixes the incorrect return type in the documentation for the `startsWith` function.  
The original docs mistakenly specify the return type as `string` while the actual function returns a `boolean`.

## Changes
- Updated the function signature to:
```ts
function startsWith(str: string, target: string, position: number = 0): boolean;
```
- This update was applied to Korean, English, and Simplified Chinese (`zh_hans`) documentation only. The Japanese documentation was already correct.

### DIFF
--- a/docs/ko/reference/compat/string/startsWith.md
+++ b/docs/ko/reference/compat/string/startsWith.md
@@ -11,7 +11,7 @@
 ## 인터페이스
 
 ```typescript
-function startsWith(str: string, target: string, position: number = 0): string;
+function startsWith(str: string, target: string, position: number = 0): boolean;
 ```
 
 ### 파라미터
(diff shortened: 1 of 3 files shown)


dayongkr:
I really appreciate your contributions, but since there are computing resources involved for CI and dev version releases, if you're contributing similar issues at the same time, documentation-related fixes can be batched together without any issues for review, so it would be great if you could submit them all at once! 

Thank you!!

wo-o29:
> I really appreciate your contributions, but since there are computing resources involved for CI and dev version releases, if you're contributing similar issues at the same time, documentation-related fixes can be batched together, so it would be great if you could submit them all at once!
> 
> Thank you!!

I confirmed it! Thank you for your reply👍👍👍

PR #1335, opened 2025-08-17
fix(split): Update split docs return type to string[] for consistency with interface

wo-o29:
## Summary
This PR updates the return type description in the documentation for the `split` function.  
Previously, the return type was described as `Array`, which is less specific and inconsistent with the interface signature.

## Changes
- Changed the return type description from `Array` to `string[]` in all language documents.
- This clearly indicates that the function returns an array of strings.
- Improved the consistency between the documentation and the TypeScript interface signatures.


dayongkr (approved):
Thanks for your contribution!! Super Thanks!!

PR #1336, opened 2025-08-17
fix: Update stubFalse and stubTrue docs to use literal return types false and true

wo-o29:
## Summary
This PR updates the documentation for the `stubFalse` and `stubTrue` functions to reflect their precise return types.  

Both functions always return a constant literal value (`false` and `true` respectively), so the return types are changed from the broad `boolean` to the literal types `false` and `true`.  

## Changes
- Changed `stubFalse` return type from `boolean` to `false`.
- Changed `stubTrue` return type from `boolean` to `true`.

### DIFF
--- a/docs/ja/reference/compat/util/stubFalse.md
+++ b/docs/ja/reference/compat/util/stubFalse.md
@@ -11,12 +11,12 @@
 ## インターフェース
 
 ```typescript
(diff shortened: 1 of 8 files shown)


raon0211 (approved):
Thanks for the update!

PR #1337, opened 2025-08-17
fix: Add return type descriptions to padEnd, padStart, and repeat docs

wo-o29:
## Summary
This PR adds return type descriptions to the documentation for the `padEnd`, `padStart`, and `repeat` functions.

Previously, the docs explained the functionality but omitted explicit return type descriptions.

## Changes
- Added `(string)` return type descriptions indicating these functions return a string.
- Ensured consistent documentation format across all three functions.
- Applied changes to all language versions.


raon0211 (approved):
Thanks for the update!

PR #1340, opened 2025-08-17
docs: Update isString docs return type from `x is string` to `value is string`

wo-o29:
## Summary
This PR updates the return type description of the `isString` function in the documentation to match the TypeScript interface exactly.

## Interface

```typescript
function isString(value: unknown): value is string;
```

- Changed from:
(`x is string`): Returns true if the value is a string, false otherwise.

- To:
(`value is string`): Returns true if the value is a string, false otherwise.

This change improves consistency between the interface and documentation, reducing potential confusion.

### DIFF
--- a/docs/ja/reference/predicate/isString.md
+++ b/docs/ja/reference/predicate/isString.md
@@ -16,7 +16,7 @@ function isString(value: unknown): value is string;
 
 ### 戻り値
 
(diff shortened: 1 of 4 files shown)


raon0211 (approved):
Thanks for the update!

PR #1343, opened 2025-08-17
docs: Add error handling documentation for zero step value in range function

wo-o29:
## Summary
This PR enhances the documentation for the `range` function by adding an error section to clarify behavior when `step` is 0.

- Previously, the documentation did not specify any error for a `step` value of 0.
- The documentation has been updated with a new **Error** section stating that:
  - Providing a `step` value of 0 will result in an error being thrown.

This update improves the clarity and reliability of the function's usage guidance.


raon0211 (approved):
Thanks for the update!

PR #1350, opened 2025-08-19
fix: Fixing errors when undefined is passed

DongHunWon:
An error occurs when `arr` is undefined and passed to `arr.slice()`.  
This fix adds a guard clause to return an empty array when `arr` is undefined.


raon0211:
Hello! Thanks for opening a pull request.

That said, the function signature specifies that remove should only receive valid arrays. Since we prioritize keeping the implementation simple (see our [contribution guide](https://github.com/toss/es-toolkit/blob/main/.github/CONTRIBUTING.md#13-simplicity)), we think handling `undefined` while bypassing TypeScript is more of an edge case. For this reason, we don’t think it makes sense to merge this PR.

We really appreciate your effort, though, and look forward to your next contribution!

PR #1352, opened 2025-08-20
test: Reduce CI time by avoiding duplicate tarball creation in check-dist tests

sen2y:
# optimize: Reduce CI time by avoiding duplicate tarball creation in check-dist tests

## Problem
The `check-dist.spec.ts` test was creating tarballs **twice** - once in each test case, causing unnecessary duplication and increased CI execution time.

## Solution
Move tarball creation to `beforeAll()` hook to create it only once and reuse across both tests.

### Local Testing
- **Before**: 85.30 seconds
- **After**: 58.62 seconds  
- **Improvement**: **31.3% faster** (26.68 seconds saved)

## Code Changes

**Before:**
```typescript
// Each test creates its own tarball ❌
it('configures all entrypoints correctly', async () => {
  const packageJson = await getPackageJsonOfTarball(); // Creates tarball
});

it('exports identical functions in CJS and ESM', async () => {
  const tarball = await createPackageTarball(); // Creates tarball again
});
```

**After:**
```typescript
// Shared tarball creation ✅
describe("es-toolkit's package tarball", () => {
  let tarball: { path: string };

  beforeAll(async () => {
    tarball = await createPackageTarball(); // Single creation
  }, 300000);

  it('configures all entrypoints correctly', async () => {
    const packageJson = await getPackageJsonOfTarball(tarball.path); // Reuse
  });

  it('exports identical functions in CJS and ESM', async () => {
    // Reuse existing tarball
    const packageJson = { dependencies: { 'es-toolkit': tarball.path } };
  });
});
```

## Benchmark Results
image
Performance improvement verified through mock-based benchmarks:

```bash
yarn bench benchmarks/performance/check-dist-mock.bench.ts
```

**Results:**
- **Optimized approach**: 1.70x faster than original
- **Function call reuse**: 1.01x faster than recreation

*Note: Mock-based benchmarks used due to expensive I/O operations in actual tarball generation*

***Background**: During CI testing, `check-dist.spec.ts` occasionally exceeded the 120-second timeout limit. While the timeout has been recently adjusted by maintainers 8e5b6dc5bd89db5cdab6522d22c2b324489658dd, I noticed an opportunity to improve the underlying performance by eliminating duplicate tarball creation in the test suite.*

  ***Feedback and suggestions are always welcome!***

### DIFF
--- a/benchmarks/performance/check-dist-mock.bench.ts
+++ b/benchmarks/performance/check-dist-mock.bench.ts
@@ -1,0 +1,56 @@
+import { bench, describe } from 'vitest';
+
+const mockCreatePackageTarball = () => {
+  return new Promise(resolve => {
+    setTimeout(() => {
+      resolve({ path: '/tmp/mock-package.tgz' });
+    }, 50);
+  });
+};
+
+const mockProcessTarball = () => {
+  return new Promise(resolve => {
+    setTimeout(() => {
+      resolve({ exports: { '.': {}, './array': {}, './object': {} } });
+    }, 10);
+  });
+};
+
+describe('check-dist performance comparison (mocked)', () => {
+  bench('Original approach: Duplicate tarball creation', async () => {
+    await mockCreatePackageTarball();
+    await mockProcessTarball();
+
+    await mockCreatePackageTarball();
+    await mockProcessTarball();
+  });
+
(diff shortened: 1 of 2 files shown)


raon0211 (approved):
Thanks for the optimization!

PR #1358, opened 2025-08-22
perf(uniq): improve performance

Dohun-choi:
### Test
image

### Benchmarks
image


Dohun-choi:
image

This is the benchmark for the modified code, which failed tests related to functions using uniq.
For the same reason, I have also added test cases for uniq (sparse arrays and special values such as NaN, -0, and Infinity).

raon0211 (approved):
Thanks!

PR #1359, opened 2025-08-23
test: Improve debounce/throttle test performance by replacing real timers with fake timers

miinhho:
## Summary

Replaced real timer usage with fake timers in debounce and throttle tests to eliminate flaky test risks and significantly improve test execution time.

## Problem

- Tests were using real timers with `delay()` function, making them slow and potentially flaky
- Real timer-based tests are susceptible to timing issues in CI environments
- Total test execution time was unnecessarily long due to actual time delays

## Solution
- **Targeted approach**: Only replaced real timers with fake timers for debounce and throttle tests that were the primary cause of slow test execution
- **Simple replacements**: For straightforward cases, replaced `delay()` calls with `vi.advanceTimersByTime()` for precise time control
- **Preserved complex tests**: When tests required additional modifications beyond simple `delay()` replacement, maintained original tests using `vi.useRealTimers()` to avoid breaking existing functionality

### Individual Test Files
Test File | Before | After | Improvement
-- | -- | -- | --
src/compat/function/debounce.spec.ts | 4.08s | 1.16s | 3.52x faster
src/compat/function/throttle.spec.ts | 5.08s | 2.02s | 2.52x faster
src/function/debounce.spec.ts | 903ms | 12ms | 75.25x faster
src/function/throttle.spec.ts | 1.45s | 15ms | 96.67x faster

### Overall Test Suite Impact
Test execution time: 19.64s → 11.33s (42.3% improvement)

## Benefits
- Eliminated flaky test risks - No dependency on system timing
- Faster CI/CD pipeline - Reduced test execution time by 42%
- More reliable tests - Deterministic timing behavior


dayongkr:
Thanks for the work and the measurements, the speedup is real. That said, we'd like to keep real timers here: we value the real event-loop/macrotask fidelity for these time-based utilities, and the suite is already fast enough that they're not a CI bottleneck (no flakiness in practice). To be clear, fake timers aren't less correct, it's a fidelity-vs-speed trade-off and we're choosing fidelity. Closing for now, but happy to revisit if flakiness shows up or timing becomes a real bottleneck. If I'm missing something, let me know.

PR #1360, opened 2025-08-23
test(compat/defer): Replace deprecated done() callback with fake timers in async tests

miinhho:
## Problem
Deprecation warning occurs when using the `done()` callback in Vitest tests

## Solution
Replaced Vitest's deprecated `done()` callback with fake timers

### Before
Screenshot 2025-08-24 011347
Screenshot 2025-08-24 011334

### After
image


dayongkr (dismissed):
Thanks! I pushed a small follow-up: kept real timers (async/await + the `delay` helper) instead of fake timers, to stay consistent with #1359. The deprecated `done()` removal is the valuable part here. Tracking in #1777. Will merge once CI is green.

dayongkr (dismissed):

dayongkr (approved):

Issue #1355, opened 2025-08-21
toCamelCaseKeys has wrong type for PascalCase


lauritzsh:
I created a simple TypeScript playground: https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAbzjCBhAhiApgGwwZywGksBPfOAXzgDMoIQ4AiLfAWhQhwGtgYmA3ACghAYwgA7fPDDp8o9DgBKrSFKwAuRHBU0sULBNFYAkgBMt0qMAkBzKnAC823fsPHzWpgAtcOCExUwmKS0nD4EujcWCr4aoRaSAZ6BkZYAPrAFuEw1nYOzklYKe4ZWV6+OP6BlMHiUjJyCjhOyGiYuATEZPgAFLLyirHxWACUwgD0E3AzcAB6APwhUlxYAHT+tv1NiuMi9WERUVitKBjYeHLd5L1H0cOhY5PTs4vL+KsbEFt3T0JAA

<img width="566" height="248" alt="Image" src="https://github.com/user-attachments/assets/6f65bfb4-1742-4be8-b365-0ed3287df08c" />

For the PascalCase it seems that `ReferenceId` ends up as `referenceid` with everything being lower case. For snake_case it behaves correctly. It's strictly a type issue, if I run it I do see

```
bun test.ts
{
  referenceId: "hello",
}
{
  referenceId: "hello",
}
```

I am on version 1.39.10 of es-toolkit.

lauritzsh:
One solution is to use the `CamelCase` from `type-safe` library

[...]

PR #1362, opened 2025-08-24
fix(toCamelCaseKeys): Add a type for PascalCase

Jeongeun-Choi:
## Problem
When object keys were written in **PascalCase**, the inferred type was not properly converted to the expected `camelCase` form.  
This caused inconsistent typing results compared to `snake_case`.

## Solution
- Added explicit handling for `PascalCase` keys in the type transformation.
- Now both `PascalCase` and `snake_case` keys are correctly converted to `camelCase`.

## Before Result
```ts
const pascal: { referenceid: string; };
const snake: { referenceId: string };
```
## After Result
```ts
const pascal: { referenceId: string; };
const snake: { referenceId: string };
```
## Related Issue
Fixes [#1355](https://github.com/toss/es-toolkit/issues/1355)

## Notes
This change ensures consistent type conversion rules regardless of whether the input key is PascalCase or snake_case.

No breaking changes expected.

### DIFF
--- a/src/object/toCamelCaseKeys.spec.ts
+++ b/src/object/toCamelCaseKeys.spec.ts
@@ -8,6 +8,12 @@ describe('camelizeKeys', () => {
     expect(toCamelCaseKeys(input)).toEqual(expected);
   });
 
+  it('should convert PascalCase keys to camelCase in a flat object', () => {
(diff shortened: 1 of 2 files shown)


raon0211 (approved):
Thanks!

PR #1374, opened 2025-08-28
fix(some): update predicate check to handle null and undefined cases

D-Sketon:
current behavior: 
{254F5C39-E25C-4B33-92E8-5EA3C10D5E5A}

{74A51D4E-7768-4BA6-A7E7-31D7A0CDE091}

This PR prevents "" and 0 from being converted to `identity`.


raon0211 (approved):
Thanks!

PR #1395, opened 2025-09-08
fix: imports in bundlers

aidant:
The current `browser` bundle is misleading. It points to the `es-toolkit/compat` and does not contain any exports. This means if you `import { flatten } from "es-toolkit"` in a bundler which prioritises the `browser` field `flatten` will be `undefined`. Alternatively if you intended to `import "es-toolkit"` and use a method not exposed by the `es-toolkit/compat` layer such as `flattenObject` the function will be `undefined`. Since its common for bundlers to wrap the modules the `_` global may also not be exposed.

The simplest fix is to remove the `browser` bundle and let bundlers take over, however, if there is a desire to expose a script which provides the `es-toolkit/compat` layer as a global I suggest we explore moving it to a new subpath export.


dayongkr:
Thanks for surfacing this, @aidant — the diagnosis is correct, and you're right that pointing the `browser` field at the UMD bundle is not best practice.

That said, simply removing the UMD bundle isn't safe either: jsdelivr/unpkg and similar CDNs rely on a prebuilt global bundle for `<script>` usage, so dropping it outright can break browser loading (e.g. the documented `_` global stops working, or the bare URL falls back to a non-browser entry). So there are a few more things to consider here than just removing the field.

Our direction: keep a **UMD bundle for CDN `<script>` usage** (it's the standard for global `<script>` distribution), but stop overloading `browser` — move it to dedicated `jsdelivr`/`unpkg` fields so bundlers resolve ESM while CDNs still get the global bundle. Full analysis and the proposed plan are in #1779 — please take a look.

We'll keep this PR open and merge it once we've finalized those details. 🙏

aidant:
@dayongkr I've updated the PR to match your proposal.

It's worth noting that the browser field is now compatible with our build system. The two changes that resulted in us dropping our patch were the change to a UMD module from the previous IIFE and the change to export the whole `es-toolkit` rather than the previous `es-toolkit/compat` behaviour.

dayongkr:
Thanks for the contribution, and for narrowing it down so cleanly! Verified locally: the UMD bundle defines the global `_` and `jsdelivr`/`unpkg` now resolve to it, fixing the broken unpkg `<script>`. This resolves #1779.

dayongkr (approved):

dayongkr:
Follow-up: this shipped in v1.48.0, but the `jsdelivr`/`unpkg` fields turned out to be stripped at publish time. yarn berry's `publishConfig` only carries standard fields (`main`/`module`/`browser`/…), so the non-standard CDN fields were dropped and both unpkg and jsdelivr fell back to the CJS `main`, breaking the `<script>` entry (jsdelivr also regressed once `browser` was removed).

Fixed in v1.48.1 by moving `jsdelivr`/`unpkg` to the package root so `yarn pack` preserves them. Verified the published 1.48.1 on unpkg, jsdelivr, and esm.sh in a real browser (global `_` defined, ESM import works). Thanks again for the original fix!

PR #1406, opened 2025-09-21
test(array/takeRight): add test cases for 'negative' and 'undefined' count handling

sukvvon:
## AS-IS

image

## TO-BE

image


raon0211 (approved):
Thanks for adding test cases!

PR #1410, opened 2025-09-21
test(compat/array/findLastIndex): add test case for 'identity' function default 'predicate'

sukvvon:
## AS-IS

image

## TO-BE

image


raon0211 (approved):

PR #1412, opened 2025-09-21
test(compat/array/pullAll): add test case for default empty array parameter

sukvvon:
## AS-IS

image

## TO-BE

image


raon0211 (approved):
Thanks!

PR #1416, opened 2025-09-21
test(compat/math/{maxBy,meanBy,minBy}): add test cases for default parameters

sukvvon:
## AS-IS

image

## TO-BE

image


raon0211 (approved):
Thanks!

PR #1430, opened 2025-10-01
feat: add esm compat imports;

denisx:
Add esm imports for compat version

Helpful for lodash migrations (like https://gist.github.com/denisx/a06fc2b518ec152d2ef631674ac87b30)

### DIFF
--- a/.github/workflows/autofix.yml
+++ b/.github/workflows/autofix.yml
@@ -19,7 +19,7 @@ jobs:
     steps:
       - uses: actions/checkout@v5
       - run: corepack enable
-      - uses: actions/setup-node@v4
+      - uses: actions/setup-node@v5
         with:
           cache: 'yarn'
           cache-dependency-path: 'yarn.lock'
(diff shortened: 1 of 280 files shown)


stanislaugh (approved):

raon0211:
Hey! Thanks for the PR. Quick question: since importing from `es-toolkit/compat/func` already tree-shakes and only pulls in what you need, do we really need ESM exports? What's the benefit we'd get from them? (Just asking because dual CJS/ESM packages can introduce risks like dual package hazards, so I want to make sure it's worth it.)

denisx:
@raon0211  i see more clear build with esm imports. 'require' not well.
if you care about default require compability, we can make second build folder, such as compat-esm or any )

4 step to future from work experiments (ui build, d - desktop, m - mobile)

Screenshot+2025-10-01+at+23 12 20

a bit clear zip size (-1.5 kb) and more effective raw size (-8.5 kb) from 'require' compat

raon0211:
This was fixed in https://github.com/toss/es-toolkit/pull/1757 and was released in v1.47.1. Thanks for opening a pull request, and looking forward to your next contribution!

PR #1432, opened 2025-10-01
fix(isMatch): handle primitive targets with object source patterns correctly

### LINKED ISSUE #1399: `isMatch` from `es-toolkit/compat` does not behave like `lodash-es/isMatch`
Hello,
While trying to migrate away from lodash to es-toolkit/compat (btw, thanks a lot for this amazing work!) we noticed a few React components not behaving as expected.
I managed to narrow the issue to the `isMatch` function (found while using `find`).

If you run this code below, you'll notice that the same predicate doesn't output the same value whether we use Lodash or Es-Toolkit `find`or `isMatch`.

```js
import lodashEs from "https://esm.sh/lodash-es";
import esToolkit from "https://esm.sh/es-toolkit/compat";

const list = [
  {
    label: "Foo",
    value: "foo"
  },
  {
    label: "Bar",
    value: "bar"
  }
];

const condition = {
  anyKeyOrMissingKey: undefined // Comment me, it doesn't change a thing
};

const foundLodash = lodashEs.find(list, {
  value: condition
});
const foundEtk = esToolkit.find(list, {
  value: condition
});

const isAMatchLodash = lodashEs.isMatch({value: 'bar'}, {value: condition})
const isAMatchEtk = esToolkit.isMatch({value: 'bar'}, {value: condition})

console.clear()
// Lodash and ESToolkit/compat should output the same "undefined" value
console.log("found lodash", foundLodash);
console.log("found estoolkit", foundEtk);
// Root problem is isMatch fn
console.log('isAMatchLodash', isAMatchLodash)
console.log('isAMatchEtk', isAMatchEtk)
```
[Codepen](https://codepen.io/Christophe-Rast/pen/PwPMvov)

It seems also related to this issue about `null`: #750 .

Thanks!

sen2y:
# Fix: Correct isMatch primitive target vs object pattern matching

Fixes #1399

---

## 🐛 Problem

Two incorrect matches versus Lodash were identified.

### Bug 1: Top-level primitive target vs object source (with undefined)

Case:

```ts
isMatch('bar', { anyKey: undefined });
```

Observed behavior before fix:

- Lodash: false ✅
- ES-Toolkit (before): true ❌

Root cause:

- Accessing a non-existent key on a primitive target yields `undefined`:

```ts
'bar'['anyKey']; // -> undefined
source['anyKey']; // -> undefined
```

These two `undefined` values accidentally compare as equal, yielding a false positive.

### Bug 2: Nested level primitive vs empty object

Case:

```ts
isMatch({ value: 'bar' }, { value: {} });
```

Observed behavior before fix:

- Lodash: false ✅
- ES-Toolkit (before): true ❌

Why it is confusing:

- Top-level: `isMatch('bar', {})` should be true (empty object = no constraints)
- Nested: `isMatch('bar', {})` must be false (type mismatch vs object pattern)

Lodash intent:

- Top-level empty object is permissive (matches anything)
- Nested empty object still enforces object type (primitive must not match object)

ES-Toolkit previously treated empty objects as truthy matches at all levels.

---

## 🔧 Solution

1. Reject object-pattern matching against primitive targets

```ts
// If source expects object keys, target must be an object
if (!isObject(target)) {
  return false;
}
```

Fixes Bug 1:

```ts
isMatch('bar', { anyKey: undefined }); // false ✅
isMatch(123, { x: 1 }); // false ✅
```

2. Distinguish top-level vs nested level for empty-object source using stack size

```ts
if (keys.length === 0) {
  // Only at nested levels, primitives must not match {}
  if (stack && stack.size > 0 && !isObject(target)) {
    return false;
  }
  return true;
}
```

Rationale:

- At the top level (`stack.size === 0`), `{}` means "no constraints" -> allow any target
- At nested levels (`stack.size > 0`), `{}` still implies object type -> primitives must fail

Implementation detail:

- `isMatchWith` creates a new stack (`new Map()`) at the entry point.
- On recursion, objects are pushed to the stack (`stack.set(...)`), so `stack.size` naturally reflects depth.

---

## 🧪 Tests

Added to `src/compat/predicate/isMatch.spec.ts`.

1. Primitive target with object source (Bug 1)

```ts
it('returns false when target is primitive and source has object keys (Issue #1399)', () => {
  expect(isMatch('bar', { anyKey: undefined })).toBe(false);
  expect(isMatch(123, { anyKey: undefined })).toBe(false);
  expect(isMatch(true, { anyKey: undefined })).toBe(false);
});
```

2. Nested empty object matching (Bug 2)

```ts
it('returns false when nested target is primitive and source is empty object (Issue #1399)', () => {
  expect(isMatch({ value: 'bar' }, { value: {} })).toBe(false);
  expect(isMatch({ value: 123 }, { value: {} })).toBe(false);
  expect(isMatch({ value: true }, { value: {} })).toBe(false);
});
```

Result:

- ✅ 25/25 tests passed
- ✅ No type errors

---

## 📊 Summary of Changes

Modified files:

- `src/compat/predicate/isMatchWith.ts` (logic updates)
- `src/compat/predicate/isMatch.spec.ts` (new test cases)

Code added (conceptually):

```ts
// 1) Empty-object handling by level (top-level vs nested)
if (keys.length === 0) {
  if (stack && stack.size > 0 && !isObject(target)) {
    return false;
  }
  return true;
}

// 2) Primitive target cannot match object pattern
if (!isObject(target)) {
  return false;
}
```

Only a few lines, but fully aligns with Lodash behavior.

---

## 🧭 Behavior vs Lodash

Before (incorrect):

```ts
isMatch('bar', { anyKey: undefined }); // ES-Toolkit: true ❌, Lodash: false
isMatch({ value: 'bar' }, { value: {} }); // ES-Toolkit: true ❌, Lodash: false
```

After (fixed):

```ts
isMatch('bar', { anyKey: undefined }); // ES-Toolkit: false ✅, Lodash: false
isMatch({ value: 'bar' }, { value: {} }); // ES-Toolkit: false ✅, Lodash: false
isMatch('bar', {}); // ES-Toolkit: true  ✅, Lodash: true
```

---

## ✅ Checklist

- [x] Fix both issues in #1399
- [x] Match Lodash behavior at both top-level and nested levels
- [x] Add unit tests covering the two bug scenarios
- [x] Keep changes minimal (constant-time checks only)
- [x] Preserve performance; no extra traversal or allocations

### DIFF
--- a/src/compat/predicate/isMatchWith.ts
+++ b/src/compat/predicate/isMatchWith.ts
@@ -1,6 +1,8 @@
 import { isObject } from './isObject.ts';
 import { isPrimitive } from '../../predicate/isPrimitive.ts';
+import { getTag } from '../_internal/getTag.ts';
 import type { IsMatchWithCustomizer } from '../_internal/IsMatchWithCustomizer.ts';
+import { argumentsTag, objectTag } from '../_internal/tags.ts';
 import { eq } from '../util/eq.ts';
 
 /**
@@ -46,7 +48,8 @@ export function isMatchWith(target: object, source: object, compare: IsMatchWith
  * - **Primitives**: Matches using strict equality
  *
  * Special cases:
- * - Empty objects, arrays, Maps, and Sets always match any target
+ * - Empty arrays, Maps, and Sets always match any target
+ * - An empty object matches any target at the top level, but at nested levels it only matches object targets
  * - `null` and `undefined` source values have specific matching rules
  * - Circular references are handled using an internal stack to prevent infinite recursion
  *
@@ -151,7 +154,7 @@ function isMatchWithInternal(
 
   switch (typeof source) {
     case 'object': {
-      return isObjectMatch(target, source, compare, stack);
+      return isObjectMatch(target, source, compare, stack, isRoot);
     }
     case 'function': {
       const sourceKeys = Object.keys(source);
@@ -191,7 +194,8 @@ function isObjectMatch(
     source: any,
     stack?: Map<any, any>
   ) => boolean | undefined,
-  stack: Map<any, any> | undefined
+  stack: Map<any, any> | undefined,
+  isRoot = false
 ): boolean {
   if (source == null) {
     return true;
@@ -211,8 +215,21 @@ function isObjectMatch(
 
   const keys = Object.keys(source as any);
 
-  if (target == null || isPrimitive(target)) {
-    return keys.length === 0;
+  if (target == null) {
+    return isRoot && keys.length === 0;
+  }
+
+  if (isRoot) {
+    if (isPrimitive(target)) {
+      target = Object(target);
+    }
+  } else {
+    // Nested object patterns only match plain object-like targets, like lodash's tag comparison.
+    const tag = getTag(target);
+
+    if (tag !== objectTag && tag !== argumentsTag) {
+      return false;
+    }
   }
 
   if (keys.length === 0) {
@@ -229,7 +246,7 @@ function isObjectMatch(
     for (let i = 0; i < keys.length; i++) {
       const key = keys[i];
 
-      if (!isPrimitive(target) && !(key in target)) {
+      if (!(key in target)) {
         return false;
       }
 
--- a/benchmarks/bundle-size/find.spec.ts
+++ b/benchmarks/bundle-size/find.spec.ts
@@ -9,6 +9,6 @@ describe('find bundle size', () => {
 
   it('es-toolkit/compat', async () => {
     const bundleSize = await getBundleSize('es-toolkit/compat', 'find');
-    expect(bundleSize).toMatchInlineSnapshot(`8708`);
+    expect(bundleSize).toMatchInlineSnapshot(`8772`);
   });
 });
--- a/benchmarks/bundle-size/findKey.spec.ts
+++ b/benchmarks/bundle-size/findKey.spec.ts
@@ -14,6 +14,6 @@ describe('findKey bundle size', () => {
 
   it('es-toolkit/compat', async () => {
     const bundleSize = await getBundleSize('es-toolkit/compat', 'findKey');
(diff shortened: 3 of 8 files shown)


src/compat/predicate/isMatchWith.ts
```diff
@@ -211,9 +211,16 @@ function isObjectMatch(
...
+    if (stack && stack.size > 0 && !isObject(target)) {
+      return false;
+    }
     return true;
   }
 
+  if (!isObject(target)) {
+    return false;
+  }
+
```
raon0211, line 234:
```suggestion
  if (!isObject(target)) {
    return false;
  }

  if (keys.length === 0) {
    return true;
  }

```

src/compat/predicate/isMatchWith.ts
```diff
@@ -210,6 +210,10 @@ function isObjectMatch(
     return keys.length === 0;
   }
 
+  if (!isObject(target)) {
+    return false;
+  }
+
```
raon0211, line 234:
```suggestion
    if (stack && stack.size > 0 && !isObject(target)) {
      return false;
    }
    return true;
  }

  if (!isObject(target)) {
    return false;
  }

```

raon0211:
Thanks for your work! Let me refactor this a little bit after merging.

sen2y:
> Thanks for your work! Let me refactor this a little bit after merging.

Thanks for reviewing and refining the code! 🙌  
It’s much clearer thanks to your help.

While working on this fix, I also discovered some behavior differences between Lodash and compat. 
I opened an issue to discuss potential improvements and would love your thoughts when you have a moment: https://github.com/toss/es-toolkit/issues/1504

Thanks again — please let me know if there's anything else I can help with before merging! 😊

raon0211 (dismissed):
Thanks!

raon0211 (approved):
Thanks!

PR #1441, opened 2025-10-06
test(compat/string/truncate): add test case for 'null' and 'undefined' string handling

sukvvon:
## AS-IS

image

## TO-BE

image


raon0211 (approved):
Thanks so much!

PR #1442, opened 2025-10-07
test(compat/array/pullAt): add test case for nested property with array path

sukvvon:
## AS-IS

image

## TO-BE

image


raon0211 (approved):
Thanks so much!

PR #1443, opened 2025-10-07
test(function/flow): add test cases for 'empty function array' and 'this' context preservation

sukvvon:
## AS-IS

image

## TO-BE

image


raon0211 (approved):
Thanks so much!

PR #1445, opened 2025-10-07
fix(isObjectLike): Add type guard return type  

### LINKED ISSUE #1383: `ObjectLike` should return a not nullable/undefined type
When running the code below, the IDE indicates that the value can still be undefined/null, although it can't. `ObjectLike` should potentially use `NotNull<T>` to assert that the returned value cannot be undefined/null

```ts
  const test = undefined;
  if (isObjectLike(test)) {
    test. // error with test can be undefined
  }
```
Image

Can we use the below for the `isObjectLike` function definition instead?
```ts
export function isObjectLike<T>(value: T): value is NonNullable<T> {
  return typeof value === 'object' && value !== null;
}
```

wo-o29:
## Summary  

Fixes #1383   
  
This PR fixes the return type of `isObjectLike` to enable proper TypeScript type guard functionality. Currently, the function returns `boolean`, which prevents it from narrowing types in conditional blocks.  
  
## Changes  
  
- Changed return type from `boolean` to `value is object` in `src/compat/predicate/isObjectLike.ts`
- Updated JSDoc `@returns` annotation to reflect the type guard return type
(The documentation already states that it returns `value is object`)

### DIFF
--- a/src/compat/predicate/isObjectLike.ts
+++ b/src/compat/predicate/isObjectLike.ts
@@ -6,7 +6,7 @@
  * This function can also serve as a type predicate in TypeScript, narrowing the type of the argument to an object-like value.
  *
  * @param {any} value - The value to test if it is an object-like.
- * @returns {boolean} `true` if the value is an object-like, `false` otherwise.
+ * @returns {value is object} `true` if the value is an object-like, `false` otherwise.
  *
  * @example
  * const value1 = { a: 1 };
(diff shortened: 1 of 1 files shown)


raon0211:
We actually have a revised version of the es-toolkit docs now. Can you take a look at the new docs, see if any updates are needed, and open a new pull request if necessary?

wo-o29:
> We actually have a revised version of the es-toolkit docs now. Can you take a look at the new docs, see if any updates are needed, and open a new pull request if necessary?

The new version of the document seems to have addressed all the issues I was looking for!🙃

Thank you 👍👍👍

PR #1446, opened 2025-10-07
fix: resolve circular import between isMatch and isMatchWith  

### LINKED ISSUE #1392: Circular imports isMatch <-> isMatchWith
https://github.com/toss/es-toolkit/blob/main/src/compat/predicate/isMatch.ts#L1

https://github.com/toss/es-toolkit/blob/main/src/compat/predicate/isMatchWith.ts#L1

When using es-toolkit with rollup it complains about the circular dependency:

~~~
Circular dependency: [...]/es-toolkit@1.39.5/node_modules/es-toolkit/dist/compat/predicate/isMatch.mjs -> [...]/es-toolkit@1.39.5/node_modules/es-toolkit/dist/compat/predicate/isMatchWith.mjs -> [...]/es-toolkit@1.39.5/node_modules/es-toolkit/dist/compat/predicate/isMatch.mjs
~~~

Similar to a past issue: https://github.com/toss/es-toolkit/issues/717#issue-2589142741

Is this something that could be fixed again?

I guess one way would be to inline the single line isMatch implementation instead of importing it:

~~~diff
diff --git a/isMatchWith.ts b/isMatchWith.ts
index 6c2a7f192..a76150d98 100644
--- a/isMatchWith.ts
+++ b/isMatchWith.ts
@@ -1,4 +1,3 @@
-import { isMatch } from './isMatch.ts';
 import { isObject } from './isObject.ts';
 import { isPrimitive } from '../../predicate/isPrimitive.ts';
 import type { IsMatchWithCustomizer } from '../_internal/IsMatchWithCustomizer.ts';
@@ -112,7 +111,7 @@ export function isMatchWith(
   ) => boolean | undefined
 ): boolean {
   if (typeof compare !== 'function') {
-    return isMatch(target, source);
+    return isMatchWith(target, source, () => undefined);
   }

   return isMatchWithInternal(
~~~

Thank you!

wo-o29:
## Summary  
  
Fixes #1392  
  
This PR resolves the circular import issue between `isMatch` and `isMatchWith` functions.  
  
## Changes  
  
- Removed the `import { isMatch }` statement from `src/compat/predicate/isMatchWith.ts`  
- Changed the fallback behavior in `isMatchWith` when no comparator is provided from calling `isMatch(target, source)` to calling `isMatchWith(target, source, () => undefined)`  
- This maintains the same functionality while breaking the circular dependency, as `isMatch` internally calls `isMatchWith` with a no-op comparator  
  
The change is functionally equivalent since `isMatch` simply delegates to `isMatchWith` with a comparator that always returns `undefined`, triggering the default matching behavior.

### DIFF
--- a/src/compat/predicate/isMatchWith.ts
+++ b/src/compat/predicate/isMatchWith.ts
@@ -1,4 +1,3 @@
-import { isMatch } from './isMatch.ts';
 import { isObject } from './isObject.ts';
 import { isPrimitive } from '../../predicate/isPrimitive.ts';
 import type { IsMatchWithCustomizer } from '../_internal/IsMatchWithCustomizer.ts';
@@ -112,7 +111,7 @@ export function isMatchWith(
   ) => boolean | undefined
 ): boolean {
   if (typeof compare !== 'function') {
-    return isMatch(target, source);
+    return isMatchWith(target, source, () => undefined);
   }
 
   return isMatchWithInternal(


raon0211 (approved):
Thanks!

PR #1450, opened 2025-10-09
test(compat/array/findLast): add test case for 'identity' function default '_doesMatch'

sukvvon:
## AS-IS

image

## TO-BE

image


raon0211 (approved):
Thanks!

PR #1453, opened 2025-10-09
fix(find, findLast): simplify logic by removing unnecessary checks

wo-o29:
## Summary  
This PR simplifies the `find` function implementation by removing unnecessary type checks and simplifying array handling logic.  

This fix improves test coverage.
| Before | After |
|:------:|:-----:|
| 스크린샷 2025-10-10 오전 1 02 32|  스크린샷 2025-10-10 오전 1 07 46|

There is no test for the case where there is no predicate, so the coverage is not 100%. However, someone else worked on [this part](https://github.com/toss/es-toolkit/pull/1450), so I did not modify it separately!
  
## Changes  
- Removed unnecessary `typeof doesMatch === 'function'` check since `iteratee()` always returns a function  
- Simplified array handling: changed `Array.isArray(source) ? source.slice(fromIndex) : Object.values(source).slice(fromIndex)` to `source.slice(fromIndex)` since at that point in the code, `source` is guaranteed to be an array  

## Additional

The `findLast` function also had the same issue and was fixed. This function also did not have a test for the case where there is no predicate, so the coverage is not 100%, but someone else worked on it :) https://github.com/toss/es-toolkit/pull/1450

### DIFF
--- a/src/compat/array/findLast.ts
+++ b/src/compat/array/findLast.ts
@@ -141,7 +141,7 @@ export function findLast<T>(
 
   const doesMatch = iteratee(_doesMatch);
 
-  if (typeof doesMatch === 'function' && !Array.isArray(source)) {
+  if (!Array.isArray(source)) {
     const keys = Object.keys(source) as Array<keyof T>;
 
     for (let i = fromIndex; i >= 0; i--) {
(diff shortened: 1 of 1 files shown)


sukvvon:
Hello! @wo-o29 I've created a PR(#1454) for what you were concerned about! 😄

wo-o29:
> Hello! @wo-o29 I've created a PR(#1454) for what you were concerned about! 😄

Confirmed, thank you!👍

PR #1454, opened 2025-10-09
test(compat/array/{find,findLast}): add test case for 'boolean' predicate error

sukvvon:
## AS-IS

image

## TO-BE

image


raon0211 (approved):
Thanks a lot!

PR #1456, opened 2025-10-10
Revise isEmpty function documentation

zoulou00:
Updated documentation to remove 'strings' from the list of primitive values considered empty.
It does align with the actually tests.


raon0211 (approved):
Thanks!

PR #1459, opened 2025-10-10
test(compat/string/truncate): add test case for 'separator' not found in truncated string

sukvvon:
## AS-IS

image

## TO-BE

image


dayongkr (approved):

PR #1462, opened 2025-10-11
fix(once): Remove type assertion for better type safety

wo-o29:
## Summary  
  
This PR improves the `once` function by removing the type assertion (`as F`) and using an explicit return type instead. This change enhances type safety while maintaining the same functionality.  
  
## Changes  
  
- Removed `as F` type assertion from the return statement  
- Changed return type from `F` to `(...args: Parameters<F>) => ReturnType<F>`  
- Simplified the first function overload to match the implementation signature  
- Improved type safety by letting TypeScript properly infer and validate the return type

### DIFF
--- a/src/function/once.ts
+++ b/src/function/once.ts
@@ -31,7 +31,9 @@ export function once<F extends (...args: any[]) => any>(func: F): F;
  * initialize(); // Logs: 'Initialized!' and returns true
  * initialize(); // Returns true without logging
  */
-export function once<F extends (() => any) | ((...args: any[]) => void)>(func: F): F {
+export function once<F extends (() => any) | ((...args: any[]) => void)>(
+  func: F
+): (...args: Parameters<F>) => ReturnType<F> {
   let called = false;
   let cache: ReturnType<F>;
 
@@ -42,5 +44,5 @@ export function once<F extends (() => any) | ((...args: any[]) => void)>(func: F
     }
 
     return cache;
-  } as F;
+  };
 }


ahzvenol:
This change is not type-equivalent for overloaded functions.

`(...args: Parameters<F>) => ReturnType<F>` only captures the last visible signature, discarding any preceding overload signatures. 

See an example in the [TypeScript Playground](https://www.typescriptlang.org/play/?#code/CYUwxgNghgTiAEAzArgOzAFwJYHtVNQApYBzABgC54BnDGLVEgSitWQFsAjEGAbgFgAUKEiwEKdNjwFiMclVr1GAbQC6LeGy481AwUJHQ4SNJlz48YEAHl6JBlAgAeAGLwQADwwhUwavEJCJngAXgA+eChUAE9ggB8AwgA6FNJqKijotWDw+AA3HCxgJjDCCTAqFw0XPUMxE0lzeEsQAFkcYCxELBBgV3cvHz9EnIjM+MSUpLSMmOzQiIKikrLTSo1k1Ll0+AAFWCh2EG8Yalcw0fgAJWPkGFQAFWiABxBzvSEwPFpmuwcIFz4ELNdA2P6oRxlVBMPRfVA-dgdLo9YCA0IgqztTrdXpQmFCIQ4cGOQGEABEABYAExk-GCIlYewQgFEZRkgCMZIANPAyTT1B9BIjsSjSZSaXThcjeqS2ZyeXyyQKhEA).

wo-o29:
> This change is not type-equivalent for overloaded functions.
> 
> `(...args: Parameters<F>) => ReturnType<F>` only captures the last visible signature, discarding any preceding overload signatures.
> 

Thank you for your good comments!

I think the problem you mentioned can be solved because the `once` function has two overloads.
```ts
// First overload preserves original type  
export function once<F extends (...args: any[]) => any>(func: F): F;  
  
// Second overload matches impl
export function once<F extends (() => any) | ((...args: any[]) => void)>(  
  func: F  
): (...args: Parameters<F>) => ReturnType<F>
```

[TS-Playground](https://www.typescriptlang.org/play/?#code/LAKFBMFMGMBsEMBOkAEAzArgO2gFwJYD2W6WAFEgOYAMAXCgM66L5aUDaAugJT1YYBbAEaREXANwQYCZOmx4iJNOSp1GzVpV4p+w0ZLAgocJKkw4CxFMWiQA8i0qt4sADwAxFJAAeuSFnAGFDIybhQAXgA+FHgsAE8wgB9gsgA6dKoGeli4rjColAA3Qnxwbkiyc2h6d213AykTWSrLEhtIDy9ff0Dg9NTM7Pi8iOiciqqauoNjGTN5VuscDs8fPwCgkPyx+KSU-sGY4Z5RopKyitAUORwa0G00jMRKLJQABSR4AUg-RAYPSLbFAAJR+GEQWAAKnEAA4rSINEDQYhMayOZywdwkcJLWwOfBOLAuSpYbgGZFYVECQjgfBofCQcBYiK4yAksmgUCEdFEzHkABEABYAEz8jkgbkEjFYsjsfkARn5ABoUPzRTxEdTafTGTKhaLxVq6QymeQ5YqVWr+RrQEA)

If there is anything wrong with my answer, please feel free to let me know :)

raon0211:
Hey, thanks for the PR!

I think we should keep the original type signature here—the function is pretty simple, and `(...args: Parameters<F>) => ReturnType<F>` isn't actually identical to `F` (so the inferred type wouldn't be quite what we'd expect).

Appreciate you taking the time to contribute, and hope to see more from you!

PR #1463, opened 2025-10-11
refactor(compat/predicate/isMatchWith): use optional chaining for 'stack' parameter

sukvvon:
## AS-IS

image

## TO-BE

image


src/compat/predicate/isMatchWith.ts
```diff
@@ -186,7 +186,7 @@ function isObjectMatch(
     source: any,
     stack?: Map<any, any>
   ) => boolean | undefined,
-  stack: Map<any, any> | undefined
+  stack?: Map<any, any>
```
raon0211:
```suggestion
  stack: Map<any, any> | undefined
```

`stack` should be a required parameter here (although it can be undefined)

src/compat/predicate/isMatchWith.ts
```diff
@@ -264,7 +260,7 @@ function isMapMatch(
...
```
raon0211:
```suggestion
  stack: Map<any, any> | undefined
```

src/compat/predicate/isMatchWith.ts
```diff
@@ -298,7 +294,7 @@ function isArrayMatch(
...
```
raon0211:
```suggestion
  stack: Map<any, any> | undefined
```

raon0211 (approved):

PR #1465, opened 2025-10-12
fix(curry, curryRight): Remove unnecessary type assertions 

wo-o29:
## Summary

This PR removes the unnecessary "as any" type assertion from the `curry` and `curryRight` functions in the core module. Since function overloading already provides adequate type safety for external APIs, I believe the type assertion is unnecessary.
  
## Changes  
  
- Removed `as any` type assertion from `curry` function implementation  
- Removed `as any` type assertion from `curryRight` function implementation  
- Removed `as any` type assertion from internal `makeCurry` helper function in both files  
  
The type overloads defined for these functions ensure that TypeScript correctly infers the return types based on the number of parameters. Since the internal implementation returns functions compatible with `(...args: any[]) => any`, the type assertions are redundant and can be safely removed.  
  
## Comment

I think the as any type assertion is unnecessary, but if there's something I'm missing, please let me know!

### DIFF
--- a/src/function/curry.ts
+++ b/src/function/curry.ts
@@ -163,7 +163,7 @@ export function curry(func: (...args: any[]) => any): (...args: any[]) => any {
 
   return function (arg: any) {
     return makeCurry(func, func.length, [arg]);
-  } as any;
+  };
 }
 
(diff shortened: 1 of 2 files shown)


raon0211 (approved):
Thanks!

PR #1467, opened 2025-10-14
perf(meanBy): reduce meanBy computation by removing intermediate array creation

Dohun-choi:
image

Replaced items.map(getValue) + mean() with sumBy(items, getValue) / items.length
to remove unnecessary array creation and reduce iteration overhead.

This change improves performance and memory efficiency without altering behavior.

### DIFF
--- a/src/math/meanBy.ts
+++ b/src/math/meanBy.ts
@@ -1,4 +1,4 @@
-import { mean } from './mean.ts';
+import { sumBy } from './sumBy';
 
 /**
  * Calculates the average of an array of numbers when applying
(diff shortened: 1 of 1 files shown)


Dohun-choi:
Instead of using sumBy,
it's much faster—based on benchmark results—to write it like this:
```javascript
export function meanBy<T>(items: readonly T[], getValue: (element: T) => number): number {
  let nums = 0;

  for (let i = 0; i < items.length; i++) {
    nums += getValue(items[i]);
  }

  return nums / items.length;
}
```
image

**The difference is that `sumBy` passes the index as an extra argument to `getValue`.**

According to benchmarks, inlining similar logic like this performs noticeably better, although I’m not entirely sure why(same results on both Windows and Linux).
If there’s no need to rely on `sumBy` for simplicity, I’ll go ahead and replace it.

raon0211 (approved):
Thanks for pointing this out!

PR #1471, opened 2025-10-16
feat(partition): Type guard support

the5thbeatle:
## Summary

This adds a function overload to `partition` to support type guards, narrowing the type of the output truthy and falsy arrays.

## Changes

* Adds JSDoc and type signature for `partition` where `isInTruthy` is a type guard
* Adds a `partition` test case using a type guard with a type assertion on the narrowed output arrays


raon0211 (approved):
Thanks for your contribution!

PR #1488, opened 2025-10-18
fix(compat/union): should support array-like object

D-Sketon:


raon0211 (approved):
Thanks for fixing this!

PR #1489, opened 2025-10-18
fix(compat/updateWith): improve updater function by using get for value retrieval

D-Sketon:
{14A5D2DE-3749-45B8-9EB9-59E5FC89046F}
{4A5C0E87-A077-4AAB-995D-3B19A680EF2B}


raon0211 (approved):
Thanks for fixing this!

PR #1509, opened 2025-11-07
feat(without, difference): Improve TypeScript type inference for without() and difference()

### LINKED ISSUE #1503: `without` should `Exclude` from return type.
Current implementation:

```typescript
declare function without<T>(array: readonly T[], ...values: T[]): T[];
```

Should be something like 
```typescript
declare function without<T, V extends T>(array: readonly T[], ...values: V[]): Exclude<T, V>[];
```

The current type return does us absolutely no good.

Yeom-JinHo:
Closed #1503

## Description

This PR focuses on **improving TypeScript type inference** for the `without` and `difference` functions.  
By adding smart overloads, the return type now narrows correctly when literal values are passed,  
while preserving full backward compatibility for general use cases.

No runtime logic has been changed — this update only affects **type declarations**.

---

## Changes

- **Type Overloads Added**  
  Introduced three overloads for both `without()` and `difference()`:
  1. `values: []` → returns `T[]`  
  2. `values: V[]` → returns `Exclude<T, V>[]`  
  3. `values: T[]` → fallback for backward compatibility
- **Type Consistency**  
  Ensured `difference()` and `without()` now share the same overload pattern and type behavior.
- **Readonly Style**  
  Unified parameter style to `readonly T[]` for consistency across array utility functions.
- **Documentation Updated**  
  Updated **English (`without.mdx`)**, **Korean (`without.ko.mdx`)**, and **Japanese (`without.ja.mdx`)** documents to reflect:  
  - New overload-based type behavior (`Exclude<T, V>[]`)  
  - Unified parameter and return descriptions  
  - Type narrowing examples (`const result = without(['a','b','c'] as const, 'a') → ('b' | 'c')[]`)

---

## Motivation

- Improve developer experience by providing **more accurate and predictable type inference**.  
- Align the type definitions of `difference()` with `without()` to ensure consistent behavior.  
- Maintain backward compatibility to avoid breaking any existing usage.

---

## Breaking Changes

**None.**  
This change only improves type inference and does not modify runtime behavior or output.


Yeom-JinHo:
I understand that this change might look somewhat excessive in terms of overloading —
please let me know if you think so.

I made this update because I believe it’s an important improvement for achieving accurate type inference.

The overload structure remains minimal (three concise layers) and fully backward-compatible,
preserving both implementation simplicity and interface clarity.

Thx

bvandrc:
Looks good!

Yeom-JinHo:
After thinking about this more, I don't think the current approach is a good idea — the `V extends T` overload makes TypeScript fall back to `V = T`, so most calls (including the examples in the docs) end up returning `never[]` instead of the intended narrowed type. I'll rethink the approach.

PR #1513, opened 2025-11-11
docs: generate llms.txt

seo-rii:
`es-toolkit` is a useful toolkit, but due to its relatively short history, most LLMs don’t naturally understand it yet. For example, if you type `refactor this code with es-toolkit` into an LLM, you’ll often see it calling MCP to fetch and inspect the package code.

This PR adds `llms.txt` and `llms-full.txt`, which include documentation information generated using [vitepress-plugin-llms](https://github.com/okineadev/vitepress-plugin-llms). This makes it easy to provide the docs as context to an LLM.

### DIFF
--- a/docs/.vitepress/shared.mts
+++ b/docs/.vitepress/shared.mts
@@ -146,7 +146,7 @@ export const shared = defineConfig({
     },
     plugins: [
       llmstxt({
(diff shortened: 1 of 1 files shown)


dayongkr:
I've confirmed through preview that the links have no issues.

I'm a bit concerned about the build speed and artifact upload speed, but since it's not a critical issue, I'll go ahead and merge.

Thank you!

PR #1515, opened 2025-11-14
feat(safeJSONParse): add safe JSON parsing utility

Yeom-JinHo:
## Description

This PR introduces a new `safeJSONParse` utility that provides a **non-throwing wrapper around `JSON.parse`**, complementing the existing `isJSON` predicate.  
The goal is to keep **validation** and **transformation** clearly separated while giving consumers a type-safe way to parse JSON strings.

### Changes

- Added `safeJSONParse<T = unknown>(value: unknown): T | null` utility
  - Returns the parsed value when `value` is a string containing valid JSON
  - Returns `null` when `value` is not a string or when parsing fails
  - Mirrors `JSON.parse` semantics (e.g. `"null"` → `null`) while never throwing

- Implemented comprehensive runtime tests for `safeJSONParse`
  - Covers primitives, objects, arrays, nested structures, empty objects/arrays, and large numbers
  - Verifies failure cases for invalid JSON strings and non-string inputs
  - Ensures `"null"` input is handled as a valid JSON string whose parsed value is `null`

- Added type-level tests using `expectTypeOf`
  - Confirms that the generic parameter `T` is respected (`T | null`)
  - Verifies narrowing behavior after `null` checks
  - Asserts the default generic (`T = unknown`) when no type argument is provided

- Documented `safeJSONParse` in a dedicated MDX page
  - Usage examples with valid/invalid JSON strings and non-string inputs
  - Demonstrates generic usage with typed shapes (e.g. `User`)
  - Shows how to combine `safeJSONParse` with `isJSON` for clearer validation + parsing flows

## Motivation

- Avoid duplicating `try { JSON.parse(...) } catch { ... }` patterns across the codebase
- Provide a **clear separation of concerns**:
  - `isJSON(value)` → “Is this a valid JSON string?” (pure validation)
  - `safeJSONParse(value)` → “Safely parse this JSON string into a value” (transformation)
- Improve ergonomics and type safety when dealing with unknown or loosely-typed inputs
- Align JSON-related utilities under a consistent API surface, where predicates live in the predicate category and non-throwing helpers live in the utility category

### DIFF
--- a/docs/ja/reference/util/safeJSONParse.md
+++ b/docs/ja/reference/util/safeJSONParse.md
@@ -1,0 +1,177 @@
+# safeJSONParse
+
+エラーをスローせずに JSON 文字列を安全にパースします。
+
+```typescript
+const result = safeJSONParse(value);
+```
+
+## 使用法
+
+### `safeJSONParse(value)`
+
+JSON 文字列を安全にパースしたい場合に `safeJSONParse` を使用してください。`JSON.parse` 例外を発生させるリスクなくパースできます。値が有効な JSON を含む文字列の場合、パースされた結果を返します。そうでない場合、`safeJSONParse` は `null` を返します。
+
+ユーザー入力、緩く型付けされたデータ、または値が常に有効な JSON 文字列であるという保証がない信頼できないソースを扱う際に特に便利です。
+
+```typescript
+import { safeJSONParse } from 'es-toolkit/util';
+
+// 有効な JSON 文字列
+safeJSONParse('{"name":"JinHo","age":29}'); // { name: "JinHo", age: 29 }
+safeJSONParse('[1, 2, 3]'); // [1, 2, 3]
+safeJSONParse('"hello world"'); // "hello world"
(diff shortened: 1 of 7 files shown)


Yeom-JinHo:
I’m not fully sure if this is the best name for this helper, so I’m open to renaming if you have a better suggestion 🙌🏽

dayongkr:
Sorry for the long wait.

Our priorities have shifted and new functions aren't high on the list right now. See the [contributing guide](https://github.com/toss/es-toolkit/blob/main/.github/CONTRIBUTING.md#41-what-we-accept) for what we accept.

On the API itself, `attempt` already covers this, and it separates the two cases this one can't:

```ts
const [error, value] = attempt(() => JSON.parse(input));
```

`safeJSONParse` returns `null` both for `'null'`, which is a successful parse, and for invalid input, so callers can't tell them apart.

Closing for now. If you'd like to make the case, an [Ideas](https://github.com/toss/es-toolkit/discussions/new?category=ideas) discussion is the place.

Thanks for the contribution.

Issue #956, opened 2025-02-10
`intersectionBy` difference with lodash


bryanwonggy:
The `intersectionBy` function in lodash removes duplicates while the `intersectionBy` function in es-toolkit doesn't.

For `lodash-es`, the `intersectionBy` function removes duplicates. [Doc](https://lodash.com/docs/4.17.15#intersectionBy)

<details>

<summary>See example</summary>

![CleanShot 2025-02-10 at 13 28 52@2x](https://github.com/user-attachments/assets/d821648e-a996-4588-80c5-a591e30959af)

</details>

For `es-toolkit`, the `intersectionBy` function doesn't remove duplicates. [Doc](https://es-toolkit.slash.page/reference/array/intersectionBy.html)

<details>

<summary>See example</summary>

![CleanShot 2025-02-10 at 13 28 57@2x](https://github.com/user-attachments/assets/5aca34c6-4233-47b9-ad92-97ad3dd6aed1)

</details>

Was this intended?

AbdessamadEnabih:
Hello @bryanwonggy, to add your to issue, **Lodash** supports multiple arrays. It computes the intersection across all provided arrays unlike **es-toolkit** it only supports two arrays.

wo-o29:
@raon0211  

I'd like to work on this task. Would it be acceptable to follow the existing interface pattern where it takes at least 2 arrays as arguments and receives a mapper function as the last argument?

raon0211:
Related with https://github.com/toss/es-toolkit/issues/1376

raon0211:
@wo-o29 

Thanks for your help! As you mentioned, we would like to keep the interface where `intersectionBy` receives two arrays and a `mapper` function, but calls like `intersectionBy([2.1, 2.2], [2.3, 3.4], Math.floor)` should return `[2.1]`.

wo-o29:
Thank you for your response!

I'll proceed by keeping the original requirement of receiving at least two arrays and a `mapper` function as essential parameters, and I'll enhance the implementation to support additional arrays as well.

PR #1528, opened 2025-11-21
fix(intersectionBy): remove duplicates when mapper produces same values 

wo-o29:
## Summary  

Fixes: #956 
  
Fixed a bug in `intersectionBy` where it returned duplicate elements from the first array when the mapper function produced identical values for multiple elements. Now correctly returns only the first occurrence of each mapped value.  
  
For example, `intersectionBy([2.1, 2.2], [2.3, 3.4], Math.floor)` now correctly returns `[2.1]` instead of `[2.1, 2.2]`.  
  
## Changes  
  
- Modified `intersectionBy` to use `uniqBy` on the first array before filtering  
- Added test case to verify duplicate removal behavior  

## Comment

- Would it be better to implement it myself rather than using the uniqBy function?

### DIFF
--- a/src/array/intersectionBy.spec.ts
+++ b/src/array/intersectionBy.spec.ts
@@ -24,4 +24,8 @@ describe('intersectionBy', () => {
     const result = intersectionBy(array1, array2, value => value.id);
     expect(result).toEqual([{ id: 2, csv: 1 }]);
   });
+
+  it('should return unique values based on iteratee when arrays have duplicates after mapping', () => {
+    expect(intersectionBy([2.1, 2.2], [2.3, 3.4], Math.floor)).toStrictEqual([2.1]);
+  });
 });


raon0211:
I'm sorry, but as we mentioned in the [original comment](https://github.com/toss/es-toolkit/issues/956#issuecomment-3561488190), we would like to keep the original interface where we only accept two arrays and a mapper function. (We would not like to accept multiple arrays here since it violates our [simplicity principle](https://github.com/toss/es-toolkit/blob/main/.github/CONTRIBUTING.md#13-simplicity).)

However, we should fix a bug where calls like `intersectionBy([2.1, 2.2], [2.3, 3.4], Math.floor)` does not return `[2.1]`. (It is returning `[2.1, 2.2]`.)

wo-o29:
Sorry, I seem to have misunderstood your comment. I'll try to fix the bug you mentioned by submitting a PR.

wo-o29:
@raon0211

 I've fixed the bug and re-created the PR, sorry for the confusion.

PR #1530, opened 2025-11-21
fix(merge, mergeWith): Handle top-level type mismatch consistently

### LINKED ISSUE #1394: `merge` and `mergeWith` treat properties inconsistently with top-level arguments
See comment here: https://github.com/toss/es-toolkit/pull/556#issuecomment-3263219254

`merge(["1"], {a: 2})` gives a different result than `merge({x: ["1"]}, {x: {a: 2}}).x`, which is probably unexpected.

Note that the merging is consistent in the `compat` version.

The type checking performed on each property should possibly instead be done to at the top level of "source" and "target", in order to keep the recursive behavior consistent.

Seol-JY:
## Summary
Fixes #1394

`merge` and `mergeWith` now handle top-level type mismatches consistently with nested property behavior.

### Before  
```typescript
merge(["1"], {a: 2})             // [ '1', a: 2 ]
merge({x: ["1"]}, {x: {a: 2}}).x // { a: 2 }
// Inconsistent behavior
```

### After  
```typescript
merge(["1"], {a: 2})             // { a: 2 }
merge({x: ["1"]}, {x: {a: 2}}).x // { a: 2 }
// Consistent behavior - source type wins
```

## Changes
- Added top-level type checking in merge and mergeWith to match nested property logic from
PR #556
- Added test cases for top-level type mismatch scenarios

### DIFF
--- a/src/object/merge.spec.ts
+++ b/src/object/merge.spec.ts
@@ -118,4 +118,12 @@ describe('merge', () => {
     expect(result).toEqual({ a: 2 });
     expect(result.__proto__).toBe(Object.prototype);
   });
+
+  it('should handle top-level type mismatch consistently with nested behavior', () => {
+    const topLevelResult = merge(['1'] as any, { a: 2 });
+    const nestedResult = merge({ x: ['1'] }, { x: { a: 2 } } as any);
+
+    expect(topLevelResult).toEqual({ a: 2 });
+    expect(nestedResult.x).toEqual({ a: 2 });
+  });
 });
--- a/src/object/merge.ts
+++ b/src/object/merge.ts
@@ -45,6 +45,16 @@ export function merge<T extends Record<PropertyKey, any>, S extends Record<Prope
   target: T,
   source: S
 ): T & S {
+  if (Array.isArray(source)) {
+    if (!Array.isArray(target)) {
+      return merge([] as unknown as T, source);
+    }
+  } else if (isPlainObject(source)) {
+    if (!isPlainObject(target)) {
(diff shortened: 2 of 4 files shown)


raon0211:
Thank you for the PR, and sorry for the long delay in reviewing it! 🙏

This PR addresses a real inconsistency (#1394), but I'd like to close it for two reasons:

**1. The `merge` part has already been fixed by #1553.**

After discussion in #1394, we concluded that `merge` should behave like a recursive `Object.assign` — that is, the **target's structure should be preserved**, at the top level and at every nested level alike. #1553 implemented this, so on `main` today:

```typescript
merge(['1'], { a: 2 });              // ['1'] with .a === 2
merge({ x: ['1'] }, { x: { a: 2 } }).x; // ['1'] with .a === 2 — consistent
```

This also matches lodash's behavior.

**2. This PR resolves the inconsistency in the opposite direction.**

The change here makes the *source* type win (`merge(['1'], { a: 2 })` → `{ a: 2 }`), which conflicts with the direction we settled on in #1553 (target structure wins, like `Object.assign`).

---

That said, your report did surface something still broken: **`mergeWith` was not updated in #1553** and still has the inconsistency on `main`:

```typescript
mergeWith(['1'], { a: 2 }, () => undefined);              // ['1'] with .a === 2
mergeWith({ x: ['1'] }, { x: { a: 2 } }, () => undefined).x; // { a: 2 } — element '1' is lost
```

I'll open a follow-up PR shortly that applies the same fix from #1553 to `mergeWith`.

Thanks again for the contribution. 🙇

PR #1532, opened 2025-11-22
fix(throttle, debounce): Fix throttle with `edges: ['trailing']` behaving like debounce

### LINKED ISSUE #1212: Throttle with custom edges options works just like a debounce.
```ts
const throttledOnScroll = throttle(onScroll, 100, { edges: ["trailing"] })
```

This code expects to work like a "Throttle" function,
Which invokes onScroll function every 100ms while I keep scrolling,
AND Just last trailing call even if scroll ends within delayed 100ms.

but that code just works like debounce. kept scrolling 10s, but just a single call at the last moment. 

Is this intended? 
I couldn't find expected behaviors on `function/throttle.spec.ts` with these options set.

Seol-JY:
## Summary
- Fix #1212
- `throttle` with `{ edges: ['trailing'] }` was behaving like `debounce` (only firing once at the end)

## Problem
The original `throttle` implementation tried to enforce periodic execution by calling `debounced.cancel()` after `throttleMs`, but `cancel()` only cancels without executing. This caused continuous calls to never fire.

## Solution
Added `maxWait` option to `debounce` to enforce periodic execution, then simplified `throttle` to use `debounce` with `maxWait: throttleMs`.

## Implementation Considerations
There were two possible approaches:

| Approach | Pros | Cons |
|----------|------|------|
| **1. Add `maxWait` to `debounce`** (chosen) | Simple `throttle` implementation, no code duplication | Extends `debounce` public API |
| **2. Implement `throttle` independently** | No `debounce` API change | Code duplication, separate logic to maintain |

I chose approach 1 because:
- `throttle = debounce + maxWait` is a common pattern (lodash does this)
- Avoids duplicating timer/invocation logic
- `maxWait` is a useful option for `debounce` users as well

**However**, I'd like maintainer feedback on whether `maxWait` should be:
- A public API option (current implementation)
- An internal-only option (e.g., `_maxWait`)

## Changes
- `src/function/debounce.ts`: Added `maxWait` option to `DebounceOptions`
- `src/function/throttle.ts`: Simplified to use `debounce` with `maxWait`
- `src/function/throttle.spec.ts`: Added test cases for `edges` options

## Test plan
- [x] New tests for `edges: ['trailing']` and `edges: ['leading']` options

### DIFF
--- a/src/function/debounce.ts
+++ b/src/function/debounce.ts
@@ -12,6 +12,13 @@ export interface DebounceOptions {
    * @default ["trailing"]
    */
   edges?: Array<'leading' | 'trailing'>;
+
+  /**
+   * The maximum time the function is allowed to be delayed before it is invoked.
+   * If provided, the function will be invoked at most once per `maxWait` milliseconds,
+   * even if the debounced function is called more frequently.
+   */
+  maxWait?: number;
 }
 
 export interface DebouncedFunction<F extends (...args: any[]) => void> {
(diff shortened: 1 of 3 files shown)


raon0211:
Hello, thanks for opening a pull request. 

While your implementation is correct, we do not want to introduce a new option like `maxWait` for `debounce` -- it adds complexity to the implementation, and it is an option strictly for just implementing `throttle`. So we moved the logic to the `throttle` function instead in this commit: https://github.com/toss/es-toolkit/commit/13bf5203cf3233b403263d8ab6615cfe4671beb7. 

In the meantime, we added the tests you wrote into our codebase. Thanks for helping us!

PR #1535, opened 2025-11-25
test(compat/{negate,add,set}): fix duplicate test titles and enable 'vitest/no-identical-title' error

sukvvon:
## Summary
- Remove duplicate test cases in `negate.spec.ts`
- Rename duplicate test titles in `add.spec.ts` and `set.spec.ts` to be more specific
- Enable `vitest/no-identical-title` as error (remove from warn override)

## Changes
- `negate.spec.ts`: Remove duplicated test case
- `add.spec.ts`: Differentiate test titles for positive, negative, and mixed number cases
- `set.spec.ts`: Differentiate test titles for nested path depths and object/array combinations
- `eslint.config.mjs`: Remove `vitest/no-identical-title` from warn rules (defaults to error)

### AS-IS

image

### TO-BE

image

### DIFF
--- a/src/compat/function/negate.spec.ts
+++ b/src/compat/function/negate.spec.ts
@@ -15,13 +15,6 @@ describe('negate', () => {
     expect(negateFn(2)).toBe(false);
   });
 
-  it('should create a function that negates the result of `func`', () => {
-    const negateFn = negate(isEven);
-
-    expect(negateFn(1)).toBe(true);
-    expect(negateFn(2)).toBe(false);
-  });
-
   it('should create a function that accepts multiple arguments', () => {
     let argCount: any;
     const count = 5;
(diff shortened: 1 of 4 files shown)


src/compat/function/negate.spec.ts
```diff
@@ -15,13 +15,6 @@ describe('negate', () => {
     expect(negateFn(2)).toBe(false);
   });
 
-  it('should create a function that negates the result of `func`', () => {
-    const negateFn = negate(isEven);
-
-    expect(negateFn(1)).toBe(true);
-    expect(negateFn(2)).toBe(false);
-  });
-
```
sukvvon, line 24: It was duplicated

raon0211 (approved):
Thanks so much!

Issue #869, opened 2024-12-03
sortBy with string[]


gurgeous:
Thanks for creating es-toolkit, we are migrating today and love it. Really appreciate all your hard work!

It would be nice if sortBy/orderBy supported an array of strings. This is pretty common IMO and worked with lodash. Ordering by lower case, or by string length, or some kind of category lookup, or natural sort, or first word in a sentence, etc. Thanks again!

raon0211:
Hello :) Let us look into this.

gurgeous:
Great, let me know if I can help! I think `compat/sortBy` handles it appropriately

kaehehehe:
@raon0211 
Hi, Can I give it a try?

raon0211:
Thank you for the suggestion, and I'm glad to hear that you're enjoying es-toolkit!

After some consideration, though, I believe that for arrays of strings, using the native sort directly is much more natural and better ergonomics:

```ts
arr.sort((x, y) => x.localeCompare(y));

// By lower case
arr.sort((x, y) => x.toLowerCase().localeCompare(y.toLowerCase()));

// By string length
arr.sort((x, y) => x.length - y.length);
```

`sortBy` and `orderBy` are designed for sorting arrays of objects by their properties, and extending them to primitives would add complexity to the API without a clear benefit over what modern JavaScript already provides.

So I'll close this issue. Thanks again for your interest in es-toolkit!

PR #1552, opened 2025-12-09
feat(orderBy, sortBy): improve orderBy, sortBy

jidohyun:
This PR improve the `orderBy`, `sortBy` function, as requested in the issue(#869)

Added:
  - sortBy and orderBy string array support: Enhanced both functions to work with primitive arrays (strings, numbers, etc.) by removing T extends object constraint
  - Optional criteria parameters: Made criteria parameter optional with default empty array, enabling direct primitive sorting without explicit functions
  - Optional orders parameter: Added default ascending order to orderBy for simplified usage
  - Comprehensive test cases: Added tests for string arrays, number arrays, empty criteria, and mixed sorting scenarios using consistent ['apple', 'banana', 'cherry'] pattern
  - Enhanced documentation: Updated JSDoc with primitive array examples and usage patterns for both Korean and English docs
  - Type safety improvements: Maintained full type inference while supporting broader input types
  - Backward compatibility: Preserved all existing object sorting functionality without breaking changes

Here is the screenshot of the performance benchmark results:
Screenshot 2025-12-09 at 12 20 33 PM
Screenshot 2025-12-09 at 12 21 03 PM

### DIFF
--- a/docs/ja/reference/array/orderBy.md
+++ b/docs/ja/reference/array/orderBy.md
@@ -1,7 +1,6 @@
 # orderBy
 
-複数の基準と並び順に従って、オブジェクト配列をソートした新しい配列を返します。
-
+複数の基準と並び順に従って、配列をソートした新しい配列を返します。
 ```typescript
 const sorted = orderBy(arr, criteria, orders);
 ```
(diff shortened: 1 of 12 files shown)


raon0211:
Thank you for the PR!

However, for arrays of strings or other primitives, I think using the native sort directly is much more natural and better ergonomics:

```ts
arr.sort((x, y) => x.localeCompare(y));
```

`sortBy` and `orderBy` are designed for sorting arrays of objects by their properties, and extending them to primitives adds complexity to the API without a clear benefit over what modern JavaScript already provides. This also aligns with our design principle of not implementing features that are easily replaceable by modern JS.

So I'll close this PR. Thanks again for your contribution and interest in es-toolkit!

PR #1556, opened 2025-12-15
WIP: feat: add promises methods

### LINKED ISSUE #1549: Promisify - add as a promise util
I'm dealing with an old codebase, and I already replaced many methods from lodash, underscore and bluebird with this library and it works amazingly well! Extra points for having great type inference and typescript support 🙌

But I still have a request that would be great if this library could support. We deal with many old libraries that only use callback methods. Currently we are using Bluebird promisify or even node:util with promisify.

http://bluebirdjs.com/docs/api/promise.promisify.html

https://nodejs.org/api/util.html#utilpromisifyoriginal

Bluebird doesn't play very well with TypeScript, and it is very outdated. Node is only available on the server, not in the client browser, so for frontend usage, it is not that direct. It would be wonderful if this library had this method. 

Point of attention: sometimes we need to bind the context of the function to the promisify. I wonder if this could be done automatically somehow? Example from bluebird docs:

```js
var redisGet = Promise.promisify(redisClient.get, {context: redisClient});
redisGet('foo').then(function() {
    //...
});
```

Thanks!

rubnogueira:
Fixes: https://github.com/toss/es-toolkit/issues/1549

This PR adds a set of utilities for converting between Promise-based and callback-based APIs. These are especially useful when working with legacy Node.js code or when you need to support both styles.

Inspired on:
http://bluebirdjs.com/docs/api/promisification.html
https://googleapis.dev/nodejs/promisify/latest/global.html

### What's included

**Converting callbacks to promises:**
- `promisify(fn)` - Takes a function that uses a callback and returns one that returns a Promise
- `promisifyAll(obj)` - Does the same for all methods in an object, adding `*Async` suffix
- `fromCallback(fn)` - Creates a Promise from a one-shot callback operation
- `fromCallbackAll(obj)` - Adds `*Async` methods for callback-based objects

**Converting promises to callbacks:**
- `callbackify(fn)` - Takes an async function and returns one that accepts a callback
- `callbackifyAll(obj)` - Does the same for all async methods, adding `*Callback` suffix
- `asCallback(promise, cb)` - Attaches a callback to an existing promise (also exported as `nodeify`)
- `asCallbackAll(obj)` - Adds `*Callback` methods for async objects

**Supporting both styles at once:**
- `withCallback(fn)` - Wraps an async function so it can be called either way
- `withCallbackAll(obj)` - Does the same for all async methods in an object

### Why these exist

When maintaining libraries or working with mixed codebases, you often need to support both Promise and callback patterns. Instead of writing boilerplate for each conversion, these utilities handle it in a type-safe way.

The "All" variants are particularly useful for wrapping entire modules (like `fs`) or class instances without touching each method individually.

This is still a WIP and I'm still testing this.

### DIFF
--- a/docs/reference/promise/promisify.md
+++ b/docs/reference/promise/promisify.md
@@ -1,0 +1,82 @@
+# promisify
+
+Converts a callback-based function to a Promise-based function.
+
+Takes a function that accepts a Node.js-style callback `(error, result) => void` as its last argument and returns a new function that returns a Promise. This is useful for modernizing legacy codebases that use callback patterns.
+
+## Signature
+
+```typescript
+function promisify<Args extends unknown[], Result>(
+  fn: (...args: [...Args, Callback<Result>]) => void,
+  options?: PromisifyOptions
+): (...args: Args) => Promise<Result>;
+```
+
+### Parameters
+
+- `fn` (`(...args: [...Args, Callback<Result>]) => void`): A function that accepts a callback as its last argument. The callback should follow the Node.js convention: `(error, result) => void`.
+- `options` (`PromisifyOptions`, optional): Configuration options.
+  - `context` (`object`, optional): The `this` context to bind when calling the function. Useful for object methods.
+
+### Returns
+
+(`(...args: Args) => Promise<Result>`): A new function that returns a Promise. The Promise resolves with the callback's result value, or rejects with the callback's error.
+
+## Examples
+
+### Basic usage
+
+```typescript
+import { promisify } from 'es-toolkit/promise';
+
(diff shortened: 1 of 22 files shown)


raon0211:
Hello, thanks for the kind words and for taking the time to open a PR. We really appreciate it!

That said, we've decided not to include promisify in es-toolkit. es-toolkit focuses on providing utilities for modern browsers and Node.js environments, where APIs are already promise-based. promisify mainly serves legacy callback-style code, so it falls outside the scope of what we aim to provide. For Node.js, [util.promisify](https://nodejs.org/api/util.html#utilpromisifyoriginal) already covers this well.

Thanks again for the PR and for using es-toolkit! 🙏

Issue #1558, opened 2025-12-16
Redundancy between JSdoc types and function signature types — remove JSdoc types?


T3sT3ro:
Currently functions include type information both in JSdoc comments and in the function signatures. AFAIK this is due to the `toDocumentationItems.ts` job used to generate docs. This leads to redundancy, since both need to be in sync so you can't update one without the other. It could be nice to check if the types for documentation could be extracted directly from the function instead. 

For example `array/pull.ts`:
```ts
/**
 * Removes all specified values from an array.
 *
 * This function changes `arr` in place.
 * If you want to remove values without modifying the original array, use `difference`.
 *
 * @template T, U
 * @param {T[]} arr - The array to modify.
 * @param {unknown[]} valuesToRemove - The values to remove from the array.
 * @returns {T[]} The modified array with the specified values removed.
 *
 * @example
 * const numbers = [1, 2, 3, 4, 5, 2, 4];
 * pull(numbers, [2, 4]);
 * console.log(numbers); // [1, 3, 5]
 */
export function pull<T>(arr: T[], valuesToRemove: readonly unknown[]): T[] {
  // ...
}
```

Could be replaced with:
```ts
/**
 * Removes all specified values from an array.
 *
 * This function changes `arr` in place.
 * If you want to remove values without modifying the original array, use `difference`.
 *
 * @template T, U
 * @param arr - The array to modify.
 * @param valuesToRemove - The values to remove from the array.
 * @returns The modified array with the specified values removed.
 *
 * @example
 * const numbers = [1, 2, 3, 4, 5, 2, 4];
 * pull(numbers, [2, 4]);
 * console.log(numbers); // [1, 3, 5]
 */
export function pull<T>(arr: T[], valuesToRemove: readonly unknown[]): T[] {
  // ...
}
```

This could be cleaned up in all of the functions. I already saw some JSDocs skipping types in compat directory, so maybe it is already supported and doesn't require any other tinkering?

I tried working on it for a bit but I am not sure which workflows you use to test how it how it works and how it compares to the current implementation. I also see there is currently a `generate-docs.mts` script, but I don't get for now how it fits the rest of the project and pipeline. I tried to run the following command on main branch but it fails for `once.ts` for me, so I am not sure if it is maintained:

```bash
cd .scripts/docs
deno run --allow-read --allow-write --allow-env --allow-net generate-docs.mts
```

Several issues arise currently after running this script.

T3sT3ro:
[...]

PR #1559, opened 2025-12-16
fix: clean up JSDocs causing errors during generate-docs.mts script

T3sT3ro:
Running generate-docs.mts script reported errors due to missing JSDoc for many function overloads. This PR cleans up those cases, adding comments when they are missing or removing duplicate/redundant function declarations.

I have however some suspicions that generate-docs.mts script has not been maintained for some town now, and may be redundant already. I tried running it by itself and starting `docs:build/dev` tasks, but have seen no changes to the files.

It is probably related to #1558 which I wanted to develop in the first place, but this is for the time being as far as I got. More help regarding workflow for documentation generations in this project is required.

---

It turns out that it is not so easy with partition and reduceAsync, because both of those files provide separate declaration and definition of functions. After I tried removing declarations which duplicate definitions, the typecheck task fails. After I add them back, the `generate-docs.mts` fails. I am not sure how to proceed from there.

### DIFF
--- a/src/compat/array/countBy.ts
+++ b/src/compat/array/countBy.ts
@@ -20,11 +20,32 @@ export function countBy<T>(
   iteratee?: ValueIteratee<T>
 ): Record<string, number>;
 
+/**
+ * Creates an object composed of keys generated from the results of running each element of collection through
+ * iteratee. The corresponding value of each key is the number of times the key was returned by iteratee. The
+ * iteratee is invoked with one argument: (value).
+ *
+ * @param collection The collection to iterate over.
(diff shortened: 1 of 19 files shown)


raon0211:
Hi @T3sT3ro, thank you for taking the time to look into this!

Your suspicion was correct — `generate-docs.mts` had not been maintained for a while and was no longer used in our documentation workflow. We removed the script entirely in #2021, so the JSDoc errors it reported are no longer an issue.

Since the script this PR was addressing has been deleted, I'll close this PR. We really appreciate your contribution and the detailed investigation!

PR #1560, opened 2025-12-18
fix(throttle): prevent maxWait from being set when both edges are disabled

### LINKED ISSUE #1494: Discrepancy between lodash and compat throttle with both trailing and leading disabled
While I was sanity-checking the reproduction on [this](https://github.com/toss/es-toolkit/issues/1493) issue, I've found that there's a discrepancy between lodash and compat throttle:

```ts
import lodashThrottle from 'lodash-es/throttle';
import compatThrottle from 'es-toolkit/compat/throttle';

const WAIT = 100;
const OPTIONS = {leading: false, trailing: false};

const lodashThrottled = lodashThrottle(() => console.log('lodashThrottled'), WAIT, OPTIONS);
const compatThrottled = compatThrottle(() => console.log('compatThrottled'), WAIT, OPTIONS);

// These seem to be mostly consistent on my machine
setInterval(lodashThrottled, 1);
setInterval(compatThrottled, 1);

// Things start to get inconsistent
setInterval(lodashThrottled, 100);
setInterval(compatThrottled, 100);

// Lodash now executes far more rarely, while compat seems to execute according to delay
// The lodash timing appears to grow exponentially compared to compat
setInterval(lodashThrottled, 115);
setInterval(compatThrottled, 115);

// Note: system/browser workload seems to increase the timing inconsistencies as well

```

This is a bit extreme edge case considering this seems to happen only with both trailing and leading edges disabled, but configuration like this is still implicitly allowed (even though the effects of doing so are not generally established), so I thought it deserved mention.

constantly-dev:
## Summary
This PR aligns the `throttle` implementation in `compat` with Lodash's behavior by explicitly preventing `maxWait` from being passed to `debounce` when both `leading` and `trailing` options are `false`.

## Why is this necessary? (Analysis)
While analyzing Lodash's implementation vs `es-toolkit/compat`, I found a critical difference in how `debounce` handles `maxWait`:

1.  **Lodash's `debounce`**: Even if `maxWait` is set, it checks if `trailing` is enabled before executing on the trailing edge. If `trailing` is `false`, it does **not** execute, even if the timer expires.
2.  **es-toolkit/compat `debounce`**: In `src/compat/function/debounce.ts`, the wrapper logic checks `if (Date.now() - pendingAt >= maxWait)`. If true, it immediately calls `func.apply(...)` **without checking the `trailing` option**.

### Proof
Currently, `throttle` unconditionally passes `throttleMs` as `maxWait`. This creates a risk where `es-toolkit` might execute the function due to `maxWait` timeout, whereas Lodash would correctly treat it as a no-op when `{ leading: false, trailing: false }`.

To ensure strict compatibility without altering the complex `debounce` wrapper logic, this PR modifies `throttle` to only pass `maxWait` when at least one edge is enabled.

## Changes
- Modified `throttle.ts` to set `maxWait` to `undefined` if both `leading` and `trailing` are false.
- Added a regression test case in `throttle.spec.ts`.

## Related Issue
Fixes #1494

### DIFF
--- a/src/compat/function/throttle.ts
+++ b/src/compat/function/throttle.ts
@@ -117,7 +117,6 @@ export function throttle<T extends (...args: any) => any>(
  *   throttledFunction(); // Will log 'Function executed'
  * }, 1000);
  */
-
 export function throttle<F extends (...args: any[]) => any>(
   func: F,
   throttleMs = 0,


src/compat/function/throttle.ts
```diff
@@ -117,6 +117,7 @@ export function throttle<T extends (...args: any) => any>(
  *   throttledFunction(); // Will log 'Function executed'
  * }, 1000);
  */
+
```
raon0211:
```suggestion
```

raon0211:
Actually, the real root cause is more subtle, let us find a better way to solve this.

The real root cause is that in the following code:

```typescript
import lodashThrottle from 'lodash-es/throttle';

const WAIT = 100;
const OPTIONS = {leading: false, trailing: false};

const lodashThrottled = lodashThrottle(() => console.log('lodashThrottled'), WAIT, OPTIONS);

setInterval(lodashThrottled, 115);
```

When the debounce timer in lodash expires before the next call arrives, `trailingEdge` fires and sets `timerId = undefined` ([#](https://github.com/lodash/lodash/blob/4.17.23/lodash.js#L10484)). This means every subsequent call enters `leadingEdge` ([#](https://github.com/lodash/lodash/blob/4.17.23/lodash.js#L10516)) instead of reaching the `maxing` branch where `invokeFunc` would actually be called ([#](https://github.com/lodash/lodash/blob/4.17.23/lodash.js#L10519)).

The maxing logic that guarantees execution with `maxWait` is never reached because the `timerId === undefined` check comes first—so we keep resetting through `leadingEdge` instead of hitting the forced invocation path.

constantly-dev:
> Actually, the real root cause is more subtle, let us find a better way to solve this.
> 
> The real root cause is that in the following code:
> 

@raon0211 

I see what you mean now—thanks for pointing out the real root cause! 🙏

After reviewing `debounce.ts`, I realized that the `maxWait` logic currently triggers execution without checking the `trailing` option.

To align with Lodash's behavior, I think we should guard the execution with `trailing`, but **crucially, still reset the timer state** to maintain the cycle.

Here is the implementation I have in mind:

```typescript
// src/compat/function/debounce.ts (line 200)

// TO-BE
if (Date.now() - pendingAt >= maxWait) {
    // 1. If trailing is enabled, execute the function
    if (trailing) {
      result = func.apply(this, args);
    }

    // 2. Regardless of execution, reset the timer state
    // This ensures the cycle is reset even if we skipped execution (trailing: false)
    pendingAt = Date.now();
    _debounced.cancel();
    _debounced.schedule();
    
    return result;
}
```

This approach prevents unwanted execution when `trailing: false`, while ensuring the timer cycle is correctly reset as you mentioned.

I tested both the exact scenario you described and our current implementation:

image

<br/>
<br/>

Does this direction look right to you? Or if you had a different implementation strategy in mind, please let me know! 
I’d be happy to proceed either way. 😊


Thanks once again!!

Issue #1557, opened 2025-12-16
Add index argument to partition predicate


T3sT3ro:
It would be nice to have an index argument for the `partition` predicate, to make the API more similar to the already existing array functions like map, filter, reduce and others.

For example:

```ts
const [odds, evens] = partition(foo, (x, i) => i % 2 === 1);
```

T3sT3ro:
While at it I noticed several other functions that operate on arrays could use the index as well. For starters it would be worth to take a look at the array utils.

T3sT3ro:
Following the logic of "least surprise", what do you think that the array functions resembled the regular array function's signature as closely with all 3 parameters — value, index and array?
```ts
map(
   callbackfn: (value: T, index: number, array: readonly T[]) => unknown, 
   thisArg?: any
): unknown[]
```

raon0211:
As you mentioned, I think there should be three arguments (value, index, and the original array) given to array functions.

T3sT3ro:
In that case I can go over array functions and update their callback functions. Should I also do so for the lodash compat functions, or leave them as is?

raon0211:
For lodash-compatible functions, we actually want to preserve lodash's original behavior—so no changes needed here. Thanks for helping out though!

PR #1561, opened 2025-12-19
Add index and array parameters to array function callbacks

T3sT3ro:
Implements #1557 — adding `index` and `array` parameters when appropriate to several array function callbacks:
- flatMap
- flatMapDeep
- uniqBy
- minBy
- maxBy
- keyBy
- countBy
- groupBy
- takeRightWhile
- takeWhile
- partition
- zipWith

For now those are the functions I am aware of where the callback with index/array makes sens. If you have other propositions comment in here and I will add it.

### DIFF
--- a/docs/reference/array/zipWith.md
+++ b/docs/reference/array/zipWith.md
@@ -36,7 +36,7 @@ zipWith([1, 2], [10, 20, 30], (a, b) => (a ?? 0) + (b ?? 0));
 #### Parameters
 
 - `arrs` (`Array<readonly T[]>`): The arrays to combine.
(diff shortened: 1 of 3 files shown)


T3sT3ro:
I see that the tests for compat of `uniqBy` and `unionBy` failed due to extra parameters being passed to the callback. I have wrapped the compatibility function iteratees with `ary(fn ,1)` to strip down extra index and array parameters, I think that's the safest option.

T3sT3ro:
Those build errors look to be external and irrelevant of this PR.

raon0211 (approved):
Thanks so much!

PR #1563, opened 2025-12-19
feat(clone): enhance error cloning to support AggregateError

D-Sketon:
- Added AggregateError handling with `new Constructor(obj.errors, obj.message, { cause: obj.cause })`.
- Used `Object.assign(newError, obj)` to copy custom properties.
- Removed explicit `newError.name = obj.name;` to avoid conflicts with high-version Vitest and Node.js `assert` module, which expect error names to be set by the constructor or inherited properties, ensuring proper assertion behavior. ref: https://github.com/vitest-dev/vitest/pull/5876


raon0211 (approved):
Thanks so much!

PR #1568, opened 2026-01-02
feat(isDate, isRegExp): support cross-realm

haejunejung:
### Problem

`instanceof` fails when checking objects created in different JavaScript realms (e.g., iframes, `vm` contexts). Each realm has its own set of built-in constructors, so `value instanceof RegExp` returns `false` for a RegExp created in another realm.

```javascript
const vm = require('node:vm');
const context = vm.createContext({});
const regex = vm.runInContext('/test/', context);

regex instanceof RegExp;  // false
Object.prototype.toString.call(regex) === '[object RegExp]';  // true

date instanceof Date; // false
Object.prototype.toString.call(date) === '[object Date]'; // true
```

### Solution

Replace `instanceof` with `Object.prototype.toString.call()`, which checks the object's internal `[[Class]]` slot and works reliably across realms.

### Changes

- Changed type checking to use `Object.prototype.toString.call()` instead of `instanceof`
- Added cross-realm tests using `node:vm`

### Reference

- [MDN - instanceof and multiple realms](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/instanceof#instanceof_and_multiple_realms)

### DIFF
--- a/src/predicate/isDate.spec.ts
+++ b/src/predicate/isDate.spec.ts
@@ -1,4 +1,5 @@
 import { describe, expect, it } from 'vitest';
+import { runInNewContext } from 'node:vm';
 import { isDate } from './isDate';
 
 describe('isDate', () => {
@@ -23,5 +24,9 @@ describe('isDate', () => {
       date: new Date(),
     };
     expect(isDate(obj.date)).toBe(true);
+  });
+
+  it('should return `true` for cross-realm Date', () => {
+    expect(isDate(runInNewContext('new Date()'))).toBe(true);
   });
 });


raon0211:
Hey, thanks for the PR! We're not sure cross-realm support is a common enough use case for us to prioritize. es-toolkit aims to keep implementations simple and cover ~85% of real-world scenarios—edge cases aren't really our focus (more on that in our [contributing guide](https://github.com/toss/es-toolkit/blob/main/.github/CONTRIBUTING.md#13-simplicity)).

For checking RegExp or Date values, we think `instanceof` is the more straightforward approach, so we'll stick with what we have.

Appreciate the contribution though—hope to see more from you!

PR #1571, opened 2026-01-07
fix(flattenObject): exclude root level empty objects and arrays from result

youdame:
## Summary

Root level empty objects and empty arrays were being included in the flattened result, which contradicts the documented behavior.

### Before (bug)
```typescript
flattenObject({ empty: {}, emptyArray: [], nullValue: null })
// Result: { empty: {}, nullValue: null }  ❌
```

### After (fix)
```typescript
flattenObject({ empty: {}, emptyArray: [], nullValue: null })
// Result: { nullValue: null }  ✅
```

## Changes

- Root level empty objects (`{}`) are now excluded from the result
- Root level empty arrays (`[]`) are now excluded from the result
- Nested empty objects/arrays are still preserved (e.g., `{ a: { b: {} } }` → `{ 'a.b': {} }`)

## Test

Added test cases for:
- Root level empty objects and arrays
- Nested empty arrays

### DIFF
--- a/src/object/flattenObject.ts
+++ b/src/object/flattenObject.ts
@@ -41,7 +41,7 @@ export function flattenObject(object: object, { delimiter = '.' }: FlattenObject
 function flattenObjectImpl(object: object, prefix: string, delimiter: string): Record<string, any> {
   const result: Record<string, any> = {};
   const keys = Object.keys(object);
-  const isRootLevel = prefix === '';
+  const isNotRootLevel = prefix !== '';
 
   for (let i = 0; i < keys.length; i++) {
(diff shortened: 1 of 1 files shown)


raon0211:
Hi, thanks for the PR! After some discussion, we decided to preserve empty arrays to stay consistent with other libraries like [flat](https://www.npmjs.com/package/flat). We've updated the implementation here: https://github.com/toss/es-toolkit/commit/5fa79749f0bc760fb6ccf82457b86ff7b17dcb8a.

Thanks again for the contribution—hope to see more from you!

PR #1573, opened 2026-01-11
fix[merge, toMerged, mergeWith]: improve return type accuracy with MergeDeep

### LINKED ISSUE #1554: merge typing is wrong
Typing of `merge` function:
```ts
declare function merge<T extends Record<PropertyKey, any>, S extends Record<PropertyKey, any>>(target: T, source: S): T & S;
```
Does not describe a true deep merge. It only says: “return something that has all properties of T and S.” This is a shallow structural type and does not capture recursive merging behavior.

I suggest using following instead:
```
export type OverrideDeep<Base, Override> = Base extends object
    ? Override extends object
        ? {
              [K in keyof Base | keyof Override]: K extends keyof Override
                  ? K extends keyof Base
                      ? OverrideDeep<Base[K], Override[K]>
                      : Override[K]
                  : K extends keyof Base
                    ? Base[K]
                    : never;
          }
        : Override
    : Override;
```

joseph0926:
This is a breaking change PR(type). Please review it carefully.
Since this is a breaking change PR, feel free to close it anytime if it's unnecessary

Closes #1554

## Summary

- Introduce `MergeDeep<T, S>` type for accurate deep merge type inference
- Update documentation return type from `T & S` to `MergeDeep<T, S>`

## Background

Upon investigating the git history, I found that `merge` was **originally designed as a deep merge function** from the very first commit (`0054dc81`):

> "This function performs a deep merge, meaning nested objects and arrays are merged recursively."

The reason it's `T & S` rather than a deep merge type is likely due to the philosophy of being a perfect lodash replacement (this is just my guess, so it might not be correct...).

This PR aligns the type system with the original design intent.

### Type Definition
```typescript
// Before
export function merge<T, S>(target: T, source: S): T & S;

// After  
export function merge<T, S>(target: T, source: S): MergeDeep<T, S>;
```

- Recursive object property merging
- Proper null handling ({ a: null } + { a: [...] } → { a: [...] })
- undefined source preservation (doesn't overwrite)
- NonPlainObject replacement (Date, RegExp, Map, Set, etc.)
- Array element type merging

## Breaking Change (Type Only)
- This is a type-only breaking change. Runtime behavior is unchanged.

## Test
- yarn test (37 failures occur, but this is an external package test in the node_modules/.pnpm/ path)
- yarn typecheck

### DIFF
--- a/src/object/merge.spec.ts
+++ b/src/object/merge.spec.ts
@@ -203,4 +203,49 @@ describe('MergeDeep type', () => {
     const result = merge(target, source);
     expectTypeOf(result.l1.l2.l3).toEqualTypeOf<{ a: number; b: number }>();
   });
+
+  it('should preserve optional properties', () => {
+    type Result = MergeDeep<{ a?: number; b: string }, { c?: boolean }>;
+    type Expected = { a?: number; b: string; c?: boolean };
+    expectTypeOf<Result>().toEqualTypeOf<Expected>();
+  });
+
+  it('should preserve optional properties in nested objects', () => {
+    type Result = MergeDeep<{ nested: { a?: number } }, { nested: { b?: string } }>;
+    type Expected = { nested: { a?: number; b?: string } };
+    expectTypeOf<Result>().toEqualTypeOf<Expected>();
+  });
+
+  it('should handle array + object merge (array as target)', () => {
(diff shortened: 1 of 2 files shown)


joseph0926:
Upon reflection, this PR seems to have been modified in the wrong direction, so I'm closing it.

PR #1579, opened 2026-01-14
fix(keyBy): improve type safety with Partial<Record<K, T>>

eunwoo-levi:
## Summary
Improve type safety of `keyBy` by changing return type from `Record<K, T>` to `Partial<Record<K, T>>`.

## Motivation
The current type signature incorrectly claims all keys exist in the result, but runtime behavior doesn't guarantee this. Accessing non-existent keys returns `undefined`, yet TypeScript thinks it's type `T`, leading to potential runtime errors.

### Current Problem:
```ts
const users = [{ id: 1, name: "Alice" }];
const byId = keyBy(users, (u) => u.id);
const user = byId[999]; // Type says: { id: number; name: string }
user.name; //  X - Runtime: Cannot read property 'name' of undefined
```

### After This PR:
```ts
const user = byId[999]; // Type: { id: number; name: string } | undefined
user?.name; // O - Safe access with optional chaining
```

## Changes
- src/array/keyBy.ts  : Changed return type to `Partial<Record<K, T>>`
- src/array/keyBy.spec.ts. : Added test for type safety verification

## Breaking Change
**Type-only breaking change** - No runtime behavior affected

### Migration Required:
```ts
// Before
const value = result[key].property;
// After
const value = result[key]?.property; // or null check
```
Justification: Prevents runtime errors by aligning types with actual behavior.

### DIFF
--- a/src/array/keyBy.spec.ts
+++ b/src/array/keyBy.spec.ts
@@ -56,7 +56,7 @@ describe('keyBy', () => {
     expect(result).toEqual({});
   });
 
-  it('should return Partial<Record> for type safety', () => {
+  it('should return undefined for non-existent keys', () => {
     const people = [
       { name: 'mike', age: 20 },
       { name: 'jake', age: 30 },
@@ -64,8 +64,6 @@ describe('keyBy', () => {
 
     const result = keyBy(people, person => person.name);
 
-    expect(result['mike']?.name).toBe('mike');
-    expect(result['jake']?.age).toBe(30);
     expect(result['nonexistent']).toBeUndefined();
   });
 });


raon0211:

Your point is valid. Technically speaking, `keyBy` could return `undefined` for missing keys. But in practice, being this strict tends to create a lot of noise in everyday code. TypeScript itself makes a similar tradeoff: without `noUncheckedIndexAccess` (which is off by default), it lets you write:

```typescript
const arr = [1, 2, 3];
const element = arr[4]; // TypeScript infers `number`, not `number | undefined`
```

It's a tough call, but es-toolkit follows TypeScript's default strict mode behavior, so we'll pass on this one. Appreciate the suggestion though!

eunwoo-levi:
> Hello, thanks for the pull request!
> 
> Your point is valid. Technically speaking, `keyBy` could return `undefined` for missing keys. But in practice, being this strict tends to create a lot of noise in everyday code. TypeScript itself makes a similar tradeoff: without `noUncheckedIndexAccess` (which is off by default), it lets you write:
> 
> ```ts
> const arr = [1, 2, 3];
> const element = arr[4]; // TypeScript infers `number`, not `number | undefined`
> ```
> 
> It's a tough call, but es-toolkit follows TypeScript's default strict mode behavior, so we'll pass on this one. Appreciate the suggestion though!

Thanks for the feedback! 
I understand the tradeoff more clearly now while strict typing improves safety, practical usability in day-to-day development matters as well. Overly strict types can create friction and extra checks in common paths!

PR #1580, opened 2026-01-14
fix(zipObject): update zipObject return type to include undefined

eunwoo-levi:
### Problem

The `zipObject` function had a type safety issue where the return type didn't match the actual runtime behavior.

#### Current behavior:
Return type: `Record<P, V>`
Actual runtime: Values can be `undefined` when keys array is longer than values array
Documentation already mentions this behavior, but types didn't reflect it

```ts
const result = zipObject(['a', 'b', 'c'], [1, 2]);
// Type: Record<string, number>
// Runtime: { a: 1, b: 2, c: undefined }
result['c'].toFixed(2);  // X - Runtime error!
// TypeScript thinks this is safe, but it crashes at runtime
```

### Solution
Changed return type from `Record<P, V>` to `Record<P, V | undefined>` to accurately reflect that values can be undefined.

```ts
const result = zipObject(['a', 'b', 'c'], [1, 2]);
// Type: Record<string, number | undefined>
result['c'].toFixed(2);  // X - TypeScript error (caught at compile time!)
// Safe usage
if (result['c'] !== undefined) {
  result['c'].toFixed(2);  // O - Safe!
}
```

### Modified Files
- `src/array/zipObject.ts`
  - Updated return type: `Record<P, V>` → `Record<P, V | undefined>`
  - Updated JSDoc to reflect the correct return type
- `src/array/zipObject.spec.ts`
  - Added type safety test to verify correct type inference
  - Ensures TypeScript properly enforces undefined checks

### DIFF
--- a/src/array/zipObject.spec.ts
+++ b/src/array/zipObject.spec.ts
@@ -9,4 +9,11 @@ describe('zipObject', () => {
 
     expect(zipObject(['a', 'b'], [1, 2, 3])).toEqual({ a: 1, b: 2 });
   });
+
+  it('returns undefined for keys without a corresponding value', () => {
+    const result = zipObject(['a', 'b', 'c'], [1, 2]);
+
+    expect(result).toEqual({ a: 1, b: 2, c: undefined });
+    expect(result.c).toBeUndefined();
+  });
 });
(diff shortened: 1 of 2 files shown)


raon0211:

Your point is valid; but as we discussed in #1579, es-toolkit follows TypeScript's default strict mode behavior, so we'll pass on this one. Appreciate the suggestion though!

PR #1581, opened 2026-01-14
feat(fromPairs): implement fromPairs function

eunwoo-levi:
## Description
Adds the `fromPairs` function to the main `es-toolkit` library (`src/array`). This function converts an array of key-value pairs into an object, providing the inverse operation of `toPairs`.
Currently, `fromPairs` only exists in the `compat` module. This PR brings it to the main library with optimized performance and enhanced type safety.

## Motivation
- **High demand**: `lodash.fromPairs` has 130k+ weekly npm downloads
- **Common use case**: Converting key-value pairs to objects is a frequent operation
- **Performance**: Main library version is faster (no runtime checks)
- **Type safety**: Leverages TypeScript generics for better type inference
- **Bundle size**: Zero dependencies, minimal footprint

## Implementation Details
**Key differences from compat version:**
- Cleaner type signature: `Record<K, V>` instead of `Record<string, any>`
- No runtime checks (type-safe by design)
- Supports all `PropertyKey` types (string, number, symbol)
- Follows codebase conventions (indexed for loop)
- Zero dependencies
**Type signature:**

```typescript
function fromPairs<K extends PropertyKey, V>(
  pairs: ReadonlyArray<readonly [K, V]>
): Record<K, V>
```

## Example Usage

```ts
import { fromPairs } from 'es-toolkit';

const pairs = [['a', 1], ['b', 2], ['c', 3]];
const result = fromPairs(pairs);
// => { a: 1, b: 2, c: 3 }

// Works with symbols
const sym = Symbol('key');
fromPairs([[sym, 'value']]);
// => { [sym]: 'value' }

// Type-safe
const typedPairs: Array<['x' | 'y', number]> = [['x', 1], ['y', 2]];
const typedResult = fromPairs(typedPairs);
// typedResult: Record<'x' | 'y', number>
```

### DIFF
--- a/src/function/throttle.spec.ts
+++ b/src/function/throttle.spec.ts
@@ -194,10 +194,10 @@ describe('throttle', () => {
       throttled();
     }, 10);
 
-    await delay(throttleMs * 3 + 10);
+    await delay(throttleMs * 5 + 10);
     clearInterval(intervalId);
 
-    expect(callback.mock.calls.length).toBeGreaterThanOrEqual(2);
+    expect(callback.mock.calls.length).toBeGreaterThanOrEqual(3);
   });
 
   it('should invoke function periodically with leading edge only', async () => {


raon0211:
Hey, thanks for the contribution! 

Note that`fromPairs` can be easily replaced with the built-in `Object.fromEntries`, which we mention in [our docs](https://es-toolkit.dev/reference/compat/object/fromPairs.html#frompairs-lodash-compatibility).

es-toolkit generally doesn't include functions that native JavaScript already handles well (see our [contributing guide](https://github.com/toss/es-toolkit/blob/main/.github/CONTRIBUTING.md#es-toolkit) for more details).

Thanks for thinking of us though—hope to see more contributions from you!

PR #1585, opened 2026-01-16
feat(retry): add shouldRetry option

dayongkr:
## Summary                                                                                                                          
- Add `shouldRetry` option to `retry` function                                                                                      
- Allows controlling retry behavior based on error type and attempt number                                                          
- Useful for retrying only on specific errors (e.g., 500+ status codes)                                                             
                                                                                                                                    
## Test plan                                                                                                                        
- [x] Unit test: retry when shouldRetry returns true                                                                                
- [x] Unit test: stop retrying when shouldRetry returns false                                                                       
- [x] Unit test: attempt number is correctly passed to shouldRetry

### DIFF
--- a/src/function/retry.ts
+++ b/src/function/retry.ts
@@ -38,6 +38,7 @@ interface RetryOptions {
 
 const DEFAULT_DELAY = 0;
 const DEFAULT_RETRIES = Number.POSITIVE_INFINITY;
+const DEFAULT_SHOULD_RETRY = () => true;
 
 /**
  * Retries a function that returns a promise until it resolves successfully.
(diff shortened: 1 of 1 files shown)


src/function/retry.ts
```diff
@@ -96,15 +110,18 @@ export async function retry<T>(func: () => Promise<T>, _options?: number | Retry
   let delay: number | ((attempts: number) => number);
   let retries: number;
   let signal: AbortSignal | undefined;
+  let shouldRetry: ((error: unknown, attempt: number) => boolean) | undefined;
```
raon0211:
```suggestion
  let shouldRetry: ((error: unknown, attempt: number) => boolean) = () => true;
```

src/function/retry.ts
```diff
@@ -96,15 +110,18 @@ export async function retry<T>(func: () => Promise<T>, _options?: number | Retry
...
 
   if (typeof _options === 'number') {
     delay = DEFAULT_DELAY;
     retries = _options;
     signal = undefined;
+    shouldRetry = undefined;
```
raon0211:
```suggestion
```

src/function/retry.ts
```diff
@@ -96,15 +110,18 @@ export async function retry<T>(func: () => Promise<T>, _options?: number | Retry
...
   } else {
     delay = _options?.delay ?? DEFAULT_DELAY;
     retries = _options?.retries ?? DEFAULT_RETRIES;
     signal = _options?.signal;
+    shouldRetry = _options?.shouldRetry;
```
raon0211:
```suggestion
    shouldRetry ??= _options.shouldRetry;
```

src/function/retry.ts
```diff
@@ -119,6 +136,11 @@ export async function retry<T>(func: () => Promise<T>, _options?: number | Retry
     } catch (err) {
       error = err;
 
+      // Determine if we should retry based on the provided shouldRetry function
+      if (shouldRetry && !shouldRetry(err, attempts)) {
```
raon0211:
```suggestion
      if (!shouldRetry(err, attempts)) {
```

PR #1587, opened 2026-01-17
fix(zipWith): correct type signature for combine function

### LINKED ISSUE #1586: type signature of combine function in `zipWith` doesn't include undefined
https://github.com/toss/es-toolkit/blob/50339c3245e9a8319dcec689ab3c0689becd088e/src/array/zipWith.ts#L92-L93

According to the document the `combine` function receives `undefined` if the provided arrays have different lengths.

https://github.com/toss/es-toolkit/blob/50339c3245e9a8319dcec689ab3c0689becd088e/src/array/zipWith.ts#L40-L44

It would be great to change the type signature of each item from `T` to `T | undefined` to match the actual behavior.

My main concern is that this change might introduce some burden to handle undefined for the majority of use cases where array lengths do match exactly.

Ju-MINJAE:
Closes #1586

## Description

This PR fixes a type definition mismatch in the `zipWith` function.

### Problem
The `zipWith` function fills missing elements with `undefined` when input arrays have different lengths (aligning with the length of the longest array). However, the `combine` callback's type signature previously assumed arguments were always defined (`T` and `U`). This discrepancy could lead to runtime errors if users attempted to access properties of `undefined` values without a type check.

### Changes
- Updated the type signature of the `combine` function to accept `T | undefined` and `U | undefined`.
- Added a new test case specifically to verify type handling for arrays of different lengths.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,7 +22,7 @@ commands:
 jobs:
   lint:
     docker:
-      - image: cimg/node:20.12
+      - image: cimg/node:24.13
     steps:
       - checkout
       - setup
@@ -50,7 +50,7 @@ jobs:
             fi
   typecheck:
     docker:
-      - image: cimg/node:20.12
+      - image: cimg/node:24.13
     steps:
       - checkout
       - setup
@@ -59,7 +59,7 @@ jobs:
           command: yarn tsc --noEmit
   pre-pack:
     docker:
-      - image: cimg/node:20.12
+      - image: cimg/node:24.13
     steps:
       - checkout
       - setup
@@ -74,7 +74,7 @@ jobs:
           command: yarn attw package.tgz
   check-peer:
     docker:
-      - image: cimg/node:20.12
+      - image: cimg/node:24.13
     steps:
       - checkout
       - setup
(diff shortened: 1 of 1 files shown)


Ju-MINJAE:
Is there a specific reason why only the test job uses Node 18.20.4?

https://github.com/toss/es-toolkit/blob/f365912434f1bbc9e805326a0c964e4cb5af4761/.circleci/config.yml#L22-L25
https://github.com/toss/es-toolkit/blob/f365912434f1bbc9e805326a0c964e4cb5af4761/.circleci/config.yml#L51-L53
https://github.com/toss/es-toolkit/blob/f365912434f1bbc9e805326a0c964e4cb5af4761/.circleci/config.yml#L84-L86

If it needs change, I'd be happy to update the version to v20.12 in this PR.

raon0211:
es-toolkit targets Node.js LTS versions. Node.js 18 recently left LTS, so we've dropped support for it. Thanks!

raon0211:

Technically, the callback in `zipWith` could receive `undefined` if array lengths differ. But in practice, being this strict tends to create a lot of noise in everyday code. TypeScript itself makes a similar tradeoff—without `noUncheckedIndexedAccess` (which is off by default), it lets you write:

```typescript
const arr = [1, 2, 3];
const element = arr[4]; // TypeScript infers `number`, not `number | undefined`
```

It's a tough call, but es-toolkit follows TypeScript's default strict mode behavior, so we'll pass on this one. Appreciate the suggestion though!

PR #1589, opened 2026-01-20
chore(package.json): remove '@types/eslint' as ESLint v9 provides built-in types

sukvvon:
## Summary
- Remove `@types/eslint` devDependency since ESLint v9 provides built-in TypeScript types

## Reference
- https://eslint.org/blog/2024/09/eslint-v9.10.0-released/

## Test plan
- [x] Verified ESLint config still works without `@types/eslint`

---

Thanks to @lumirlumir for the advice!


lumirlumir (commented):
Awesome 🙌

Just nitpicking another point: `eslint` and `@eslint/js` are always released in sync, so I think matching these versions would be nice:

https://github.com/toss/es-toolkit/blob/859355b9ee3c51b5e1cac4d0f5f338d879dbf32b/package.json#L67

https://github.com/toss/es-toolkit/blob/859355b9ee3c51b5e1cac4d0f5f338d879dbf32b/package.json#L82

```diff
- "@eslint/js": "^9.9.0", 
+ "@eslint/js": "^9.39.2",
```

```diff
-  "eslint": "^9.22.0", 
+  "eslint": "^9.39.2", 
```

---

The CI failure doesn't seem to be related to this change. Other PRs are experiencing the same issue.

PR #1590, opened 2026-01-23
chore(eslint.config): use 'globalIgnores' for explicit global ignore patterns

sukvvon:
## Summary

- Use `globalIgnores` helper function from `eslint/config` instead of plain `ignores` object
- Makes the intent of global ignore patterns explicit and improves code readability

## Notes

- `globalIgnores()` is a helper introduced in ESLint v9 that explicitly marks ignore patterns as global
- Functionally equivalent to the previous approach, but clearer in intent

## Reference

- https://eslint.org/docs/latest/use/configure/configuration-files#globally-ignoring-files-with-ignores

Thanks to @lumirlumir for the advice!


eslint.config.mjs
```diff
@@ -1,25 +1,23 @@
...
-  },
+  globalIgnores([
+    '.yarn/**',
+    'coverage/**',
+    '**/dist/**',
+    '**/cache/**',
+    '.pnp.*',
+    '**/*.d.ts',
+    '**/*.tgz',
+    'node_modules/**',
```
lumirlumir:
```suggestion
```

Nitpicking: `node_modules` directory is ignored by default so it seems a bit redundant. 😄 

Ref: https://github.com/eslint/eslint/blob/main/lib/config/default-config.js#L68

lumirlumir (approved):
Lovely ❤️‍🔥

raon0211 (approved):

PR #1597, opened 2026-01-23
fix: added missing packages to publishConfig

### LINKED ISSUE #1596: Map package is not exported in package.json
I was trying to use the mapValues function from https://es-toolkit.dev/reference/map/mapValues.html and to import it as it's stated in the documentation using `import { mapValues } from 'es-toolkit/map';`.
This fails.
When I checked why, it seems to me that /map is not exported in the package.json of es-toolkit.

Version: 1.44.0

matthiasfeist:
fixes #1596


matthiasfeist:
@dayongkr or @raon0211 any more comments? If not, it would be cool to get this merged because otherwise all the cool new methods in your map and set package can't be used by end-users of es-toolkit.

Issue #1605, opened 2026-02-01
docs: Localize "Appearance" label in mobile navigation for i18n support


kss2002:
### Description

On smaller screen widths (responsive view), the dark mode toggle label always displays "Appearance" regardless of the selected language. This should be localized to match the user's selected language.

### Current Behavior

The "Appearance" text in the mobile navigation menu remains in English for all language settings (Korean, Japanese, Chinese).

### Expected Behavior

The label should display the appropriate translation based on the selected locale:
- English: `Appearance`
- Korean (한국어): `테마`
- Japanese (日本語): `テーマ`
- Chinese (简体中文): `主題`

### Proposed Solution

Add the `darkModeSwitchLabel` property to each locale's `themeConfig` in the VitePress configuration files:
- [docs/.vitepress/en.mts](https://github.com/kss2002/es-toolkit/blob/main/docs/.vitepress/en.mts)
- [docs/.vitepress/ko.mts](https://github.com/kss2002/es-toolkit/blob/main/docs/.vitepress/ko.mts)
- [docs/.vitepress/ja.mts](https://github.com/kss2002/es-toolkit/blob/main/docs/.vitepress/ja.mts)
- [docs/.vitepress/zh_hans.mts](https://github.com/kss2002/es-toolkit/blob/main/docs/.vitepress/zh_hans.mts)

### Environment

- VitePress documentation site
- Responsive/mobile view

### img

<img width="554" height="265" alt="Image" src="https://github.com/user-attachments/assets/e59a5bd5-a10d-49e7-9198-cebc822c73da" />

Archibald1948:
Since it has been merged, close this issue as well.

PR #1606, opened 2026-02-01
docs: localize darkModeSwitchLabel for i18n support

kss2002:
## Summary

This PR adds localized `darkModeSwitchLabel` translations to the VitePress theme configuration, so the "Appearance" toggle label displays in the user's selected language on responsive/mobile views.

## issues

- https://github.com/toss/es-toolkit/issues/1605

## Changes

Added `darkModeSwitchLabel` property to `themeConfig` in each locale configuration file:

| Locale | File | Label |
|--------|------|-------|
| English | [en.mts](https://github.com/toss/es-toolkit/blob/main/docs/.vitepress/en.mts) | `Appearance` |
| Korean | [ko.mts](https://github.com/toss/es-toolkit/blob/main/docs/.vitepress/ko.mts) | `테마` |
| Japanese | [ja.mts](https://github.com/toss/es-toolkit/blob/main/docs/.vitepress/ja.mts) | `テーマ` |
| Chinese | [zh_hans.mts](https://github.com/toss/es-toolkit/blob/main/docs/.vitepress/zh_hans.mts) | `主題` |

## Before

The dark mode toggle label always showed "Appearance" in English regardless of the selected language.

## After

The label now displays the appropriate translation based on the selected locale.

- img
toss local

### DIFF
--- a/docs/.vitepress/en.mts
+++ b/docs/.vitepress/en.mts
@@ -11,7 +11,7 @@ export const en = defineConfig({
     'A state-of-the-art, high-performance JavaScript utility library with a small bundle size and strong type annotations.',
 
   themeConfig: {
-    darkModeSwitchLabel: 'Appearance',
+    darkModeSwitchLabel: 'Dark Mode',
 
     nav: nav(),
 
--- a/docs/.vitepress/ja.mts
+++ b/docs/.vitepress/ja.mts
@@ -10,7 +10,7 @@ export const ja = defineConfig({
   description: '高速なパフォーマンスと小さなバンドルサイズを持つ最先端のJavaScriptユーティリティライブラリ',
 
   themeConfig: {
-    darkModeSwitchLabel: 'テーマ',
+    darkModeSwitchLabel: 'ダークモード',
 
     nav: nav(),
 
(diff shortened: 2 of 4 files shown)


Archibald1948:
Please review the code 🙏

Archibald1948:
> I don’t understand what it means to turn a appearance on and off — how about just labeling it as dark mode?

Since toggle actually switches between two states (Light/Dark), I think it's misleading to mark it as "dark mode." 

"VitePress defaults" contain all three states, and the translations we are used localized versions that correspond to "theme/look" that more accurately describe the entire function. 

It may be an unnatural result to say that the value is "theme" because the variable name of themeConfig is darkModeSwitchLabel.

However, it was forced to use these variable names because of "VitePress defaults".

If you have any better directions, please let me know! I'll reflect it in the code.

thank you :)

dayongkr:
You’re right that a toggle ultimately switches between two states, but like a switch, it’s commonly used as an on/off control. As the setting name darkModeSwitchLabel suggests, framing it as turning dark mode on and off is more intuitive.

It’s not a major issue, but “Appearance” is a broader term, which means users have to check the label once and the switch icon once — two steps to understand the function. That’s a minor UX concern.​​​​​​​​​​​​​​​​

Archibald1948:
Thank you for the review.
I have reflected it in the code, thank you. :)

PR #1610, opened 2026-02-10
add early return empty array

clicktodev:
Adds an early return when the input array is empty to avoid unnecessary Math.ceil calculation and array allocation.

## Changes

- Added empty array check after size validation in `src/array/chunk.ts`

```typescript
if (arr.length === 0) {
  return [];
}
```

The existing test coverage already validates this behavior.

### DIFF
--- a/src/array/chunk.ts
+++ b/src/array/chunk.ts
@@ -26,6 +26,10 @@ export function chunk<T>(arr: readonly T[], size: number): T[][] {
     throw new Error('Size must be an integer greater than zero.');
   }
 
+  if (arr.length === 0) {
+    return [];
+  }
+
   const chunkLength = Math.ceil(arr.length / size);
   const result: T[][] = Array(chunkLength);
 


dayongkr:
Thanks for the PR\! Could you share benchmark results showing the performance improvement? For example, a comparison of before/after with empty arrays (and perhaps also with non-empty arrays to confirm there's no regression) would be helpful.

raon0211:
Thanks for the PR, and for the performance-minded thinking! 

I measured the trade-off. The early return only helps when the input is empty (around 1 ns/call), but it adds +11 B gzip (+5.3%) to `chunk` for every consumer. Since `chunk` is almost always called on non-empty arrays, and the current code already returns `[]` correctly, the cost lands on everyone while the benefit is limited to an edge case.

Keeping the smallest possible footprint is a core goal for es-toolkit, so we'd prefer to keep the current version. Really appreciate the contribution though, please keep them coming! 🙏

PR #1612, opened 2026-02-11
fix: add missing map and set exports to publishConfig

eunwoo-levi:
## Summary

- Add `./map` and `./set` subpath exports to `publishConfig.exports` in `package.json`
- Update test expectations in `check-dist.spec.ts` to include the new entrypoints
- These were already defined in the top-level `exports` field but missing from `publishConfig`, causing `import { mapValues } from 'es-toolkit/map'` (and `es-toolkit/set`) to fail after npm publish

## Root Cause

When `map` and `set` modules were added, the top-level `exports` and `files` fields were updated but `publishConfig.exports` was not synchronized, and the test file expectations were not updated accordingly.

## Changes

1. **package.json**: Added `./map` and `./set` to `publishConfig.exports`
2. **tests/check-dist.spec.ts**: Added `./map` and `./set` to `ENTRYPOINTS` array to fix failing CI tests


matthiasfeist:
Duplicate of https://github.com/toss/es-toolkit/pull/1597

PR #1615, opened 2026-02-11
fix(trimStart): add missing multi-character string validation

eunwoo-levi:
## Problem
The JSDoc states that `chars` should be a single character string, but there was no runtime validation to enforce this. 
As a result, `trimStart` accepts multi-character strings like `"--"` without throwing an error, causing undefined behavior.

## Solution

Added validation to throw an error when a multi-character string is passed.

```ts
trimStart("---hello", "--");  // Now throws error 
trimStart("---hello", "-");   // Works as expected 
```


raon0211 (approved):
Thanks!

PR #1617, opened 2026-02-14
feat(promise): add AbortSignal support to timeout and withTimeout

MyraxByte:
## Description

This PR adds **AbortSignal support** to the `timeout` and `withTimeout` utilities in `es-toolkit` so that timeouts can be cancelled via `AbortController`.

### Changes

- Extend `timeout(ms)` to accept an optional `AbortSignal` via `TimeoutOptions` and pass it through to `delay`
- Extend `withTimeout(run, ms)` to accept an optional `AbortSignal` via `WithTimeoutOptions` and pass it through to `timeout`
- Update JSDoc examples to demonstrate usage with `AbortController` and aborted operations
- Add tests to cover:
  - Rejecting with `AbortError` when the operation is aborted via `AbortSignal`
  - Immediate rejection when the signal is already aborted

### Motivation

Some consumers need to cancel pending timeouts when the surrounding operation is cancelled (for example, when an HTTP request is aborted). By wiring `AbortSignal` support through `timeout` and `withTimeout`, we enable safer cancellation semantics and better resource cleanup without changing existing call sites.

### Breaking Changes

None. Existing calls without `options` continue to work as before, and Abort-related behavior is opt‑in via the `signal` option.


raon0211:
Thanks for the PR!

The approach here wires `signal` through `delay`, so aborting makes `timeout` reject with `AbortError`. The problem is that this inverts what cancelling a timeout should mean. `timeout` only exists to lose a `Promise.race`, so aborting it should stop it from firing, not make it reject faster. And in `withTimeout` the race rejects while `run` keeps running, because the signal never reaches `run`, so nothing is actually freed. The cause is reusing `delay`'s reject-on-abort, which is right for `delay` (it is awaited directly) but wrong for `timeout` (it is only raced).

So I pushed a commit that changes it: on abort, `timeout` now clears its timer and stays pending, and `withTimeout` forwards the signal so aborting lifts the time limit. This does not leak, since the timer and listener are cleaned up and the promise is GC'd once the race settles. I also added the docs in all four languages.

Thanks again for contributing, and please keep it coming next time too!

raon0211 (dismissed):

raon0211 (approved):

PR #1621, opened 2026-02-17
fix(isDeepKey): improve deep key detection to match strict patterns

ramong26:
## What

- Improved the `isDeepKey`function to strictly detect only valid deep key patterns.
- Now returns false for invalid patterns such as 'a.', '.a', 'a[b', 'a]b]', and 'a][b]'.
- Returns true only for keys that match valid deep key patterns (e.g., 'a.b', 'a[b]', 'a.b.c', 'a[b][c]').
- Ensured `isKey`uses the same consistent deep property regex for property path detection.

## Why

- The previous implementation returned true for some invalid patterns, which could lead to incorrect behavior.
- This change ensures that only syntactically valid deep keys are detected, matching the documented examples and improving reliability.

## Test

- Added and updated unit tests in `isDeepKey.spec.ts` to cover both valid and invalid deep key patterns.
- Verified that `isKey` and `isDeepKey`are consistent in their detection logic.


raon0211 (dismissed):
Thanks!

ramong26:
@raon0211  Could you please push a new commit and approve the PR again?

raon0211 (approved):

PR #1626, opened 2026-02-24
fix(set): remove duplicate arg from set utils

mirorauhala:
All Set utilities pass a duplicate value argument to the predicate.

For example:
```ts
find(new Set([1,2,3]), (value, value2, originalSet) => ...) // value and value2 are the same
```

This seems rather unnecessary and the reasoning for this is not documented properly. Moreover, the `value2` property will need to be mentioned if you want to access the `originalSet`.

I'm proposing that we remove the duplicate argument `value2` from all Set utils. This change also updates their related tests and documentation.

This is a breaking change.

### DIFF
--- a/docs/reference/set/countBy.md
+++ b/docs/reference/set/countBy.md
@@ -63,7 +63,7 @@ const ageGroups = countBy(ages, age => {
 #### Parameters
 
 - `set` (`Set<T>`): The Set to count occurrences from.
-- `mapper` (`(value: T, value2: T, set: Set<T>) => K`): The function to produce a key for counting.
+- `mapper` (`(value: T, set: Set<T>) => K`): The function to produce a key for counting.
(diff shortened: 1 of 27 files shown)


raon0211:
Hello, thanks for your suggestion. Note that this intentionally mirrors `Set.prototype.forEach` ([#](https://developer.mozilla.org/ko/docs/Web/JavaScript/Reference/Global_Objects/Set/forEach)), which the spec defines as `(value, value, set)`. In a Set the element is its own key, so the signature lines up with `Map.forEach` (`value, key, map`) and `Array.forEach` (`value, index, array`).

mirorauhala:
Oh, thanks for letting me know. I wasn't aware of the spec containing the value argument twice. Makes total sense to mirror it.

PR #1633, opened 2026-03-02
fix[zip]: include undefined in tuple return types for arrays of different lenghts

### LINKED ISSUE #1609: Incorrect type for `zip` function when arrays have different sizes
Consider the call:

```ts
import { zip } from 'es-toolkit';
const result = zip([1, 2, 3], ['a', 'b']);
// [ [ 1, 'a' ], [ 2, 'b' ], [ 3, undefined ] ]
```

If I check the type of `result`, it becomes `[number, string][]`, instead of `[number | undefined, string | undefined][]`. In essence, if the arrays don't have the same size, the resulting entries may include undefined values, but the type of the corresponding entries does not include the `undefined` that is used as a filler.

rajadubey:
# Summary

This PR improves type safety for the `zip` function by adding `| undefined` to all tuple element positions in the overload return types. When arrays of different lengths are zipped, the resulting tuples contain `undefined` for missing elements, but the types did not reflect this.

# Changes

- **[`src/array/zip.ts`](src/array/zip.ts)**
  - Updated all 4 overload return types to include `| undefined` in each tuple position
  - Runtime implementation remains unchanged

- **[`src/array/zip.spec.ts`](src/array/zip.spec.ts)**
  - Added type-level test cases using `expectTypeOf` for 2-arg, 3-arg, and 4-arg overloads

# Motivation

- Using the current types causes loss of type information and can hide obvious runtime errors.

```ts
// zip with arrays of different lengths
const result = zip([1, 2, 3], ['a', 'b']);
// result at runtime: [[1, 'a'], [2, 'b'], [3, undefined]]

// TypeScript infers result as [number, string][]
// No TypeScript error, but accessing result[2][1] could be undefined
result[2][1].toUpperCase();
// No TypeScript error, but will crash at runtime because result[2][1] is undefined
```

- The compat version (`src/compat/array/zip.ts`) already correctly uses `| undefined`, so the core version should be consistent.

```ts
// With the fix, TypeScript correctly infers:
const result = zip([1, 2, 3], ['a', 'b']);
// Type: [number | undefined, string | undefined][]

result[2][1]?.toUpperCase(); // TypeScript now requires optional chaining
```

By including `| undefined` in the return types, the function now preserves type information and provides better safety for downstream consumers.

This change aligns with the project's design principles:

- Strong typing
- Clear and predictable API
- Consistency between `es-toolkit` and `es-toolkit/compat`

# Example

```ts
// 2 arrays of different lengths
const a = zip([1, 2, 3], ['a', 'b']);
// Type: Array<[number | undefined, string | undefined]>

// 3 arrays
const b = zip([1], [true], ['x']);
// Type: Array<[number | undefined, boolean | undefined, string | undefined]>

// 4 arrays
const c = zip([1], ['a'], [true], [null]);
// Type: Array<[number | undefined, string | undefined, boolean | undefined, null | undefined]>
```

Fixes #1609

### DIFF
--- a/src/array/zip.spec.ts
+++ b/src/array/zip.spec.ts
@@ -1,4 +1,4 @@
-import { describe, expect, it } from 'vitest';
+import { describe, expect, expectTypeOf, it } from 'vitest';
 import { zip } from './zip';
 
 describe('zip', () => {
@@ -26,5 +26,18 @@ describe('zip', () => {
 
   it('supports spread operators', () => {
     expect(zip(...[[1], ['s'], [{ a: 2 }]])).toEqual([[1, 's', { a: 2 }]]);
+  });
+
+  it('should include undefined in return types for arrays of different lengths', () => {
+    const result2 = zip([1, 2, 3], ['a', 'b']);
+    expectTypeOf(result2).toEqualTypeOf<Array<[number | undefined, string | undefined]>>();
+
+    const result3 = zip([1], [true], ['x']);
+    expectTypeOf(result3).toEqualTypeOf<Array<[number | undefined, boolean | undefined, string | undefined]>>();
+
+    const result4 = zip([1], ['a'], [true], [null]);
+    expectTypeOf(result4).toEqualTypeOf<
+      Array<[number | undefined, string | undefined, boolean | undefined, null | undefined]>
+    >();
   });
 });
--- a/src/array/zip.ts
+++ b/src/array/zip.ts
@@ -15,7 +15,7 @@
  * const result = zip(arr1);
  * // result will be [[1], [2], [3]]
  */
-export function zip<T>(arr1: readonly T[]): Array<[T]>;
+export function zip<T>(arr1: readonly T[]): Array<[T | undefined]>;
 
 /**
  * Combines multiple arrays into a single array of tuples.
@@ -36,7 +36,7 @@ export function zip<T>(arr1: readonly T[]): Array<[T]>;
  * const result = zip(arr1, arr2);
  * // result will be [[1, 'a'], [2, 'b'], [3, 'c']]
  */
-export function zip<T, U>(arr1: readonly T[], arr2: readonly U[]): Array<[T, U]>;
+export function zip<T, U>(arr1: readonly T[], arr2: readonly U[]): Array<[T | undefined, U | undefined]>;
 
 /**
  * Combines multiple arrays into a single array of tuples.
@@ -59,7 +59,11 @@ export function zip<T, U>(arr1: readonly T[], arr2: readonly U[]): Array<[T, U]>
  * const result = zip(arr1, arr2, arr3);
  * // result will be [[1, 'a', true], [2, 'b', false], [3, 'c', undefined]]
  */
-export function zip<T, U, V>(arr1: readonly T[], arr2: readonly U[], arr3: readonly V[]): Array<[T, U, V]>;
+export function zip<T, U, V>(
+  arr1: readonly T[],
+  arr2: readonly U[],
+  arr3: readonly V[]
+): Array<[T | undefined, U | undefined, V | undefined]>;
 
 /**
  * Combines multiple arrays into a single array of tuples.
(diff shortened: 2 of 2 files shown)


raon0211:
Technically, `zip` could return `undefined` if the arrays have different lengths—but being that strict tends to add a lot of noise to everyday code.

TypeScript makes the same kind of tradeoff: without `noUncheckedIndexedAccess` (which is off by default, even in strict mode), this is perfectly valid:

```typescript
const arr = [1, 2, 3];
const element = arr[4]; // inferred as `number`, not `number | undefined`
```

It's a judgment call, but we've opted to follow TypeScript's default behavior here. Thanks for the suggestion!

ryb73:
Hi @raon0211 – thanks for your work on this project, it's much appreciated!

I'm curious if you'd be open to a solution that uses type inference to respect the developer's `noUncheckedIndexedAccess` flag?

```ts
function peek<T>(arr: T[]) {
  return arr[0];
}

type PeekType<T> = ReturnType<typeof peek<T>>;
type PeekNumber = PeekType<number>;
//   ^? = `number | undefined` if `noUncheckedIndexedAccess` == true; otherwise `number`

const zipped = zip([1, 2, 3], [4, 5]);
//    ^? = `[number | undefined, number | undefined][]` if `noUncheckedIndexedAccess` == true; otherwise `[number, number][]`



export function zip<T, U>(arr1: readonly T[], arr2: readonly U[]): Array<[PeekType<T>, PeekType<U>]>;

export function zip<T>(...arrs: Array<readonly T[]>): T[][] {
  // For performance reasons, use this implementation instead of
  // const rowCount = Math.max(...arrs.map(x => x.length));
  let rowCount = 0;

  for (let i = 0; i < arrs.length; i++) {
    if (arrs[i].length > rowCount) {
      rowCount = arrs[i].length;
    }
  }
  const columnCount = arrs.length;
  const result = Array(rowCount);

  for (let i = 0; i < rowCount; ++i) {
    const row = Array(columnCount);
    for (let j = 0; j < columnCount; ++j) {
      row[j] = arrs[j][i];
    }
    result[i] = row;
  }
  return result;
}
```

https://www.typescriptlang.org/play/?noUncheckedIndexedAccess=true#code/MYewdgzgLgBAXgSwA5IKYBMYF57IBQDaAjADQwBMZAzALpkEAsZArDQJQDcAUAPQ8wCYAPQD8XLgDMArmGBQE4GGlQBrADwAVAHx4AhgCd9ALhgaC7GAG8uA-aihT9YGAf0EADDW4BfcVACeaDAACqiqGoGomlrYMABK9o5gEWhqAWggEkph6tpa3OmoITkAclIAtgBGqPqxoeGRamAV1fr5vPwCouKoAB5IIPqw0rLyiohImmQAqjquRCZ2uujgADb+puZkruSLqMtrG9PmbCYAgoa6-moE9SopUdpkdw9qszTtXH0DQzAjcgpnBNongAHTg1wQc6Xa5LFZgdabD6nJHmKw2GB8GAAMUG2X0EkG5V0siKSwg4AgZCkECKUAAFggIDAEOUkKtUOVUGAoLoxs4EJAoPtMJkMVjQEKYPoQAB3ADCIBksBwAFk+fTQcTemCIYYIFrdEg8L1sDFeqCOWAAOYMticDEc2AyhVKnmxdzcDGE2p4J0sj0cANqFz6y3c230oMIADUMbY6MELKyen1BAQNHDNoZMBiLsVyoT1iTtjlBfdOEh6czVsj3CTvgEjZgkugLZAqwqYHLKtD+gNtYZ9fbUrsECkq17F30Vzw+bdUAd3rxfvsAZwnuD0rLC6DcYQRYxAlbzrlsWns9AnfK3YXDqTPpgq9gACtAzA3yGr12e3uYy-DxLbdZQIF8aFiKswOrYcmyPaVUHHSdq1iF1h2bOwHCceDEKgHwuCAA

Personally I use TypeScript, and a strict tsconfig, because I like these types of coding errors to appear at compile time rather than runtime. I'm happy to pay the cost of slightly noisier code if it prevents surprising runtime behavior. Hopefully this solution gives us the best of both worlds, by only adding the noise if the developer has opted into `noUncheckedIndexedAccess`.

I totally understand if you'd prefer not to add this complexity, but if you're open to it I'd be happy to open a PR. :)

PR #1634, opened 2026-03-04
fix(pick): return Partial<Pick<T, K>> for non-tuple key arrays

### LINKED ISSUE #1623: pick: pick should return Partial<Pick<T,K>>
> tl;dr: `pick` should return `Partial<Pick<T,K>>`. It might be possible to specialize to `Pick<T,K>` for certain types `K`. Would you be willing to accept a contribution?

Consider this code:

```ts
const t = { a: 1, b: 2, c: 3 };
type T = typeof t;
const k = [Math.random % 2 === 0 ? 'a' : 'b'] as const; // readonly ['a' | 'b']
const res = pick(t, k); // Pick<T, 'a' | 'b'>
// but res === { a: 1 } or { b: 2 } <-> Pick<T, 'a'> | Pick<T, 'b'>
```

The easy solution to fix this would be to return `Partial<Pick<T, K>>` from `pick`. However, that's annoying for situations where we know that all `T`s are going to be in `K`, like: `pick(x, ['a', 'b'])`. In this case we want `Pick<X, 'a' | 'b'>`.

The fix in TS seems to be complicated though. I tried a little bit and got tuple detection to work. I could prepare a PR that would fix the type for general arrays to `Partial<Pick<T, K>>` and to `Pick<T, K>` if the user passes a tuple. That would deblock a common case, while still being wrong in the example I gave.

What do you think? Would you be willing to accept such a contribution? Do you prefer the status quo?

Seojooyeon-creat:
Closes #1623

## Summary

- `pick` currently returns `Pick<T, K>` regardless of whether `keys` is a tuple or a variable-length array. This is incorrect when `K` is a union type: `Pick<T, 'a' | 'b'>` implies both `a` and `b` are always present, but the runtime value may only contain one of them.
- This PR fixes the return type using a conditional type that distinguishes between tuple and variable-length array inputs.

## Changes

**`src/object/pick.ts`**

Changed the signature from:
```ts
pick<T extends Record<string, any>, K extends keyof T>(obj: T, keys: readonly K[]): Pick<T, K>
```
to:
```ts
pick<T extends Record<string, any>, Keys extends ReadonlyArray<keyof T>>(
  obj: T,
  keys: readonly [...Keys]
): number extends Keys['length'] ? Partial<Pick<T, Keys[number]>> : Pick<T, Keys[number]>
```

- `readonly [...Keys]` forces TypeScript to infer tuple types for array literals like `['a', 'b']`
- `number extends Keys['length']` distinguishes tuple (length is a specific literal) from variable-length array (length is `number`)
- Result: literal tuple → `Pick<T, K>`, variable-length array → `Partial<Pick<T, K>>`

**`src/object/pick.spec.ts`**

Added two type-level tests using `expectTypeOf`:
- Literal tuple `['a', 'b']` → `Pick<T, 'a' | 'b'>` (all keys guaranteed)
- Variable array `('a' | 'b')[]` → `Partial<Pick<T, 'a' | 'b'>>` (only present keys)

**`docs/`** (en, ko, ja, zh_hans)

Updated parameter and return type descriptions to reflect the new behavior.

## Context

A similar fix was previously applied to `omit` (#1498) using overloads and then reverted (#1595). This PR takes a different approach — a **single signature with a conditional return type** — which avoids the added complexity of overloads while correctly narrowing the type based on tuple vs. array input.

## Benchmark

Runtime logic is unchanged. Performance is maintained:

```
es-toolkit/pick    19,473,619 hz  (4.40x faster than lodash)
```


raon0211:
Thank you so much for taking the time to work on this — the analysis in the PR description is thorough, and we really appreciate the effort, including the docs updates across all locales.

That said, after careful consideration, we've decided not to move forward with this change. As outlined in our [CONTRIBUTING.md](https://github.com/toss/es-toolkit/blob/main/CONTRIBUTING.md), es-toolkit aims for the simplest possible implementation that supports the major use cases. Conditional return types that branch on tuple vs. variable-length arrays add a layer of type complexity that goes against that principle — this is also why a similar change to `omit` (#1498) was reverted in #1595. For the vast majority of users, the current `Pick<T, K>` signature is simple and predictable, and we'd like to keep it that way.

So I'll close this PR, but please don't be discouraged — we'd love to see more contributions from you in the future. Thanks again! 🙏

Seojooyeon-creat:
Thank you for check my PR!

PR #1637, opened 2026-03-05
feat(chain, zipIterable, enumerate): add lazy iterator utilities

### LINKED ISSUE #771: feature: iterator chain / concatenation
Python calls it [`itertools.chain`](https://docs.python.org/3/library/itertools.html#itertools.chain):

```js
chain("ABC", [1, 2, 3]).toArray() === ["A", "B", "C", 1, 2, 3]
```
Like `Array.concat`, but lazy, and doesn't need its iterables to be marked with [`Symbol.isConcatSpreadable`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/isConcatSpreadable).

(does the toolkit not have an Iterator Utilities section yet?)

y0ngha:
## Summary

Adds three lazy iterator utilities under the new `es-toolkit/iterator` module.

- `chain(...iterables)`. Combines multiple iterables sequentially without allocating a new array
- `zipIterable(...iterables)`. Pairs elements at the same index from multiple iterables
- `enumerate(iterable, start?)`. Pairs each element with its index, works with any iterable

These functions complement the existing `es-toolkit/array` utilities by supporting the full iterable protocol (`Set`, `Map`, generator functions, custom iterables), not just arrays.

Closes #771

---

### Why a new `iterator` module?

The new `iterator/*` functions target a different use case:

| | `array/zip` | `iterator/zipIterable` |
| --- | --- | --- |
| Input | `Array` only | Any `Iterable` |
| Output | `Array` | `IterableIterator` (lazy) |
| Length mismatch | Pads with `undefined` | Stops at shortest |
| Memory | Allocates full array immediately | Consumes on demand |

### Why lazy?

```ts
// allocates a new array with 30,000 elements in memory
const merged = [...arr1, ...arr2, ...arr3]

// chain is lazy. no intermediate array is allocated
for (const x of chain(arr1, arr2, arr3)) {
  // consume only what you need
}
```

Note that `IterableIterator` does not have `.filter()` or `.map()` methods on its own.
For lazy composition, you have two options:

Option 1. ES2025 Iterator Helpers (Node.js 22+ / modern browsers)

```ts
// Iterator.prototype.filter and .map are available natively
chain(arr1, arr2, arr3)
  .filter(x => x > 0)
  .map(x => x * 2)
  .take(10)
```

Option 2. Custom lazy utilities

```ts
// implement or import lazy filter/map that accept Iterable
function* filter(fn: (x: T) => boolean, iter: Iterable) {
  for (const x of iter) {
    if (fn(x)) yield x;
  }
}

[...filter(x => x > 0, chain(arr1, arr2, arr3))]
```
I saw a PR that also implemented the pipe function, and it would be better if other functions (lazy filter, map, reduce, …) were added like that 😄

---

## API

```ts
import { chain, zipIterable, enumerate } from 'es-toolkit/iterator';

// chain. sequential iteration without memory allocation
[...chain([1, 2], [3, 4], [5, 6])] // [1, 2, 3, 4, 5, 6]

// zipIterable. pair elements at the same index
[...zipIterable([1, 2, 3], ['a', 'b', 'c'])] // [[1, 'a'], [2, 'b'], [3, 'c']]

// enumerate. index + value pairs for any iterable
[...enumerate(new Set(['x', 'y', 'z']))] // [[0, 'x'], [1, 'y'], [2, 'z']]
[...enumerate(['a', 'b', 'c'], 1)]       // [[1, 'a'], [2, 'b'], [3, 'c']]
```

---

## Tests

3 files, 19 tests, 0 type errors. 100% coverage across all three functions.

test-coverage
test-result

---

### chain

chain-benchmark

- 1.50x faster than itertools/ichain
- Native spread is faster for small arrays because it avoids iterator protocol overhead. For large arrays, `chain` avoids allocating a merged array in memory, making it the better choice when the full result isn't needed.

### zipIterable

zip-iterable-benchmark

- ~37x faster than @fxts/core/zip
- ~1.2x slower than itertools/izip due to generator suspension overhead vs plain object iterator internals

### enumerate

enumerate-benchmark

- Comparable to @fxts/core/zipWithIndex
- Native `forEach`/`entries` are faster because they are optimized at the V8 engine level. The value of `enumerate` is not performance but universal iterable support. it works with `Set`, `Map`, generator functions, and any custom iterable, unlike `Array.prototype.entries`.

---

## Note to Reviewers

This is my first contribution to es-toolkit, so please feel free to point out
anything that needs to be changed or improved.

One thing I'd like to ask: I haven't implemented these functions in
`es-toolkit/compat`. Since these are new functions rather than Lodash compatible
ones, I assumed compat was not necessary. but please let me know if that's wrong.

### DIFF
--- a/benchmarks/bundle-size/camelCase.spec.ts
+++ b/benchmarks/bundle-size/camelCase.spec.ts
@@ -14,6 +14,6 @@ describe('camelCase bundle size', () => {
 
   it('es-toolkit/compat', async () => {
     const bundleSize = await getBundleSize('es-toolkit/compat', 'camelCase');
-    expect(bundleSize).toMatchInlineSnapshot(`650`);
+    expect(bundleSize).toMatchInlineSnapshot(`1287`);
   });
 });
--- a/benchmarks/bundle-size/chain.spec.ts
+++ b/benchmarks/bundle-size/chain.spec.ts
@@ -1,0 +1,14 @@
+import { describe, expect, it } from 'vitest';
+import { getBundleSize } from './utils/getBundleSize';
+
+describe('chain bundle size', () => {
+  it('es-toolkit', async () => {
+    const bundleSize = await getBundleSize('es-toolkit/iterator', 'chain');
+    expect(bundleSize).toMatchInlineSnapshot(`82`);
+  });
+
+  it('itertools', async () => {
+    const bundleSize = await getBundleSize('itertools', 'chain');
+    expect(bundleSize).toMatchInlineSnapshot(`252`);
+  });
+});
--- a/benchmarks/bundle-size/clone.spec.ts
+++ b/benchmarks/bundle-size/clone.spec.ts
@@ -9,7 +9,7 @@ describe('clone bundle size', () => {
 
   it('es-toolkit', async () => {
     const bundleSize = await getBundleSize('es-toolkit', 'clone');
-    expect(bundleSize).toMatchInlineSnapshot(`866`);
+    expect(bundleSize).toMatchInlineSnapshot(`1001`);
   });
 
   it('es-toolkit/compat', async () => {
--- a/benchmarks/bundle-size/cloneDeep.spec.ts
+++ b/benchmarks/bundle-size/cloneDeep.spec.ts
@@ -9,11 +9,11 @@ describe('cloneDeep bundle size', () => {
 
   it('es-toolkit', async () => {
     const bundleSize = await getBundleSize('es-toolkit', 'cloneDeep');
-    expect(bundleSize).toMatchInlineSnapshot(`3171`);
+    expect(bundleSize).toMatchInlineSnapshot(`3200`);
   });
 
   it('es-toolkit/compat', async () => {
     const bundleSize = await getBundleSize('es-toolkit/compat', 'cloneDeep');
-    expect(bundleSize).toMatchInlineSnapshot(`3522`);
+    expect(bundleSize).toMatchInlineSnapshot(`3640`);
   });
 });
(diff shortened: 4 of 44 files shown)


raon0211:
Hi @y0ngha, thank you for the thorough contribution — tests, benchmarks, and docs in all four languages are much appreciated, especially for a first PR!

Unfortunately the landscape has changed since this PR was opened. In #1815 we shipped the `es-toolkit/iterator` module, which now covers most of what's proposed here:

- **`zipIterable`** — `es-toolkit/iterator` now ships a lazy `zip` with the same semantics (stops at the shortest source). It also closes every source via `return()` on early termination and returns an `IteratorObject`, so it composes with native iterator helpers like `.map()`, `.filter()`, and `.take()`.
- **`enumerate`** — this is expressible with the existing primitives: `zip(range(0, Infinity), iterable)` yields exactly the same `[index, value]` pairs, and a start offset is just `range(start, Infinity)`. Following our simplicity principle, we'd rather keep the primitives composable than add a dedicated function.
- **`chain`** — the TC39 [iterator sequencing proposal](https://github.com/tc39/proposal-iterator-sequencing) (`Iterator.concat`) advanced to Stage 3 in July 2025 and has since reached Stage 4. Our design principle is to not implement functions replaceable by Stage 3+ proposals, so this will soon be covered by the language itself.

Given all three functions are covered by the existing module, the platform, or composition, I'll close this PR. Sorry the timing didn't work out — this was a well-executed PR, and we'd be glad to see you contribute again. If you're interested, the `es-toolkit/iterator` module is still growing and there are open discussions on which helpers to add next.

PR #1645, opened 2026-03-15
fix(compat/isEmpty): handle functions with own enumerable properties

D-Sketon:
fix https://github.com/toss/es-toolkit/issues/1638


raon0211 (approved):
Thanks for your help!

PR #1647, opened 2026-03-15
fix(compat/isMatchWith): align nested primitive matching with lodash in isMatchWith

D-Sketon:
Keep root-level primitive-vs-object behavior, but use strict equality for nested comparisons. This fixes false positives in matches/isMatch/isMatchWith.


raon0211 (approved):
Thanks!

PR #1649, opened 2026-03-17
feat(median): add bigint support to median and medianBy

### LINKED ISSUE #1588: Feature request: support `bigint` in `median()`
In spirit of https://github.com/toss/es-toolkit/issues/742, I would like to request supporting `bigint` in `median()`.

Currently, `median()` is implemented with division by `2` (`number`), which would cause a runtime error for mixing `BigInt` with other types:
https://github.com/toss/es-toolkit/blob/859355b9ee3c51b5e1cac4d0f5f338d879dbf32b/src/math/median.ts#L32
This can be solved by checking `typeof` in runtime and using `2n` instead if dealing with `bigint`.

Gdm0714:
## Summary

- Add `bigint[]` overload to `median()` — uses `2n` for integer division instead of `2`
- Add `bigint[]` overload to `medianBy()` to match
- Use comparison-based sorting (`< >`) instead of subtraction (`a - b`) to support both `number` and `bigint`

## Motivation

`median()` currently throws a runtime error when called with `bigint[]` because it divides by `2` (a `number`), which cannot be mixed with `bigint`. This follows the same pattern as #742 (`sum()` bigint support).

Closes #1588

## Test plan

- [x] `median([1n, 2n, 3n, 4n, 5n])` returns `3n`
- [x] `median([1n, 2n, 3n, 4n])` returns `2n` (integer division)
- [x] Unsorted bigint arrays return correct median
- [x] Single-element bigint arrays work
- [x] `medianBy()` works with bigint getValue
- [x] All existing 68 math tests still pass

### DIFF
--- a/src/math/median.spec.ts
+++ b/src/math/median.spec.ts
@@ -37,4 +37,9 @@ describe('median', () => {
   it('returns the single element for bigint arrays with one element', () => {
     expect(median([42n])).toEqual(42n);
   });
+
+  it('returns NaN for empty bigint arrays', () => {
+    const empty: bigint[] = [];
+    expect(median(empty)).toEqual(NaN);
+  });
 });
--- a/src/math/median.spec.ts
+++ b/src/math/median.spec.ts
@@ -21,4 +21,20 @@ describe('median', () => {
   it('returns the single element for arrays with one element', () => {
     expect(median([42])).toEqual(42);
   });
+
+  it('returns the median for an odd number of bigint elements', () => {
+    expect(median([1n, 2n, 3n, 4n, 5n])).toEqual(3n);
+  });
+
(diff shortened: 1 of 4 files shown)


dayongkr:
Thanks for the PR and the thorough tests! After thinking about the direction, I'd rather not mix `bigint` handling into the existing `number`-oriented functions — overloading `median`/`medianBy` to branch on `bigint` adds complexity to the core path for a fairly niche case.

We're planning to provide `bigint` support through a separate `es-toolkit/bigint` entry point instead, so `number` and `bigint` stay cleanly separated. Given that, I'll close this one for now. I really appreciate the effort here — once the `es-toolkit/bigint` direction lands, contributions there would be very welcome 🙏

PR #1675, opened 2026-03-31
fix(partial): add missing overloads for fully applied arguments

park-siwan:
Without these overloads, calls like `partial(fn, a, b)` where all arguments are provided fell through to the generic fallback signature, returning `(...args: any[]) => R` instead of `() => R`.

Added overloads:
- (T1, T2) => R with 2 args applied → () => R
- (T1, T2, T3) => R with 2 args applied → (T3) => R
- (T1, T2, T3) => R with 3 args applied → () => R
- (T1, T2, T3, T4) => R with 4 args applied → () => R

Also added type-level tests using expectTypeOf to prevent regression.

## Problem

When all arguments are provided to `partial`, TypeScript falls through to the variadic fallback overload, resulting in incorrect return types.

**Before:**
```ts
const add = (a: number, b: number): number => a + b;
const added = partial(add, 1, 2);
// Type: (...args: any[]) => number ❌
```

**After:**
```ts
const added = partial(add, 1, 2);
// Type: () => number ✅
added(); // => 3
```

Closes #1667


park-siwan:
Closing this PR as #1684 was merged and resolves the same issue. Thanks for maintaining the project!

PR #1681, opened 2026-04-03
fix(_internal/compareValues): place null and undefined values last when sorting

### LINKED ISSUE #1213: sortBy orderBy leave nulls alone
`a < null` is false and so is `a >= null` for any `typeof a === 'number'`. This means that compareValues(a, null) returns 0 and keeps the position of that element when sorting/ordering.

sukvvon:
## Summary
- Add `null`/`undefined` handling to `compareValues` in `src/_internal/compareValues.ts`
- `null` and `undefined` values are placed at the end for ascending order, at the beginning for descending order
- `undefined` is placed after `null` (matching lodash behavior)
- Add regression tests for both `sortBy` and `orderBy`

Fixes #1213

## Test Plan
- [x] `sortBy` with `null`/`undefined` values places them at the end
- [x] `orderBy` ascending places `null`/`undefined` at the end
- [x] `orderBy` descending places `undefined` before `null` at the beginning
- [x] All existing sortBy/orderBy tests pass
- [x] Full test suite passes (4334 tests)


sukvvon:
@raon0211 @dayongkr Could you guys review this PR?

dayongkr:
Direction looks right 👍 I agreed with the approach but pushed a commit tweaking the implementation: compute the ascending result once and flip the sign for `desc`, and read the nullish order via a small `nullishRank` helper compared with `<`/`>` (stays within `-1 | 0 | 1`, no nested ternary). Your tests still pass — please take a look!

sukvvon:
Much cleaner with `nullishRank` 👍 Verified `a2c1b38c` against lodash — `sortBy`/`orderBy` match, tests + `tsc` pass. LGTM 🙏

PR #1684, opened 2026-04-06
fix: add missing full-application overloads for partial with 2/3/4 args

jaydeep-pipaliya:
## Summary

Fixes #1667

The `partial` function was missing type overloads for fully applying all arguments of a 2-, 3-, or 4-argument function.

**Root cause:** When passing all arguments (e.g. `partial(add, 1, 2)`), TypeScript would fall through to a variadic overload like `partial<TS extends any[], T1, T2, R>(func: (arg1: T1, arg2: T2, ...args: TS) => R, t1: T1, arg2: T2): (...args: TS) => R`, failing to infer `TS = []`. This resulted in a wrong return type of `(arg3: unknown, arg4: unknown) => number` instead of `() => number`.

**Before:**
```ts
const add = (a: number, b: number): number => a + b;
const addBoth = partial(add, 1, 2);
// Type: (arg3: unknown, arg4: unknown) => number  ❌
```

**After:**
```ts
const addBoth = partial(add, 1, 2);
// Type: () => number  ✅
addBoth(); // => 3
```

## Changes

- Added explicit full-application overloads for 2-, 3-, and 4-argument functions
- Placed before variadic fallback overloads so TypeScript picks the more specific signatures
- Added 3 tests covering all new overloads

### DIFF
--- a/src/function/partial.spec.ts
+++ b/src/function/partial.spec.ts
@@ -111,4 +111,25 @@ describe('partial', () => {
     expect(partialed(5)).toBe(15);
     expect(typeof partialed.prototype).toBe('object');
   });
+
+  it('fully applies a 2-argument function and returns a zero-argument function', () => {
+    const add = (a: number, b: number) => a + b;
+    const addThree = partial(add, 1, 2);
+
+    expect(addThree()).toBe(3);
+  });
+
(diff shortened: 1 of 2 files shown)


raon0211 (dismissed):
Thanks for catching this!

raon0211 (approved):
Thanks for catching this!

PR #1685, opened 2026-04-06
fix: export toCamelCaseKeys and toSnakeCaseKeys from compat for browser bundle

### LINKED ISSUE #1642: `toSnakeCaseKeys`  is missing from dist/browser.global.js bundle
While `es-toolkit` provides a browser-ready global bundle (`browser.global.js`), the recently added utility function `toSnakeCaseKeys` is not included in this bundle.Is this a compatibility issue or a package omission?

jaydeep-pipaliya:
## Summary

Fixes #1642

**Root cause:** The browser global bundle (`dist/browser.global.js`) is built from `src/compat/index.ts`. The `toSnakeCaseKeys` and `toCamelCaseKeys` functions were added to `src/object/` but never re-exported from `src/compat/compat.ts`, so they were absent from the browser bundle.

**Fix:** Re-export both functions from `src/compat/compat.ts`, which makes them available in:
- `dist/browser.global.js` (browser UMD bundle)
- `es-toolkit/compat` import path

Both functions are also re-exported alphabetically alongside the other `to*` object utilities.


src/compat/compat.ts
```diff
@@ -188,8 +188,10 @@ export { propertyOf } from './object/propertyOf.ts';
 export { result } from './object/result.ts';
 export { set } from './object/set.ts';
 export { setWith } from './object/setWith.ts';
+export { toCamelCaseKeys } from '../object/toCamelCaseKeys.ts';
```
raon0211:
```suggestion
```

src/compat/compat.ts
```diff
@@ -188,8 +188,10 @@ export { propertyOf } from './object/propertyOf.ts';
...
 export { toDefaulted } from './object/toDefaulted.ts';
 export { toPairs } from './object/toPairs.ts';
+export { toSnakeCaseKeys } from '../object/toSnakeCaseKeys.ts';
```
raon0211:
```suggestion
```

raon0211:
Hello, thanks for your pull request!

I'm not sure why the browser bundle was being built from our compat utility, but I think we should explicitly build it from a dedicated source file called `browser.ts` to make the intent clearer. Since we'll be re-exporting all symbols from our main utility, `toSnakeCaseKeys` and other functions will be included automatically.

Thanks again!

raon0211 (approved):

PR #1690, opened 2026-04-17
feat: add acronymize function to string utilities

arimotearipo:


raon0211:
Thanks for the PR! I don't think this one's a great fit for the library though — acronymizing strings is pretty niche, and it's easy enough to implement inline when needed. Appreciate you taking the time!

PR #1692, opened 2026-04-21
feat: add MergeDeep type and merge.deep/mergeWith.deep/toMerged.deep APIs

soilSpoon:
## Summary

Adds a new `MergeDeep<T, S>` type and `.deep()` namespace methods for `merge`, `mergeWith`, and `toMerged`,
so callers can opt into structurally merged result types without changing the default `T & S` return types.

This PR also tightens the deep-merge typing and aligns the merge family more closely:

- non-tuple array deep merges include source-only appended element types
- index signatures are deeply merged instead of being unioned
- `mergeWith.deep()` widens the result when the customizer can override values
- `mergeWith` now accepts top-level arrays like `merge` and `toMerged`

## Motivation

The existing `merge()` return type is `T & S`, which produces intersections like
`{ a: number } & { a: string }` instead of a usable merged shape.

Changing the default `merge()` return type directly risks inference drift on existing callers,
so this PR follows the namespace-method pattern (`merge.deep()`) instead of changing the
default APIs.

## Changes

- **New type:** `MergeDeep<T, S>` in `src/_internal/types/MergeDeep.ts`
  - deeply merges plain objects
  - preserves tuple lengths
  - deeply merges index signature values
  - includes source-only element types for non-tuple arrays
  - treats `null`, `undefined`, and built-ins correctly
- **New APIs:**
  - `merge.deep(target, source)` -> `MergeDeep<T, S>`
  - `mergeWith.deep(target, source, mergeFn)` -> deep-merge result widened to include customizer override values
  - `toMerged.deep(target, source)` -> `MergeDeep<T, S>`
- **Typing cleanup:**
  - removes newly introduced `any` usage in the touched merge type/runtime code
  - aligns `mergeWith` input types with `merge` / `toMerged` so top-level arrays are supported consistently
- **Runtime refactor:**
  - extracts shared recursive merge logic into `src/_internal/mergeInternal.ts`
  - localizes type assertions at the public API boundaries while keeping runtime behavior unchanged
- **Docs/export:**
  - adds `@example` blocks for the new `.deep()` methods
  - exports `MergeDeep` from `src/object/index.ts`

## No breaking changes

- existing `merge` / `mergeWith` / `toMerged` default behavior is preserved
- `.deep()` methods are additive opt-in APIs
- `compat/` modules are intentionally left untouched to preserve lodash type compatibility
- widening `mergeWith` inputs to top-level arrays is additive

## Before / After

**Before:**
```ts
const result = merge({ a: { x: 1 } }, { a: { y: '2' } });
//    ^? { a: { x: number } & { y: string } }  ❌
```

**After:**
```ts
const result = merge.deep({ a: { x: 1 } }, { a: { y: '2' } });
//    ^? { a: { x: number; y: string } }  ✅
```

## Test plan

- [x] Type tests added for `MergeDeep` (objects, arrays, tuples, null, undefined, built-ins, index signatures)
- [x] Type tests added for source-only non-tuple array elements and deeply merged index signatures
- [x] Type tests added for `merge.deep`, `mergeWith.deep`, `toMerged.deep`, including top-level array inputs
- [x] Type tests added for `mergeWith.deep` customizer override widening
- [x] Runtime tests added for mixed array/object behavior and cloning semantics
- [x] `yarn vitest run src/_internal/types/MergeDeep.spec.ts src/object/merge.spec.ts src/object/mergeWith.spec.ts src/object/toMerged.spec.ts`
- [x] `yarn typecheck`
- [x] Local benchmark comparison for `merge` / `mergeWith` / `toMerged` after the internal refactor (no material regression after optimization)


raon0211:
Hi @soilSpoon, thank you for this PR and for pushing the deep-merge typing forward — and apologies for the long wait.

The core problem this PR tackles is now addressed in #1959: `merge()`'s return type is no longer `T & S`, but a new `Merge<T, S>` type that merges nested object types property by property, so overlapping keys no longer collapse to `never`. The type is also exported publicly from `es-toolkit/types`, so you can use it standalone (docs: `Merge`).

We ended up taking a slightly different direction from the `.deep()` namespace approach proposed here: instead of adding new opt-in APIs, we changed the default return type of `merge()` itself. Since the new type models what `merge` already does at runtime (skipping `undefined` source values, replacing non-plain objects, merging tuples index by index), we felt the risk of inference drift was acceptable and preferred keeping a single API surface — in line with es-toolkit's principle of the simplest interface for the common case.

Some ideas from this PR (e.g. aligning `mergeWith`'s input types with `merge`/`toMerged`, and deep result types for `mergeWith`/`toMerged`) aren't covered by #1959 — if you'd like to pursue those separately, a focused follow-up would be very welcome.

Closing this in favor of #1959. Thanks again for the thorough work here! 🙏

PR #1694, opened 2026-04-22
fix(AbortError, TimeoutError): fall back to Error when DOMException is undefined

### LINKED ISSUE #1693: [1.46.0]: React Native/Expo DOMException error
when updating es-toolkit to 1.46.0 i am getting an error when using debounce in React Native/ Expo mobile app

Image

SrAnthony:
## Summary

Since 1.46.0 (#1660), `AbortError` and `TimeoutError` extend `DOMException`. This breaks on any JS runtime that does not expose `DOMException` as a global — most notably **Hermes** (the default engine for React Native and Expo), but also older Node versions, some workers, and QuickJS.

On those runtimes, merely importing the module throws at evaluation time:

```
ReferenceError: Property 'DOMException' doesn't exist
```

which cascades through every consumer (`es-toolkit/function` via `debounce` / `throttle`, `es-toolkit/promise` via `delay` / `timeout`, etc.), crashing the whole bundle at startup.

Fixes #1693.

## Fix

Pick the base constructor at module load — `DOMException` when it exists, `Error` otherwise:

```ts
const AbortErrorBase: { new (message?: string): Error } =
  typeof DOMException !== 'undefined' ? DOMException : Error;

export class AbortError extends AbortErrorBase {
  constructor(message = 'The operation was aborted') {
    super(message);
  }
}
```

- Browsers and modern Node keep extending `DOMException`, so `instanceof DOMException` checks and the AbortSignal-detection compatibility that motivated #1660 continue to work.
- Hermes / older Node / other runtimes without `DOMException` get a plain `Error` subclass instead of a crash.
- `instanceof Error` holds in both paths.

Same change applied to `TimeoutError`.

## Tests

Added `src/error/AbortError.spec.ts` and `src/error/TimeoutError.spec.ts` covering:

- default and custom messages
- `instanceof Error`
- `instanceof DOMException` when available (skipped via `it.skipIf` on runtimes without it)

Full suite passes locally (`yarn vitest run` — 504 files, 4342 tests). `yarn tsc --noEmit` and `yarn eslint src/error/` are clean.

## Test plan

- [x] `yarn vitest run src/error src/promise`
- [x] `yarn vitest run` (full suite)
- [x] `yarn tsc --noEmit`
- [x] `yarn eslint src/error/`
- [x] Verified that downgrading to this fix unblocks a real React Native / Expo (Hermes) app that was crashing on `import { debounce } from 'es-toolkit/function'`

### DIFF
--- a/src/error/AbortError.spec.ts
+++ b/src/error/AbortError.spec.ts
@@ -1,4 +1,4 @@
-import { describe, expect, it } from 'vitest';
+import { afterEach, describe, expect, it, vi } from 'vitest';
 import { AbortError } from './AbortError';
 
 describe('AbortError', () => {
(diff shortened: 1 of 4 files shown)


raon0211 (approved):
Thanks for fixing this!

PR #1695, opened 2026-04-23
fix: add missing forEach/countBy exports to map and set

umsungjun:
Implementations, tests, and docs already exist, but the exports were missing from the index files, so added them.


raon0211 (approved):
Thanks for spotting this!

PR #1700, opened 2026-04-27
fix[throttle]: align repeated call timing

### LINKED ISSUE #1611: Timing of calls from `throttle` not as expected.
Part of the description of `throttle` states:
> Unlike `debounce`, throttle ensures the function is executed at least once during the specified time period.

The actual implementation does *not* ensure this. (... it's useful as-is, but the actual behavior seems difficult to describe...)

The "Lodash compatible" variant does not make promises regarding timing beyond "at most once per specified time interval". This is similar to what Lodash itself promises. There is, however, a promise of 100% compatibility with Lodash for `es-toolkit/compat`, and they don't actually behave the same.

I made a test-script to compare the behavior of es-toolkit to Lodash and Underscore.js:

```typescript
#!/usr/bin/env -S yarn tsx

import toolkitCompat from "es-toolkit/compat";
import * as toolkit from "es-toolkit";
import lodash from "lodash";
import underscore from "underscore";

function delay(ms: number): Promise<void> {
  return new Promise((resolve, _reject) => {
    setTimeout(() => {
      resolve();
    }, ms);
  });
}

function logTimes(start: number, invoked: number): void {
  console.log({
    throttled: invoked - start,
    actual: Date.now() - start,
  });
}

const throttleMs = 50;

async function test(
  label: string,
  throttled: (throttledInvoked: number) => void,
): Promise<void> {
  console.log("");
  console.log(label);
  throttled(Date.now());
  setTimeout(() => {
    throttled(Date.now());
  }, 20);
  setTimeout(() => {
    throttled(Date.now());
  }, 40);
  setTimeout(() => {
    throttled(Date.now());
  }, 60);
  setTimeout(() => {
    throttled(Date.now());
  }, 80);

  setTimeout(() => {
    throttled(Date.now());
  }, 300);
  setTimeout(() => {
    throttled(Date.now());
  }, 320);

  await delay(400);
}

async function main() {
  await test(
    "Underscore.js",
    underscore.throttle(logTimes.bind(null, Date.now()), throttleMs),
  );
  await test(
    "Lodash",
    lodash.throttle(logTimes.bind(null, Date.now()), throttleMs),
  );
  await test(
    "es-toolkit -- Lodash compatibility",
    toolkitCompat.throttle(logTimes.bind(null, Date.now()), throttleMs),
  );
  await test(
    "es-toolkit",
    toolkit.throttle(logTimes.bind(null, Date.now()), throttleMs),
  );
}

main();

```

I set up functions throttled to 50 ms, and log when the throttled function was called as well as when the underlying function was called. The throttled function is called first in a burst every 20 ms, at 0, 20, 40, 60 and 80 ms, then later at 300 and 320 ms.

All of the throttle variants are called with default options, meaning leading and trailing calls enabled.

On my computer, with es-toolkit 1.44.0, lodash 4.17.23, underscore 1.13.7, a sample run gives:

```
Underscore.js
{ throttled: 0, actual: 1 }
{ throttled: 41, actual: 51 }
{ throttled: 81, actual: 101 }
{ throttled: 301, actual: 301 }
{ throttled: 321, actual: 351 }

Lodash
{ throttled: 0, actual: 0 }
{ throttled: 40, actual: 50 }
{ throttled: 80, actual: 110 }
{ throttled: 300, actual: 300 }
{ throttled: 320, actual: 351 }

es-toolkit -- Lodash compatibility
{ throttled: 0, actual: 1 }
{ throttled: 81, actual: 81 }
{ throttled: 301, actual: 301 }
{ throttled: 321, actual: 371 }

es-toolkit
{ throttled: 1, actual: 1 }
{ throttled: 62, actual: 62 }
{ throttled: 81, actual: 132 }
{ throttled: 301, actual: 301 }
{ throttled: 321, actual: 371 }
```

I assume that the offsets of one or two milliseconds are from overhead from running the code in Node.js in general, and not relevant wrt. comparison of throttle implementations.

The behavior of Underscore.js is entirely straightforward, the underlying function is called at the specified rate.

es-toolkit seems to have some preference for waiting a bit and performing a later call "immediately", which seems reasonable, though it is arguably in conflict with the documentation.

Lodash does something more complicated after the second call to the underlying function, and I'm not convinced that it's intentional.

The "Lodash compatible" throttle from es-toolkit does something else again. It doesn't behave like es-toolkit, which a naive implementation just translating the `options` parameter format might do, but it *also* doesn't behave like Lodash. ... in fact, the timings of the es-toolkit `throttle` which does *not* attempt to behave like Lodash behaves closer to Lodash?

KimHyeongRae0:
## Summary

Fixes #1611.

This updates both `throttle` and `compat/throttle` for repeated calls with default leading/trailing behavior.

`throttle` now schedules trailing calls from the last actual invocation instead of pushing them back on every incoming call. `compat/throttle` now follows Lodash's timing more closely for the repeated-call case from the issue.

## Problem

With calls made repeatedly inside the throttle window, the previous implementation could delay the next invocation until the burst stopped.

That conflicted with the strict `throttle` documentation, and the compat version also did not match Lodash's behavior for the same timing pattern.

## Changes

- Implement strict `throttle` timing directly instead of relying on debounce behavior
- Keep the throttle window based on the last actual invocation
- Add a lodash-style scheduler for `compat/throttle`
- Add regression tests for repeated calls in both implementations

## Test plan

- [x] `yarn vitest run src/function/throttle.spec.ts src/compat/function/throttle.spec.ts`
- [x] `yarn typecheck`
- [x] `yarn eslint src/function/throttle.ts src/function/throttle.spec.ts src/compat/function/throttle.ts src/compat/function/throttle.spec.ts`
- [x] `yarn prettier --check src/function/throttle.ts src/function/throttle.spec.ts src/compat/function/throttle.ts src/compat/function/throttle.spec.ts`

### DIFF
--- a/src/compat/function/throttle.spec.ts
+++ b/src/compat/function/throttle.spec.ts
@@ -55,6 +55,51 @@ describe('throttle', () => {
     expect(callCount).toBe(2);
   });
 
+  it('should match lodash timing for repeated default calls', () => {
+    vi.useFakeTimers();
+    vi.setSystemTime(0);
+
+    try {
+      const calls: Array<[number, number]> = [];
+      const throttled = throttle((value: number) => {
+        calls.push([value, Date.now()]);
+      }, 50);
+
+      throttled(0);
+      vi.advanceTimersByTime(20);
+      throttled(20);
+      vi.advanceTimersByTime(20);
+      throttled(40);
+      vi.advanceTimersByTime(20);
+      throttled(60);
+      vi.advanceTimersByTime(20);
+      throttled(80);
+      vi.advanceTimersByTime(30);
+
+      expect(calls).toEqual([
+        [0, 0],
+        [40, 50],
+        [80, 110],
+      ]);
+
+      vi.advanceTimersByTime(190);
+      throttled(300);
+      vi.advanceTimersByTime(20);
+      throttled(320);
+      vi.advanceTimersByTime(30);
+
+      expect(calls).toEqual([
+        [0, 0],
+        [40, 50],
+        [80, 110],
+        [300, 300],
+        [320, 350],
+      ]);
+    } finally {
+      vi.useRealTimers();
+    }
+  });
+
   it('should not trigger a trailing call when invoked once', async () => {
     let callCount = 0;
     const throttled = throttle(() => {
--- a/src/compat/function/throttle.ts
+++ b/src/compat/function/throttle.ts
@@ -1,4 +1,4 @@
-import { debounce, DebouncedFunc, DebouncedFuncLeading } from './debounce.ts';
+import type { DebouncedFunc, DebouncedFuncLeading } from './debounce.ts';
 
 interface ThrottleSettings {
   /**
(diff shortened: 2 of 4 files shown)


dayongkr:
Thank you for the contribution. The spec for debounce and throttle requires some clarification and internal discussion, so please expect the review to take some time.

joshgummersall:
Any updates on this?

PR #1704, opened 2026-05-04
fix: (flatten) prevent stack overflow in flatten for deeply nested arrays

sweety-HJH223:
**Problem**
**flatten** with `Infinity` depth uses recursion internally, causing 
RangeError: Maximum call stack size exceeded` for deeply nested arrays.

**Solution**
Replaced recursive approach with an iterative stack-based implementation.

**Changes**
- flatten.ts: converted recursive function to iterative using explicit stack
- flatten.spec.ts: added test for 10,000 depth nested array with `Infinity`

### DIFF
--- a/src/array/flatten.spec.ts
+++ b/src/array/flatten.spec.ts
@@ -60,4 +60,13 @@ describe('flatten', () => {
 
     expect(flatten(originArr, 2)).toEqual([]);
   });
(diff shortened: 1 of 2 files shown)


sweety-HJH223:
Closing in favor of the maintainer's upcoming PR. Happy to have contributed 
by identifying the bug in the proposed solution!

PR #1705, opened 2026-05-05
perf(array): Prevent stack overflow in flatten for deeply nested arrays

### LINKED ISSUE #1703: Prevent stack overflow in flatten for deeply nested arrays
### Problem

`flatten` with `Infinity` depth uses recursion internally, making it vulnerable to stack overflow on deeply nested arrays:

```typescript
import { flatten } from 'es-toolkit';

const deep = [1];
for (let i = 0; i < 10000; i++) {
  deep[0] = [deep[0]];
}

flatten(deep, Infinity); 
// RangeError: Maximum call stack size exceeded
```

While 10,000+ depth is uncommon, it represents a **reliability concern**: 
- Using `Infinity` implies "flatten all levels," but the recursive implementation can't guarantee that
- Unexpected deeply nested data (malicious input, serialization edge cases) can crash the process

An iterative approach would make `flatten` robust regardless of nesting depth.

### Proposed Solution

Convert to an iterative approach using an explicit stack:

```typescript
export function flatten<T, D extends number>(arr: readonly T[], depth = 1): Array<FlatArray<T[], D>> {
  const flooredDepth = Math.floor(depth);
  if (flooredDepth < 1) return Array.from(arr) as Array<FlatArray<T[], D>>;
  
  const result: Array<FlatArray<T[], D>> = [];
  const stack: [readonly unknown[], number, number][] = [[arr, 0, flooredDepth]];
  
  while (stack.length > 0) {
    const [current, index, depthRemaining] = stack[stack.length - 1]!;
    
    if (index >= current.length) {
      stack.pop();
      continue;
    }
    
    const item = current[index];
    stack[stack.length - 1]![1]++;
    
    if (depthRemaining > 1 && Array.isArray(item)) {
      stack.push([item, 0, depthRemaining - 1]);
    } else {
      result.push(item as FlatArray<T[], D>);
    }
  }
  
  return result;
}
```

### Benchmark Results

image

| Nesting Depth | Recursive (old) | Iterative (new) | Winner |
|--------------|----------------|-----------------|--------|
| depth 3 | 4,571,786 ops/s | 3,221,234 ops/s | Recursive (1.42x) |
| depth 5 | 3,856,432 ops/s | 2,489,123 ops/s | Recursive (1.55x) |
| depth 100 | 775,432 ops/s | 805,671 ops/s | Iterative (1.04x) |
| depth 1000 | 75,834 ops/s | 80,833 ops/s | Iterative (1.07x) |

**Trade-off**: Slightly slower for shallow nesting (depth < 10), but faster for deep nesting and eliminates stack overflow risk entirely.

### Test Added

```typescript
it('should handle deeply nested arrays without stack overflow', () => {
  const createNestedArray = (depth: number): unknown[] => {
    if (depth === 0) return [1];
    return [createNestedArray(depth - 1)];
  };
  
  const arr = createNestedArray(10000);
  expect(() => flatten(arr, Infinity)).not.toThrow();
});
```

### Checklist
- [x] Added tests for deeply nested arrays
- [x] Verified all existing tests pass
- [x] Benchmarked performance across different nesting depths

byungsker:
## Summary

Converts the recursive implementation of `flatten` to an iterative approach using an explicit stack. This prevents `RangeError: Maximum call stack size exceeded` on deeply nested arrays when using `Infinity` depth.

## Changes

- Replaced recursive calls with an explicit stack-based iteration
- Added stress test for arrays nested up to 10,000 levels deep
- Maintains backward compatibility with all existing behavior (depth=0, depth=1, Infinity, NaN, negative values, floating point depths)

## Benchmarks

| Nesting Depth | Recursive (old) | Iterative (new) | Winner |
|--------------|----------------|-----------------|--------|
| depth 3 | 4,571,786 ops/s | 3,221,234 ops/s | Recursive (1.42x) |
| depth 5 | 3,856,432 ops/s | 2,489,123 ops/s | Recursive (1.55x) |
| depth 100 | 775,432 ops/s | 805,671 ops/s | Iterative (1.04x) |
| depth 1000 | 75,834 ops/s | 80,833 ops/s | Iterative (1.07x) |

Trade-off: Slightly slower for shallow nesting (depth < 10), but faster for deep nesting and eliminates stack overflow risk entirely.

## Testing

- All existing tests pass
- Added new test for deeply nested arrays without stack overflow

Fixes #1703

### DIFF
--- a/src/array/flatten.spec.ts
+++ b/src/array/flatten.spec.ts
@@ -60,4 +60,16 @@ describe('flatten', () => {
 
     expect(flatten(originArr, 2)).toEqual([]);
   });
+
+  it('should handle deeply nested arrays without stack overflow', () => {
+    // Create a deeply nested array (depth 10000)
+    let deepArr: unknown[] = [42];
+    for (let i = 0; i < 10000; i++) {
+      deepArr = [deepArr];
+    }
+
+    // This would throw "Maximum call stack size exceeded" with recursive implementation
+    expect(() => flatten(deepArr, Infinity)).not.toThrow();
+    expect(flatten(deepArr, Infinity)).toEqual([42]);
+  });
 });
--- a/src/array/flatten.ts
+++ b/src/array/flatten.ts
@@ -19,7 +19,7 @@ export function flatten<T, D extends number = 1>(arr: readonly T[], depth = 1 as
   const flooredDepth = Math.max(0, Math.floor(depth));
 
   if (flooredDepth === 0) {
-    return arr as Array<FlatArray<T[], D>>;
+    return Array.from(arr) as Array<FlatArray<T[], D>>;
   }
 
   // Use iterative approach to avoid stack overflow on deeply nested arrays
--- a/src/array/flatten.ts
+++ b/src/array/flatten.ts
@@ -16,19 +16,34 @@
  */
 export function flatten<T, D extends number = 1>(arr: readonly T[], depth = 1 as D): Array<FlatArray<T[], D>> {
   const result: Array<FlatArray<T[], D>> = [];
-  const flooredDepth = Math.floor(depth);
+  const flooredDepth = Math.max(0, Math.floor(depth));
 
-  const recursive = (arr: readonly T[], currentDepth: number) => {
-    for (let i = 0; i < arr.length; i++) {
-      const item = arr[i];
-      if (Array.isArray(item) && currentDepth < flooredDepth) {
-        recursive(item, currentDepth + 1);
-      } else {
-        result.push(item as FlatArray<T[], D>);
-      }
+  if (flooredDepth === 0) {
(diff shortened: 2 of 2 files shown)


raon0211:
I see why this could look like a bug, but flattening with `Infinity` depth is quite an edge case. es-toolkit's principle is to optimize for the common case, and supporting this would hurt performance there. 

I'd also note that `flatten(arr, Infinity)` is generally understood as *"flatten the array completely"*, which feels subtly different from *"flatten more than 10,000 levels deep."*

The native `Array#flat` has the same limitation, by the way. This also overflows the stack:

```typescript
const deep = [1];
for (let i = 0; i < 10000; i++) {
  deep[0] = [deep[0]];
}
deep.flat(Infinity);
```

image

raon0211:
Going to close this one out for now, but really appreciate you opening it!

PR #1723, opened 2026-05-15
fix(invert/invertBy): skip unsafe properties __proto__ in invert and invertBy

Antoliny0919:
## Problem

When proto or other special properties are passed as values to `invert` and `invertBy`, unexpected behavior may occur.

```js
const result = invert({ a: '__proto__', b: 'foo' });
```

**expected**

`{"__proto__": "a", "foo": "b"}`

**Actual**

`{"foo": "b"}`

Unlike expected, the key "proto" does not exist.
This behavior is an attempt to assign a value to a special property, but since it is not an actual object, it will not be modified.

```js
> const x = {'a': 1};
undefined
> x
{ a: 1 }
> x['b'] = 2;
2
> x
{ a: 1, b: 2 }
> x['__proto__'] = 3
3
> x
{ a: 1, b: 2 }
> x.__proto__
[Object: null prototype] {}
```

## Solution

Since only string, number, and symbol(`PropertyKey`) types can be used as values, there is no risk of prototype chain pollution.
However, it would be worth explicitly notifying users that keys such as `__proto__` are ignored.

### DIFF
--- a/src/compat/object/invertBy.ts
+++ b/src/compat/object/invertBy.ts
@@ -1,3 +1,4 @@
+import { isUnsafeProperty } from '../../_internal/isUnsafeProperty.ts';
 import { identity } from '../../function/identity.ts';
 import { isNil } from '../../predicate/isNil.ts';
 import { ValueIteratee } from '../_internal/ValueIteratee.ts';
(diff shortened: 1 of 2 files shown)


Antoliny0919:
On reflection, I think the correct behavior is not that `__proto__` is ignored, but rather that it should work like `"__proto__": "foo" ` 🤔

Antoliny0919:
I found this message in the documentation!
`Handling cases where internal object prototypes, like Array.prototype, have been modified.`
I think this behavior also falls under this guideline.
Sorry for the ping.
I'll close the PR :)

PR #1725, opened 2026-05-17
docs(compat): improve consistency of "Seq" method chaining guideline

Antoliny0919:
Improve consistency of method chaining guidelines in the compatibility documentation.

The english version was used as the reference for the guidelines.

https://github.com/toss/es-toolkit/blob/970ae85401f7e43c938bb83535d9145297bdf6cc/docs/compatibility.md?plain=1#L48


raon0211:
I guess the Korean version is more intuitive; I updated the English and Chinese docs accordingly. (https://github.com/toss/es-toolkit/pull/1731). Thanks for your suggestion!

Antoliny0919:
Thank you @raon0211 
Oh, I assumed the English version was always the primary one.
I also agree that it makes more sense for the more intuitive version to take precedence 👍
Next time I find an inconsistent sentence, I'll try to update it to the more intuitive version!

PR #1726, opened 2026-05-18
fix[isNumber]: remove instanceof Number check

JetProc:
## Summary

`isNumber` was the only core predicate that included an `instanceof` check alongside `typeof`. The other primitive type guards — `isString`, `isBoolean`, and `isSymbol` — all rely solely on `typeof`, making `isNumber` inconsistent with the rest of the predicate API.

This PR removes the `instanceof Number` check from `src/predicate/isNumber.ts` so that `isNumber` only returns `true` for primitive numbers, consistent with the other predicates.

## Changes

- `src/predicate/isNumber.ts`: removed `|| x instanceof Number`, updated JSDoc
- `src/predicate/isNumber.spec.ts`: moved `new Number(42)` from the `true` case to the `false` case
- `docs/reference/predicate/isNumber.md` (en, ko, ja, zh_hans): updated to reflect the new behavior

## Notes

The `compat` version (`src/compat/predicate/isNumber.ts`) is intentionally left unchanged, as it must match Lodash's behavior.

Closes #1652


dayongkr:
Removing `instanceof Number` is right — it matches `isString`/`isBoolean`/`isSymbol` 👍 I pushed a small commit keeping the docs minimal: just the `isNumber(new Number(42)); // false` example, and dropped the extra "primitive number" wording and the explanatory note since the example (and the spec) already make it clear. Thanks!

JetProc:
Thanks for merging the PR and for refining the documentation! Keeping it minimal and letting the example speak for itself definitely makes the docs cleaner. Appreciate the help! 🙌

PR #1727, opened 2026-05-20
fix(docs): fix issues in playground page

seungrodotlee:
# Summary

This PR fixes three issues in the Playground page.

1. Fixed an issue where functions with the same name exported from different categories could not be selected correctly in the Playground.
The root cause was that both the Playground data and selection state used only the function name as the key.
This was resolved by using a composite key consisting of both the category name and function name.
2. Fixed an issue where the interface description section at the top of the Playground box would shrink when the code became too long.
This was caused by an incorrect flex-shrink behavior, and was fixed by properly applying the flex-shrink setting.
3. Fixed an issue where console.log insertion behaved incorrectly when variable declarations spanned multiple lines.
Previously, console.log was always inserted on the line immediately following the start of the variable declaration.
This was updated to correctly detect the end of multi-line variable declarations and insert console.log on the following line.

# Changes

## Fix Same-Named Function Selection
|AS-IS|TO-BE|
|-|-|
|<video src="https://github.com/user-attachments/assets/5727be8f-7b46-4f80-b65d-038b68351ea1" />|<video src="https://github.com/user-attachments/assets/29287a75-2ce0-444d-bb0f-b228762b071d" />|

## Fix Description Area Shrinking
|AS-IS|TO-BE|
|-|-|
|image|image|

## Fix console.log Insertion for Multi-line Variables
|AS-IS|TO-BE|
|-|-|
|image|image|


raon0211 (approved):
Thanks so much!

PR #1729, opened 2026-05-21
fix(uniqWith): match lodash comparator argument order in compat

### LINKED ISSUE #1728: uniqWith (compat) calls comparator with reversed argument order vs lodash
### Summary

The JSDoc on `src/compat/array/uniqWith.ts` states the comparator is "invoked with two arguments: `(arrVal, othVal)`", matching the lodash documentation. The implementation in fact invokes it as `(othVal, arrVal)`. The argument order is reversed.

Any comparator whose body is asymmetric in its two parameters will silently produce a different result after switching from `lodash` to `es-toolkit/compat`.

### Versions

- `es-toolkit@1.46.1`
- `lodash@4.18.1`
- Node 20+

### Probe 1: argument order

```js
import uniqWithLodash from 'lodash/uniqWith.js';
import { uniqWith as uniqWithCompat } from 'es-toolkit/compat';

const probe = (impl) => {
  const calls = [];

  impl(['A', 'B'], (a, b) => {
    calls.push({ first: a, second: b });

    return false;
  });

  return calls;
};

console.log('lodash uniqWith:', probe(uniqWithLodash));
// lodash uniqWith: [ { first: 'B', second: 'A' } ]

console.log('es-toolkit/compat uniqWith:', probe(uniqWithCompat));
// es-toolkit/compat uniqWith: [ { first: 'A', second: 'B' } ]
```

When processing the second element (`'B'`), lodash calls the comparator as `(candidate, kept) = ('B', 'A')`, while es-toolkit/compat calls it as `(kept, candidate) = ('A', 'B')`.

### Probe 2: behavioral consequence

The comparator below assumes its first argument is the candidate and its second is the already-kept element, matching lodash's documented order. It drops the candidate when the candidate's set is a proper subset of the kept element's set. This is an edge-case use of `uniqWith`, but the documented argument order is still part of the contract.

```js
const sets = {
  Big: ['x', 'y', 'z'],
  Small: ['x'],
};

const bySize = Object.keys(sets).sort((a, b) => sets[b].length - sets[a].length);

const isSubset = (a, b) => {
  const diff = sets[a].filter((t) => !sets[b].includes(t));

  if (diff.length) {
    return false;
  }

  return sets[a].length < sets[b].length;
};

console.log('lodash result:', uniqWithLodash(bySize, isSubset));
// lodash result: [ 'Big' ]

console.log(
  'es-toolkit/compat result:',
  uniqWithCompat(bySize, isSubset)
);
// es-toolkit/compat result: [ 'Big', 'Small' ]
```

With lodash, `'Small'` is correctly identified as a subset of `'Big'` and dropped. With es-toolkit/compat, the comparator runs against the wrong direction (`difference(Big, Small)` instead of `difference(Small, Big)`) and keeps both keys.

ATOM00blue:
## Summary
`es-toolkit/compat`'s `uniqWith` invoked the comparator with reversed argument order compared to lodash. lodash calls the comparator as `(candidate, kept)` (documented `(arrVal, othVal)`), but compat passed the comparator straight to the non-compat implementation, which calls it as `(kept, candidate)`. Any asymmetric comparator silently produced different results after migrating from lodash.

## Changes
- Wrap the comparator in compat `uniqWith` so the toolkit's `(kept, candidate)` call is forwarded as `comparator(candidate, kept)`, matching lodash.
- Add a test asserting the comparator is invoked with `(candidate, kept)` order, plus a behavioral test for an asymmetric (subset) comparator.

The non-compat `uniqWith` keeps its own documented `(item1, item2)` semantics and is unchanged.

## Test plan
- `yarn vitest run src/compat/array/uniqWith.spec.ts` — 19 passed (new tests fail before the fix)
- `yarn vitest run src/compat/array/` — 1000 passed
- non-compat `uniqWith` unaffected; typecheck + lint clean

Closes #1728

### DIFF
--- a/src/compat/array/uniqWith.spec.ts
+++ b/src/compat/array/uniqWith.spec.ts
@@ -1,4 +1,4 @@
-import { describe, expect, expectTypeOf, it } from 'vitest';
+import { describe, expect, expectTypeOf, it, vi } from 'vitest';
 import * as lodashStable from 'es-toolkit/compat';
 import type { uniqWith as uniqWithLodash } from 'lodash';
 import { uniqWith } from './uniqWith';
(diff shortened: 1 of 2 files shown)


raon0211 (approved):
Thanks for catching ths!

PR #1734, opened 2026-05-24
docs: align sandpack header text across locales

Antoliny0919:
## Summary

Some sections using CodeSandbox have the header "Try It" while others use "Demo". For consistency, it would be better to unify the header text. Since CodeSandbox is meant for users to experiment with code, "Try It" feels more appropriate than "Demo" as it conveys the interactive nature of the sandbox.


raon0211 (approved):
Thanks!

PR #1735, opened 2026-05-24
docs[playground]: add link to playground editor title

Antoliny0919:
## Summary

Added a link to the playground editor title that navigates to the function's detail page. Users may want to explore a function in more detail after testing it in the playground, so linking to the detail page is expected to provide a better UX experience.

### Before

스크린샷 2026-05-24 오후 1 06 55

### After

es-toolkit-link-example

### DIFF
--- a/docs/.vitepress/components/Playground.vue
+++ b/docs/.vitepress/components/Playground.vue
@@ -30,9 +30,10 @@
     </div>
     <div class="playground-editor">
       <div class="playground-editor-header">
-        <span class="playground-editor-title">{{
-          selectedCategoryLabel === '' ? t.selectFunction : `${selectedCategoryLabel}/${selectedFunction}`
-        }}</span>
+        <a v-if="selectedFunction && docUrl" :href="docUrl" class="playground-editor-title playground-editor-title-link"
+          >{{ selectedCategoryLabel }}/{{ selectedFunction }}</a
+        >
+        <span v-else class="playground-editor-title">{{ t.selectFunction }}</span>
         <div class="playground-header-actions">
           <button
             v-if="currentDoc"
@@ -229,6 +230,13 @@ const selectedCategory = ref('');
 const selectedCategoryLabel = ref('');
 const openCategories = ref(new Set(['array']));
 const sandpackKey = ref(0);
+
+const docUrl = computed(() => {
+  if (!selectedFunction.value || !selectedCategory.value) return null;
+  const locale = lang.value;
+  const prefix = locale && locale !== 'en' && locale !== 'root' ? `/${locale}` : '';
+  return `${prefix}/reference/${selectedCategory.value}/${selectedFunction.value}`;
+});
 
 const currentDoc = computed(() => {
   if (!selectedFunction.value) {
@@ -454,6 +462,16 @@ function resetCode() {
   font-weight: 600;
   color: var(--vp-c-text-1);
   font-family: var(--vp-font-family-mono);
+}
+
+.playground-editor-title-link {
+  text-decoration: underline;
+  text-underline-offset: 4px;
+  transition: color 0.15s;
+}
+
+.playground-editor-title-link:hover {
+  color: var(--vp-c-brand-1);
 }
 
 .playground-header-actions {


raon0211 (approved):
Thanks for your help!

Antoliny0919:
Hello @raon0211 🙌
I have a question regarding the direction of the playground.
It seems the playground was introduced a bit early! (Very agile 😄)
All async-related examples are currently failing due to top-level await.(`filterAsync`, `flatMapAsync` ....)

I've been thinking about how to resolve this and would appreciate some guidance on the direction.
The first approach I considered was detecting await usage and automatically transforming the code into an IIFE in [playground.data.mts]() like ...
`(async () => ...)()`

```diff --git a/docs/.vitepress/data/playground.data.mts b/docs/.vitepress/data/playground.data.mts
index 2990880e..9a3646f6 100644
--- a/docs/.vitepress/data/playground.data.mts
+++ b/docs/.vitepress/data/playground.data.mts
@@ -46,6 +46,46 @@ interface ParseResult {
   doc: PlaygroundFunctionDoc;
 }
 
+function hasTopLevelAwait(code: string): boolean {
+  const lines = code.split('\n');
+  let braceDepth = 0;
+  let parenDepth = 0;
+  let bracketDepth = 0;
+
+  for (const line of lines) {
+    const trimed = line.trim();
+    if (trimed.startsWith('import ') || trimed === '') {
+      continue;
+    }
+
+    const awaitIndex = line.indexOf('await ');
+    if (awaitIndex !== -1 && braceDepth === 0 && parenDepth === 0 && bracketDepth === 0) {
+      const beforeAwait = line.slice(0, awaitIndex);
+      if (!beforeAwait.includes('=>')) {
+        return true;
+      }
+    }
+
+    for (const char of trimed) {
+      if (char === '{') {
+        braceDepth++;
+      } else if (char === '}') {
+        braceDepth--;
+      } else if (char === '(') {
+        parenDepth++;
+      } else if (char === ')') {
+        parenDepth--;
+      } else if (char === '[') {
+        bracketDepth++;
+      } else if (char === ']') {
+        bracketDepth--;
+      }
+    }
+  }
+
+  return false;
+}
+
 function parseMarkdown(content: string, fnName: string): ParseResult | null {
   // 1. Extract description (first paragraph after "# heading")
   let description = '';
@@ -122,6 +162,9 @@ function parseMarkdown(content: string, fnName: string): ParseResult | null {
     return "from 'es-toolkit'";
   });
 
+  // Sandpack doesn't support top-level await, so wrap body in async IIFE if needed.
+  const needsAsyncIIFE = hasTopLevelAwait(code);
+
   // Auto-insert console.log so results are visible in Sandpack console.
   if (!code.includes('console.log')) {
     const lines = code.split('\n');
@@ -217,7 +260,14 @@ function parseMarkdown(content: string, fnName: string): ParseResult | null {
   const importMatch = code.match(/^(import\s+\{[^}]*\}\s+from\s+'[^']+';?\s*\n?)/);
   if (importMatch) {
     const importLine = importMatch[1].trimEnd();
-    const restCode = code.slice(importMatch[0].length).replace(/^\n+/, '');
+    let restCode = code.slice(importMatch[0].length).replace(/^\n+/, '');
+    if (needsAsyncIIFE) {
+      const indented = restCode
+        .split('\n')
+        .map(l => (l.trim() ? `  ${l}` : ''))
+        .join('\n');
+      restCode = `(async () => {\n${indented}\n})();`;
+    }
     return { code: `${importLine}\n\n${commentLines.join('\n')}\n\n${restCode}`, doc };
   }
```

However, the format of examples extracted from markdown files turned out to be more inconsistent than expected, causing many edge cases.

So I've come to think that embedding Sandpack directly in each markdown file would be the most flexible solution.
For now, would it be a good idea to start by adding "Try It" Sandpack blocks to the markdown files of the affected functions, and work through them one by one?

raon0211:
Yeah, this one's tricky... What if we parse the example code with an AST parser like [oxc-parser](https://oxc.rs/docs/guide/usage/parser), check whether it has top-level awaits, and wrap it in an async IIFE if it does? Seems pretty straightforward to implement (and you could lean on an LLM if you're not sure how -- this is very straightforward for LLMs to implement).

Antoliny0919:
> Yeah, this one's tricky... What if we parse the example code with an AST parser like [oxc-parser](https://oxc.rs/docs/guide/usage/parser), check whether it has top-level awaits, and wrap it in an async IIFE if it does? Seems pretty straightforward to implement (and you could lean on an LLM if you're not sure how -- this is very straightforward for LLMs to implement).

Thank you for the great suggestion! :)
I think checking for top-level awaits with oxc-parser and wrapping them in an IIFE is a great idea!
However, I think it would still be difficult to solve the problem even with this approach 🤔

Most playground examples with async related functions use functions that are not defined.

### filterAsync

스크린샷 2026-05-30 오전 11 24 06

(checkUserStatus, isEvenAsync)

- FlatMapAsync
- forEachAsync
- limitAsync
- mapAsync
- reduceAsync
...


I think this issue comes from pulling the code directly from usage.
In my opinion, the code used in usage and playground should be expressed differently since they serve different purposes.
Of course, there are some functions where the usage code can be applied to the playground as well!
But async-related functions don't seem to fit that case.
I still think async-related function docs need a "Try It" section.
It might seem a bit odd that we need to modify the documentation for a feature just to fix a playground issue.
However, since the playground is set up to prioritize the sandpack present in the "Try It" section, I think this is an anticipated handling of such cases, so I'm totally fine with it 😊

Could you take another look at this?
If you have any better ideas, feel free to share them!

PR #1736, opened 2026-05-24
style[banner]: adjust banner content navigation button size

Antoliny0919:
## Summary

The navigation buttons inside the banner located at the bottom right of the screen are too small.(6 * 6)
too-small

Even I, who don't usually have difficulty with small buttons, had to attempt clicking two or three times.

According to [W3C guidelines](https://www.w3.org/WAI/WCAG22/Understanding/target-size-minimum.html), a minimum size of 24 x 24 is recommended.
However, applying this standard to these buttons would likely involve style trade offs.

That said, I still think the button size needs to be increased somewhat, even if it doesn't fully meet the 24 x 24 requirement.

### Before (6 * 6)

look-example-full-sc

### After (12 * 12)

(If 12 feels a bit too large, 10 would look fine as well  😁)

look-example-af-img


raon0211:
I think adding previous/next buttons works better aesthetically than making the dots bigger.

Antoliny0919:
> I think adding previous/next buttons works better aesthetically than making the dots bigger.

I agree. Adding navigation buttons would be more visually explicit.
Thanks for suggesting the direction!

@raon0211 🙌
If you have time, could you also take a look at this comment?
https://github.com/toss/es-toolkit/pull/1735#issuecomment-4527406995

raon0211:
Thank you for the contribution! Could you open a new PR that adds the previous/next buttons? I'll close this PR in favor of that one.

PR #1744, opened 2026-05-30
docs(banner): add ja and zh-hand versions banner

Antoliny0919:
## Summary

Added Japanese and Chinese banner versions.
I believe the frontendFundamental docs also support Japanese and Chinese.
I also think the other libraries can be used regardless of the user's language.

### ja

스크린샷 2026-05-30 오후 12 08 09

### zh-hans

스크린샷 2026-05-30 오후 12 08 45


Antoliny0919:
Was there a specific reason why they weren't added at that time? 🤔

https://github.com/toss/es-toolkit/commit/f12166f29abb3b57f3d6841b9b8063513c519792

raon0211 (approved):
Thanks!

PR #1745, opened 2026-05-30
docs(banner): fix incorrect unmounted syntax

Antoliny0919:
## Summary

The existing code attempts cleanup by returning from `onMounted`.(Just like React `useEffect`)
However, this syntax doesn't actually exist in Vue.
Therefore, `onUnmounted` should be used for proper cleanup.
It won't cause issues right away, but it could potentially lead to problems down the line.

(Please let me know if I'm wrong about this! 🥲)


raon0211 (approved):
Thanks for catching this!

PR #1747, opened 2026-05-31
refactor: remove inferable JSDoc types from TS signature

### LINKED ISSUE #1558: Redundancy between JSdoc types and function signature types — remove JSdoc types?

Antoliny0919:
## Summary
Closes: https://github.com/toss/es-toolkit/issues/1558

Removed types from JSDoc that can be inferred from TS signatures.

## Changes

Converted JSDoc parameters across all functions to the following format.

### Before

```ts
 * @param {T[]} firstArr - The array from which to get the difference.
 * @param {U[]} secondArr - The array containing elements to exclude from the first array.
 * @param {(x: T, y: U) => boolean} areItemsEqual - A function to determine if two items are equal.
 * @returns {T[]} A new array containing the elements from the first array that do not match any elements in the second array
 export function differenceWith<T, U>(
  firstArr: readonly T[],
  secondArr: readonly U[],
  areItemsEqual: (x: T, y: U) => boolean
): T[] {
```

### After

```ts
 * @param firstArr - The array from which to get the difference.
 * @param secondArr - The array containing elements to exclude from the first array.
 * @param areItemsEqual - A function to determine if two items are equal.
 * @returns A new array containing the elements from the first array that do not match any elements in the second array
 export function differenceWith<T, U>(
  firstArr: readonly T[],
  secondArr: readonly U[],
  areItemsEqual: (x: T, y: U) => boolean
): T[] {
```


Antoliny0919:
It seems like this pattern was introduced by automatic documentation generation .
But I'm not sure whether the automatic documentation generation, as mentioned in the issue, is still being used or actively maintained. 🤔

```sh
> deno run --allow-read --allow-write --allow-env --allow-net generate-docs.mts
error: Uncaught (in promise) Error: No description provided for cartesianProduct (file:///.../es-toolkit/src/array/cartesianProduct.ts:83:0).
```

It would be greatly appreciated if you could point out the parts I should review.
The behavior before and after should remain the same :)
I'll make time to review as thoroughly as possible!

raon0211 (approved):
Yeah, seems that the types in JSDoc were redundant. Thanks!

Antoliny0919:
Hey @raon0211  !!!

Do you still use Deno to generate the docs? As far as I know, it seems like it's no longer being used. How about removing the related features/dependencies? I think it could be a confusing point for beginner contributors who are trying to understand the codebase!

If you agree, would it be okay for me to remove it? ☺️

PR #1749, opened 2026-06-01
fix(maxBy/minBy): propagate NaN to match Math.max/Math.min

pbk95120:
## Problem
`maxBy` / `minBy` return a leading `NaN` element even when valid values exist:

```ts
maxBy([NaN, 1, 3], x => x); // returned NaN, expected 3
minBy([NaN, 3, 1], x => x); // returned NaN, expected 1
```

## Root cause
Before #1176, the running extreme was seeded with `-Infinity` / `Infinity`
and the loop started at index 0, so a `NaN` value lost every comparison and
was skipped regardless of position.

#1176 ("improve type definitions and empty array handling") added the
empty-array guard but also changed the seed to `getValue(items[0])` and the
loop start to `i = 1` (described in that PR as a loop-start performance
optimization). That change was unnecessary once the guard existed, and it
reintroduced the leading-NaN bug. The existing NaN test only covered a
mid-array NaN (`[1, NaN, 3, 2]`, first element valid), so the regression
went unnoticed.

In the #1176 review, skipping `NaN` was confirmed as the intended behavior
("skipping NaN is the intended behavior"), which this change makes hold
regardless of the NaN's position.

## Fix
Restore the `-Infinity` / `Infinity` seed and iterate from index 0, and add
a regression test for a leading `NaN`. Empty-array (`undefined`), all-NaN
(first element), tie (first element), and legitimate `-Infinity` / `Infinity`
values are unchanged. The only runtime difference is one extra primitive
comparison per call (`getValue` is still invoked exactly once per element).


raon0211:
`maxBy` and `minBy` should match the behavior of the built-in `Math.max()` and `Math.min()` exactly. For example, `Math.min(NaN, 1, 2, 3)` returns `NaN`, so `minBy([NaN, 1, 2, 3], x => x)` should do the same.

pbk95120:
Thanks for the review!

Before changing direction, one clarification — the current implementation doesn't match `Math.max` / `Math.min` today either.

Mid-array `NaN` is already silently skipped, since `value > max` (and `value < min`) is false when `value` is `NaN`:

```ts
maxBy([1, NaN, 3], x => x); // 3, not NaN
minBy([1, NaN, 3], x => x); // 1, not NaN
```

So the only inconsistency was the leading position, which is what this PR fixes.

Fully matching `Math.max` / `Math.min` would mean propagating `NaN` from any position — a larger change that would also flip the mid-array cases above.

Worth noting:

* In #1176, skipping `NaN` was confirmed as the intended behavior. This PR just makes it consistent regardless of position.
* Lodash's `_.maxBy` / `_.minBy` also skip `NaN` (via the `current === current` check in `baseExtremum`), so propagating would diverge from lodash as well.

Which direction would you prefer?

(a) Skip `NaN` everywhere — this PR
(b) Propagate `NaN` to match `Math.max` / `Math.min` exactly — a separate, larger change affecting mid-array `NaN` too

Happy to rework for (b) if that's the intent.

raon0211:
Thanks for laying out the options so clearly.

Let's go with (b). es-toolkit is separate from lodash (lodash compatibility lives in es-toolkit/compat), and our intent is to follow modern JavaScript built-in behavior as closely as possible. So matching `Math.max` / `Math.min` exactly — propagating `NaN` regardless of position — is the direction we want.

pbk95120:
Switched to (b) as discussed — thanks for the direction

`maxBy`/`minBy` now propagate `NaN` regardless of its position, matching `Math.max`/`Math.min` exactly:

```ts
maxBy([NaN, 1, 3], x => x); // NaN
maxBy([1, NaN, 3], x => x); // NaN
maxBy([1, 3, NaN], x => x); // NaN
minBy([NaN, 3, 1], x => x); // NaN  (same for every position)

raon0211 (approved):
Thanks for your help!

PR #1750, opened 2026-06-01
docs: fix grammar in clone JSDoc examples

shaked-shlomo:
Several `clone`-family `@example` comments read **"Clone a primitive values"** (and one **"Clone a primitive objs"**) — the article *a* with a plural noun.

Each example clones a single value (`const num = 29; clone(num);`), so the singular is correct. This changes them to **"Clone a primitive value"** / **"Clone a primitive object"**, which also matches the existing `cloneDeepWith` docs that already say "Clone a primitive value".

Affected files:
- `src/object/clone.ts`
- `src/object/cloneDeep.ts`
- `src/compat/object/clone.ts`
- `src/compat/object/cloneDeep.ts`
- `src/compat/object/cloneWith.ts`

Comment-only change; no code or behavior affected.

### DIFF
--- a/src/compat/object/clone.ts
+++ b/src/compat/object/clone.ts
@@ -35,7 +35,7 @@ import { isTypedArray } from '../predicate/isTypedArray.ts';
  * @returns {T} - A shallow clone of the given object.
  *
  * @example
- * // Clone a primitive objs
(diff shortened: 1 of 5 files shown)


raon0211 (approved):
Thanks!

PR #1752, opened 2026-06-02
docs(search): add ja tanslation to search

Antoliny0919:
## Summary

It appears that Japanese translations for the search feature were missing.
The translation values already existed but were not being applied 🥲

### Before

스크린샷 2026-06-02 오후 8 00 55

### After

스크린샷 2026-06-02 오후 8 01 28


raon0211 (approved):
Thanks!

PR #1758, opened 2026-06-07
fix(compat/toLength): coerce non-numeric values to 0

sarathfrancis90:
## Summary

`compat/toLength` returns `NaN` for values that coerce to `NaN`, where lodash returns `0`:

```js
import { toLength } from 'es-toolkit/compat';

toLength('a');    // NaN  (lodash: 0)
toLength('12px'); // NaN  (lodash: 0)
toLength(NaN);    // NaN  (lodash: 0)
toLength({});     // NaN  (lodash: 0)
```

lodash's `toLength` always produces a valid length in `[0, 2^32 - 1]` and never returns `NaN`. The cause here is `Math.floor(Number(value))`: for a non-numeric input that's `NaN`, and `clamp` passes `NaN` straight through.

I noticed `toInteger`, `toFinite`, and `toSafeInteger` already agree with lodash on these inputs, and the sibling `toSafeInteger` is implemented as `clamp(toInteger(value), ...)`. `toInteger` maps `NaN` to `0`, so routing `toLength` through it fixes the inconsistency and lines the two functions up. This also matches lodash, whose `toLength` is `baseClamp(toInteger(value), 0, MAX_ARRAY_LENGTH)`.

## Test

Extended `toLength.spec.ts` with the non-numeric cases (`'a'`, `'12px'`, `NaN`, `{}`, `undefined`); they failed before and pass now. The full `compat` suite stays green.

### DIFF
--- a/src/compat/util/toLength.spec.ts
+++ b/src/compat/util/toLength.spec.ts
@@ -23,6 +23,14 @@ describe('toLength', () => {
     expect(1 / toLength(-0)).toBe(Infinity);
   });
 
+  it('should coerce non-numeric values to `0`', () => {
+    expect(toLength('a')).toBe(0);
+    expect(toLength('12px')).toBe(0);
+    expect(toLength(NaN)).toBe(0);
+    expect(toLength({})).toBe(0);
+    expect(toLength(undefined)).toBe(0);
+  });
+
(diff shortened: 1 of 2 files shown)


raon0211 (approved):
Thanks for your pull request!

PR #1759, opened 2026-06-09
fix(retry): pass error to delay callback

### LINKED ISSUE #1707: Suggestion: Retry's delay with error parameter like shouldRetry
Hi,

We already replaced all of lodash with es-toolkit and now came forward to replace other dependencies as well in order to have less dependencies.
One very reliable source that we use is https://github.com/vercel/async-retry.
The API is a little sluggish and out-dated.

One main feature of it is, that we can implemented our own waiting time after a response delivered a status code like 429.
The response has HTTP headers that tell how long to wait before trying again.
We are doing this in the main function itself, wait and rethrow. But it looks promising that we could do this within the `delay` option if the error would be available as parameter, just like it is in `shouldRetry`
So the delay depends on the thrown error from the original function.

dohwi:
## Summary

Closes #1707. Enable error-aware retry intervals by passing the error object to the `delay` callback, consistent with how `shouldRetry` already receives it.

## Changes

- `delay` callback now receives `(attempts: number, error: unknown)` — error is passed as the second argument alongside the attempt number
- `delay?: number | ((attempts: number, error: unknown) => number)` type updated in both the `RetryOptions` interface and implementation
- Added a test verifying delay receives the correct error on retry
- Updated documentation in all 4 languages (en, ko, ja, zh_hans) with error-aware delay examples

## Test results

```
✓ src/function/retry.spec.ts (10 tests) 175ms
  ✓ should pass error to delay function
  ✓ should retry with a dynamic delay function (backward compat)
  ... 8 more all passing
```

Full suite: 560 test files, 4482 tests passed, 0 failures.


dohwi:
Thanks @copilot-pull-request-reviewer for the review.

- **Comment on `error` variable (line 149)**: The `error` variable is declared at line 131 in the outer function scope (`let error;`) and assigned in the catch block (`error = err;`). By line 149, `error` is always populated — the catch parameter `err` and the outer variable `error` are distinct. This is a false positive.

- **Comments on `attempt` vs `attempts` (lines 10, 91)**: Good catch — fixed in the latest commit.

- **`ResponseError` not defined in docs example**: Fixed — replaced with a self-contained type narrowing using `(error as { status?: number })`.

raon0211 (dismissed):

raon0211 (approved):

PR #1761, opened 2026-06-10
fix(globalThis): remove Function eval fallback for edge runtime compatibility

### LINKED ISSUE #1760: Function("return this")() in _internal/globalThis breaks Next.js / Vercel Edge Runtime
## Description

Since the `isBuffer`/`DOMException` refactor that introduced `src/_internal/globalThis.ts` ([commit](https://github.com/toss/es-toolkit/commit/7cc004bdc1e60e90e7ae51a577657f1a83837afb)), importing any function whose graph reaches `isBuffer` or `DOMException` fails to build under restricted runtimes (Next.js / Vercel Edge Runtime, Cloudflare Workers) with:

```
Error: Dynamic Code Evaluation (e.g. 'eval', 'new Function', 'WebAssembly.compile') not allowed in Edge Runtime
```

The culprit is the final fallback in `dist/_internal/globalThis.{js,mjs}`:

```js
const globalThis_ =
  (typeof globalThis === "object" && globalThis) ||
  (typeof window === "object" && window) ||
  (typeof self === "object" && self) ||
  (typeof global === "object" && global) ||
  (function () { return this; })() ||
  Function("return this")();   // ← rejected by Edge Runtime
```

Edge/worker bundlers scan for `eval`/`new Function`/`Function(...)` **syntactically and reject the module even though this branch is unreachable** — `globalThis` has been standard since Node 12 / all modern edge runtimes, so the `Function(...)` fallback can never execute.

## Reproduction

1. Next.js app (App Router, Next 14.2.x or 15.x).
2. In `middleware.ts` (which always runs in the Edge Runtime), import and use `cloneDeep` or `isEqual` from `es-toolkit` (or any function reaching `isBuffer`/`DOMException`).
3. `next build` → fails with the error above.

https://github.com/PSoltes/es-toolkit-edge-compatibility-issue

Functions importing only e.g. `merge`, `pick`, `chunk` are unaffected, since they never reach `_internal/globalThis`.

## Affected chains

- `cloneDeep` → `cloneDeepWith` → `isBuffer` → `_internal/globalThis`
- `isEqual` → `isEqualWith` → `isBuffer` → `_internal/globalThis`
- `delay` / `timeout` → `AbortError`/`TimeoutError` → `DOMException` → `_internal/globalThis`
- the entire `es-toolkit/compat` subpath (via `isBuffer`)

## Versions

- Last working: `1.45.1` (`isBuffer` was `typeof Buffer !== "undefined" && Buffer.isBuffer(x)`, no `globalThis` indirection).
- Broken: `^1.46.0`

CheHyeonYeong:
## Summary

Importing any function whose graph reaches `isBuffer` or `DOMException` fails to build under Next.js Edge Runtime / Cloudflare Workers since `1.46.0`:

```
Error: Dynamic Code Evaluation (e.g. 'eval', 'new Function', 'WebAssembly.compile') not allowed in Edge Runtime
```

The culprit is the final fallback in `_internal/globalThis.ts`:

```ts
Function('return this')()
```

Edge runtimes scan for `Function(...)` syntactically and reject the module at build time, regardless of whether the branch is reachable. Since `globalThis` is standard from Node 12 and all modern edge runtimes, the entire fallback chain is dead code.

This removes it and assigns `globalThis` directly:

```ts
const globalThis_: typeof globalThis = globalThis;
```

Fixes #1760

## Test

Existing tests for `isBuffer`, `cloneDeep`, `isEqual`, `AbortError`, `TimeoutError`, and the full `compat` suite all pass.

### DIFF
--- a/src/_internal/globalThis.ts
+++ b/src/_internal/globalThis.ts
@@ -1,14 +1,4 @@
-declare let global: typeof globalThis;
-
 // eslint-disable-next-line @typescript-eslint/naming-convention
-const globalThis_: typeof globalThis =
-  (typeof globalThis === 'object' && globalThis) ||
-  (typeof window === 'object' && window) ||
-  (typeof self === 'object' && self) ||
-  (typeof global === 'object' && global) ||
-  (function (this: unknown) {
-    return this;
-  })() ||
-  Function('return this')();
+const globalThis_: typeof globalThis = globalThis;
 
 export { globalThis_ as globalThis };


raon0211:
Hey, thanks for the contribution! I'm not sure we should remove the globalThis polyfill entirely though — for now let's just drop the Function(...) call. I've made that change in 10cc7ec4, so I'll close this PR. Thanks again for your help!

PR #1766, opened 2026-06-12
fix(compat/size): count array-like objects by their length

sarathfrancis90:
## Summary

`compat/size` returns the number of own enumerable keys for array-like objects, so it diverges from lodash for values that have a valid `length` but aren't real arrays:

```js
import { size } from 'es-toolkit/compat';
import _ from 'lodash';

size({ length: 3 });             // 1   (expected 3)
_.size({ length: 3 });           // 3

size({ 0: 'a', 1: 'b', length: 2 }); // 3   (expected 2)
_.size({ 0: 'a', 1: 'b', length: 2 }); // 2
```

lodash's `size` checks `isArrayLike(collection)` first and returns `collection.length` for array-like values, only falling back to the own-key count for plain objects. es-toolkit skipped that check and went straight to `Object.keys(target).length`.

## Fix

Route array-like values through `isArrayLike` and return their `length`, matching lodash. Objects whose `length` is not a valid array length (negative, fractional, non-number, or above `MAX_SAFE_INTEGER`) are still not treated as array-like and fall back to the own-key count, exactly as before — the existing `size` tests covering those cases keep passing.

Arrays, strings, and `arguments` objects are unaffected (they were already array-like and produced the same result). Maps and Sets are not array-like, so they still return `.size`.

## Tests

Added cases asserting array-like objects are counted by `length`. The new tests fail on `main` and pass with the fix. The full `src/compat` suite (3093 tests), `tsc --noEmit`, and lint all pass.

### DIFF
--- a/src/compat/array/size.spec.ts
+++ b/src/compat/array/size.spec.ts
@@ -66,6 +66,16 @@ describe('size', () => {
     expect(size(str)).toBe(10);
   });
 
+  it('should return the `length` of array-like objects', () => {
+    expect(size({ length: 3 })).toBe(3);
+    expect(size({ 0: 'a', 1: 'b', length: 2 })).toBe(2);
+    expect(size({ length: 0 })).toBe(0);
+  });
+
+  it('should prefer `length` over the number of own enumerable keys for array-like objects', () => {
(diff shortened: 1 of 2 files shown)


raon0211 (approved):
Thanks!

PR #1767, opened 2026-06-13
fix(compat/random): compatibility issues with random function

Antoliny0919:
### Summary

Currently, when floating is true, values are returned in a range 1 higher than max.
e.g. `when min=1, max=2 with floating enabled`, values like 2.xx can be returned.

### Reproduce
```ts
import { random as loRandom } from 'lodash';
import { random as esRandom } from './src/compat/math/random';

const call = 500;

function checkOver(value) {
  if (value > 2) {
    return true;
  }
}

let loOver2 = 0;
let esOver2 = 0;

for (let i = 0; i < call; i++) {
  const value = loRandom(1, 2, true);
  if (checkOver(value)) {
    loOver2 += 1;
  }
}

for (let i = 0; i < call; i++) {
  const value = esRandom(1, 2, true);
  if (checkOver(value)) {
    esOver2 += 1;
  }
}

console.log(loOver2); // 0
console.log(esOver2); // 246
```

### Summary

- random(1.5, 2.5)
- random(1.5)
- random(1.5, 2)

In all of the above cases, integers are returned.
However, Lodash returns a float if at least one float value is included in the arguments.

### REproduce

```ts
import { random as loRandom } from 'lodash';
import { random as esCompatRandom } from './src/compat/math/random';

const call = 500;

let loIsInteger = 0;
let esCompatIsInteger = 0;

for (let i = 0; i < call; i++) {
  const value = loRandom(1.5, 2.5);
  if (Number.isInteger(value)) {
    loIsInteger += 1;
  }
}

for (let i = 0; i < call; i++) {
  const value = esCompatRandom(1.5, 2.5);
  if (Number.isInteger(value)) {
    esCompatIsInteger += 1;
  }
}

console.log(loIsInteger); // 0
console.log(esCompatIsInteger); // 500
```

### Summary

Returns an integer when called with (max=some..., floating=true)
Shouldn't it return a float since `floating=true` 😁

### Reproduce

```ts
import { random as loRandom } from 'lodash';
import { random as esCompatRandom } from './src/compat/math/random';

const call = 500;

let loIsInteger = 0;
let esCompatIsInteger = 0;

for (let i = 0; i < call; i++) {
  const value = loRandom(2, true);
  if (Number.isInteger(value)) {
    loIsInteger += 1;
  }
}

for (let i = 0; i < call; i++) {
  const value = esCompatRandom(2, true);
  if (Number.isInteger(value)) {
    esCompatIsInteger += 1;
  }
}

console.log(loIsInteger); // 0
console.log(esCompatIsInteger); // 500
```

### DIFF
--- a/src/compat/math/random.spec.ts
+++ b/src/compat/math/random.spec.ts
@@ -90,13 +90,13 @@ describe('random', () => {
 
   it('should support providing a `floating`', () => {
     let actual = random(true);
-    expect(actual % 1 && actual >= 0 && actual <= 1);
+    expect(actual % 1 && actual >= 0 && actual <= 1).toBeTruthy();
 
     actual = random(2, true);
     expect(actual % 1 && actual >= 0 && actual <= 2);
 
     actual = random(2, 4, true);
-    expect(actual % 1 && actual >= 2 && actual <= 4);
+    expect(actual % 1 && actual >= 2 && actual <= 4).toBeTruthy();
   });
 
   it('should work as an iteratee for methods like `_.map`', () => {
--- a/src/compat/math/random.spec.ts
+++ b/src/compat/math/random.spec.ts
@@ -111,7 +111,7 @@ describe('random', () => {
     expect(actual % 1 && actual >= 0 && actual <= 1).toBeTruthy();
 
     actual = random(2, true);
-    expect(actual % 1 && actual >= 0 && actual <= 2);
+    expect(actual % 1 && actual >= 0 && actual <= 2).toBeTruthy();
 
     actual = random(2, 4, true);
     expect(actual % 1 && actual >= 2 && actual <= 4).toBeTruthy();
--- a/src/compat/math/random.spec.ts
+++ b/src/compat/math/random.spec.ts
@@ -79,13 +79,31 @@ describe('random', () => {
     }
   });
 
+  it('should support a float `min` or `max`', () => {
+    const floatMinSamples = array.map(() => random(1.5, 2));
+    expect(floatMinSamples.some(v => v % 1 !== 0)).toBeTruthy();
+    expect(floatMinSamples.every(v => v >= 1.5 && v <= 2)).toBeTruthy();
+
+    const floatMaxSamples = array.map(() => random(1, 2.5));
+    expect(floatMaxSamples.some(v => v % 1 !== 0)).toBeTruthy();
+    expect(floatMaxSamples.every(v => v >= 1 && v <= 2.5)).toBeTruthy();
+  });
+
+  it('should support a float `max`', () => {
+    const max = 2.5;
+    const samples = array.map(() => random(max));
+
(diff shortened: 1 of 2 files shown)


Antoliny0919:
I came across a useful tool for checking compat(skill)-> https://github.com/toss/es-toolkit/blob/main/.claude/skills/compat-review/SKILL.md
What do you think about making it available to run on PR 🤔
We could trigger it when a specific label is applied.

Having it automatically generate a report would make it much easier for reviewers to evaluate the changes.

src/compat/math/random.spec.ts
```diff
@@ -79,24 +79,42 @@ describe('random', () => {
     }
   });
 
-  it('should support floats', () => {
```
raon0211, line 82: This test actually comes from `lodash`, so I'd rather not change it much — especially the `expect` assertions. Could we keep those as-is and just add more tests on top instead?

Antoliny0919: Of course! I'll take a look over the weekend 😁

src/compat/math/random.spec.ts
```diff
@@ -79,24 +79,42 @@ describe('random', () => {
...
+
+  it('should support a float `max`', () => {
+    const max = 2.5;
+    const samples = array.map(() => random(max));
+
+    expect(samples.some(v => v % 1 !== 0)).toBeTruthy();
+    expect(samples.every(v => v >= 0 && v <= max)).toBeTruthy();
+  });
+
   it('should support floats', () => {
```
Antoliny0919, line 100: https://github.com/toss/es-toolkit/blob/2733faf3b64ac63c9a054c81cdb3499293417ab7/src/compat/array/chunk.spec.ts#L7
One thing I'm curious about: I noticed that some compat tests include links to the corresponding Lodash test cases, but many others (such as random) don't.

Is there a general convention around this?

If the goal is to keep Lodash's original test cases intact, would it make sense to split them into two describe blocks?

first describe: test cases from Lodash
second describe: additional test cases added by es-toolkit

I'm not sure why some tests have links while others don't, but it feels like we could benefit from a bit more consistency. Separating the tests might also make their purpose clearer.

(Though I admit that could turn into a somewhat tedious task... maybe manageable with a little help from AI 🤔)

raon0211: Ideally, every test should have a link pointing to its equivalent Lodash test. I think some contributors just omitted them along the way as we received contributions from many different people.

raon0211 (approved):

src/compat/math/random.spec.ts
Antoliny0919: Sounds good, I've taken a look 😁
It might take a bit of time, but I'll work on it.

While working on this, I felt it would be better to keep the two tests separated. When we revisit them in the future, it will be much easier to identify which tests originated from Lodash and which ones were added by us. I think that distinction will make the test suite much easier to understand and maintain.

PR #1770, opened 2026-06-13
docs: remove performance comparison section from the feature

Antoliny0919:
## Summary

I removed the feature comparison section from the feature documentation.
I think this section is one of the parts that makes it hard to maintain (and it's also inconsistent).
Of course, it's a good point to highlight the strengths of es-toolkit, but I think the bundle size and performance pages already do that sufficiently.

- https://es-toolkit.dev/ko/bundle-size.html
- https://es-toolkit.dev/ko/performance.html

If we want to keep the feature comparison section, it would be better to only include the functions that are already covered in the bundle size and performance pages.(sample, diffrence, sum, debounce ...)


raon0211 (approved):
Thanks so much!

PR #1771, opened 2026-06-13
fix(compat/parseInt): coerce input to string before parsing

sarathfrancis90:
While migrating some code from lodash I noticed `compat/parseInt` diverges from lodash on non-string inputs.

```js
import { parseInt } from 'es-toolkit/compat';

parseInt(null);          // lodash: NaN   es-toolkit: 23  (parses "null" as base 25 via map iteratee, "null" otherwise)
parseInt(null, 25);      // lodash: NaN   es-toolkit: 23
parseInt(Symbol('x'));   // lodash: NaN   es-toolkit: throws TypeError
```

lodash's `parseInt` runs the input through its own `toString` before handing it to the native `parseInt`, so `null`/`undefined` become `''` and symbols stringify instead of crashing. es-toolkit was passing the raw value straight to `Number.parseInt`, which relies on native coercion (`null` → `"null"`, symbols throw).

The fix routes the input through the existing `compat/util/toString`, matching lodash. Added tests for the `null`/`undefined` and symbol cases (they fail on `main`, pass with the change). Full `compat` suite, `tsc`, and lint are green.

### DIFF
--- a/src/compat/math/parseInt.spec.ts
+++ b/src/compat/math/parseInt.spec.ts
@@ -58,6 +58,23 @@ describe('parseInt', () => {
     expect(parseInt('0x20', object)).toBe(32);
   });
 
+  it('should coerce `string` to a string before parsing', () => {
+    // `null`/`undefined` coerce to `''`, not to `'null'`/`'undefined'`.
+    // @ts-expect-error - unusual usage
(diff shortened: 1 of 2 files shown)


raon0211 (approved):
Thanks!

PR #1772, opened 2026-06-13
docs: remove Lodash Compability sections

Antoliny0919:
## Summary

It seems like a section that is no longer needed.
(Probably a remnant from the past 🤔)
I also noticed that the remaining sections differ by language. I think it would be fine to remove it!


raon0211 (approved):
Thanks!

PR #1773, opened 2026-06-13
fix(compat/difference): treat sparse array holes as undefined for compability

Antoliny0919:
## Summary

There is a difference in return values between `compat/difference` and lodash when dealing with sparse arrays.
For compatibility, the two return values need to be made consistent.

## Reproduce

```ts
import { difference as loDiff } from 'lodash';
import { difference as esCompatDiff } from './src/compat/array/difference';

const arr = [1, , 3];

console.log(loDiff(arr, [3])); // [1, undefined]
console.log(esCompatDiff(arr, [3])); [1]
```


raon0211 (approved):
Thanks!

PR #1775, opened 2026-06-13
fix(compat/pick): prioritize literal key over dot notation path when value is undefined

Antoliny0919:
## Summary

In pick, when using dot notation, a literal key matching the dot-notation path takes priority, as documented.

스크린샷 2026-06-13 오후 9 43 10

However, there is a bug where the literal key is ignored when its value is `undefined`.

## Reproduce

```ts
import { pick as loPick } from 'lodash';
import { pick as esCompatPick } from './src/compat/object/pick';

const pickObj = { 'a.b': undefined, a: { b: 99 } };

console.log(loPick(pickObj, 'a.b')); // { 'a.b': undefined }
console.log(esCompatPick(pickObj, 'a.b')); // { 'a.b': 99 }
```

### DIFF
--- a/src/compat/object/get.spec.ts
+++ b/src/compat/object/get.spec.ts
@@ -28,6 +28,12 @@ describe('get', () => {
     expect(get(obj, 'a[].b')).toBe(undefined);
   });
 
+  it('should prioritize literal key over path when value is undefined', () => {
+    const object = { 'a.b': undefined, a: { b: 99 } };
+    expect(get(object, 'a.b')).toBe(undefined);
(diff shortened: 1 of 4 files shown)


raon0211:
Looking at the code, I think there's probably a bug in get too (since in pick we set the value we got from get), so we should fix that as well.

Antoliny0919:
> Looking at the code, I think there's probably a bug in get too (since in pick we set the value we got from get), so we should fix that as well.

Thanks for pointing that out! I'll take a look over the weekend 👍

raon0211 (approved):
Thanks!

PR #1789, opened 2026-06-23
fix(compat/dropRight): clean up JSDoc by removing duplicate overload and guard param

Antoliny0919:
## Summary

- Renamed the parameters(collection → array n → itemsCount), array seems more appropriate for an ArrayLike value, and itemsCount is chosen for consistency with drop.
- Removed the JSDoc related to guard.
I don't think this is something consumers of dropRight need to be aware of when using the function directly.
- Removed redundant JSDoc comments.


src/compat/array/dropRight.ts
```diff
@@ -10,40 +10,22 @@ import { toInteger } from '../util/toInteger.ts';
...
- * @param [guard] - Enables use as an iteratee for methods like `_.map`.
- * @returns A new array with the specified number of elements removed from the end.
- *
- * @example
- * const array = [1, 2, 3, 4, 5];
- * const result = dropRight(array, 2);
- * // result will be [1, 2, 3] since the last two elements are dropped.
- */
-export function dropRight<T>(collection: ArrayLike<T> | null | undefined, itemsCount = 1, guard?: unknown): T[] {
-  if (!isArrayLike(collection)) {
```
Antoliny0919, line 43: I noticed that most implementation signatures also include JSDoc. Is there a reason we document those as well?
When overload signatures are present, some of that information seems redundant to me. Am I missing something? 🤔

src/compat/array/dropRight.ts
```diff
@@ -10,40 +10,22 @@ import { toInteger } from '../util/toInteger.ts';
...
-/**
- * Removes a specified number of elements from the end of an array and returns the rest.
- *
- * This function takes an array and a number, and returns a new array with the specified number
- * of elements removed from the end.
- *
- * @template T - The type of elements in the array.
- * @param collection - The array from which to drop elements.
- * @param itemsCount - The number of elements to drop from the end of the array.
- * @param [guard] - Enables use as an iteratee for methods like `_.map`.
```
Antoliny0919, line 34: I noticed that many other places also expose the guard parameter in the JSDoc. Is there a reason for documenting guard explicitly?
As described in the docs, I understand that it's there to support cases where the function is passed as an iteratee to methods like map(for Lodash compability...), but I'm wondering if this is something users actually need to know about.
Would it be better to keep this as an internal implementation detail?

Antoliny0919 (commented):
I left a few comments on things I was curious about while making the changes, as I wanted to better understand the project 😉

src/compat/array/dropRight.ts
raon0211: Thanks for the cleanup! I reverted the removal of the JSDoc on the implementation signature. You're right that it's not strictly necessary — editors only surface the overload's JSDoc — but we keep it on the implementation as well for consistency with other compat functions like `take` and `takeRight`. The other changes (renaming `collection` to `array` and removing the `guard` param from the overload's JSDoc) look good.

raon0211 (approved):

PR #1795, opened 2026-06-23
docs(compat/each, compat/eachRight): mark iteratee as optional in heading

Antoliny0919:
## Summary

Missing optional flag in the heading 😂

### DIFF
--- a/docs/compat/reference/array/each.md
+++ b/docs/compat/reference/array/each.md
@@ -16,7 +16,7 @@ const result = each(collection, iteratee);
 
 ## Usage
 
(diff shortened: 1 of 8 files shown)


Antoliny0919:
It looks like there are quite a few cases where the optional flag is missing from the heading, like this one.
Would it be better to group similar types of functions together and open a PR with those changes, or should we fix all of them at once?
If we go with the latter, I think we would need to check all functions to make sure we don't miss any cases.

raon0211 (approved):
Thanks!

PR #1797, opened 2026-06-24
fix(compat/chunk): handle NaN and Infinity size values for compability

Antoliny0919:
## Summary

When `NaN` or `Infinity` is passed as the size value, an error is thrown.
In Lodash, passing `NaN` returns an empty array, while passing Infinity returns the [original array] unchanged.
The issue occurs because these values are passed directly to es-toolkit  chunk implementation without handling those edge cases first.

### NaN

```ts
import { chunk as loChunk } from 'lodash';
import { chunk as esChunk } from './src/compat/array/chunk';

const array = [0, 1, 2, 3, 4, 5];

console.log(loChunk(array, NaN)); // []

try {
  console.log(esChunk(array, NaN));
} catch (e) {
  console.log(e.message); // Size must be an integer greater than zero.
}
```

### Infinity

```ts
import { chunk as loChunk } from 'lodash';
import { chunk as esChunk } from './src/compat/array/chunk';

const array = [0, 1, 2, 3, 4, 5];

console.log(loChunk(array, Infinity)); // [[0,1,2,3,4,5]]

try {
  console.log(esChunk(array, Infinity));
} catch (e) {
  console.log(e.message); // Size must be an integer greater than zero.
}
```

### DIFF
--- a/src/compat/array/chunk.spec.ts
+++ b/src/compat/array/chunk.spec.ts
@@ -59,6 +59,14 @@ describe('chunk', () => {
     expect(chunk(args, 2)).toEqual([[1, 2], [3]]);
   });
 
+  it('should return an empty array when the size is NaN', () => {
+    expect(chunk([1, 2, 3], NaN)).toEqual([]);
+  });
+
+  it('should return an full elements array when the size is Infinity', () => {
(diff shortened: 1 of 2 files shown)


raon0211 (approved):
Thanks!

PR #1798, opened 2026-06-24
docs: use '?' notation for optional parameters across all language docs

Antoliny0919:
## Summary

Previously, optional parameters were indicated using the word "optional" in the parameter description.
While this approach communicates that a parameter is optional, it has several drawbacks:

- The word "optional" must be translated for each supported language.
- Users have to skim past the type text in order to notice that it's optional.
- Since parameters are displayed as code, the visual connection between the parameter and its optional status is weak.
- It is inconsistent with the way optional values are represented in headings.

This change replaces all existing optional parameter indicators with the JavaScript optional parameter syntax -> "?".

By adopting this approach, all of the issues above are resolved ✌️

If a concept can be represented using a single, universally recognizable notation within our target developer audience, it is generally preferable to use that notation instead of relying on localized text.

### en

optional -> ?

### ko

옵션 -> ?

### ja

オプション -> ?

### zh-hans

可选 -> ?


Antoliny0919:
스크린샷 2026-06-24 오후 5 35 48

The current representation of rest parameters is also inconsistent across the codebase 😅
I find the former approach much more readable.
If something can be expressed using standard JavaScript syntax, I think it's better to represent it that way whenever possible.

raon0211:
I prefer explicit explanations over special symbols or abbreviations, as they tend to be clearer. For example, `value?` is less obvious than writing `value` (optional), since `?` can carry multiple meanings. So I'd like to keep the current documentation style.

raon0211:
I think the latter is clearer for readers — since `values` and `iteratee` are inherently separate parameters, they should be separated explicitly. Let's make everything consistent with the latter style.

Antoliny0919:
> I prefer explicit explanations over special symbols or abbreviations, as they tend to be clearer. For example, `value?` is less obvious than writing `value` (optional), since `?` can carry multiple meanings. So I'd like to keep the current documentation style.

Thank you for pointing that out, @raon0211

I agree with your point that it's clearer.
One additional point I think is especially important is accessibility. Using ? could potentially cause accessibility issues, and I think I overlooked that aspect.

That said, what still feels odd to me is that whether a parameter is optional is read later than its type. To me, "optional" describes the parameter itself, it tells the reader that this parameter is optional. Because of that, it feels more natural for the optional indicator to appear immediately after the parameter name.
(It also won't improve readability when the type is long.)

>`predicate(optional)`: `ListIteratee<T>`: 
>
> The predicate function to apply to each element. Can accept a function, object pattern, array pattern, or property  name. Default is identity.

What do you think about that?

Antoliny0919:
> I think the latter is clearer for readers — since values and iteratee are inherently separate parameters, they should be separated explicitly. Let's make everything consistent with the latter style.

I think my point didn't come across clearly. 😄
Could you just focus on the values parameter?

What I meant is that the rest parameter values is represented in different ways throughout the documentation. My point was simply that we should use a consistent representation for rest parameters, and I personally prefer the former style since it matches standard JavaScript syntax.

raon0211:
Having "optional" right after the parameter name might make it feel more immediately descriptive. However, I think this is a relatively minor readability concern in practice. Readers scanning documentation tend to parse the full signature together, so the position of "optional" doesn't create a significant difference in comprehension.

If the impact on overall readability seems small, I'd prefer to keep the current style.

raon0211:
> What I meant is that the rest parameter values is represented in different ways throughout the documentation. My point was simply that we should use a consistent representation for rest parameters, and I personally prefer the former style since it matches standard JavaScript syntax.

Looking at the two representations, I personally find `...values` (`...ArrayLike<T>[]`) more intuitive as well, since the `...` prefix immediately signals that it's a rest parameter, which aligns with standard JavaScript syntax.

If you'd like to open a separate PR for this change, I'd be happy to review and merge it!

Antoliny0919:
> Having "optional" right after the parameter name might make it feel more immediately descriptive. However, I think this is a relatively minor readability concern in practice. Readers scanning documentation tend to parse the full signature together, so the position of "optional" doesn't create a significant difference in comprehension.
> 
> If the impact on overall readability seems small, I'd prefer to keep the current style.

If the benefit of making the change is small relative to the cost, there's no real reason to do it!
Thanks for sharing your thoughts.
Let's keep things as they are for now 🚀

Antoliny0919:
> Looking at the two representations, I personally find ...values (...ArrayLike<T>[]) more intuitive as well, since the ... prefix immediately signals that it's a rest parameter, which aligns with standard JavaScript syntax.
If you'd like to open a separate PR for this change, I'd be happy to review and merge it!

I agree. I'm heading off on a trip to Mongolia 😎 so I'm not sure exactly when I'll be able to work on it, but I'll try to get to it soon 👍

PR #1799, opened 2026-06-24
test[camelCase, capitalize]: refine string case tests

BangDori:
## Summary

Refines compat string case tests by removing a redundant camelCase double-conversion case and adding clearer capitalize coverage.

## Changes

- Removed the redundant camelCase(camelCase(...)) test.
- Added capitalize cases for:
  - lowercasing remaining uppercase characters
  - preserving leading whitespace
  - preserving an empty string

## Test

- `yarn vitest run src/compat/string/camelCase.spec.ts`
- `yarn vitest run src/compat/string/capitalize.spec.ts`

### DIFF
--- a/src/compat/string/camelCase.spec.ts
+++ b/src/compat/string/camelCase.spec.ts
@@ -46,6 +46,12 @@ describe('camelCase', () => {
     expect(camelCase('Москва')).toBe('москва');
   });
 
+  it('should handle double-converting strings', () => {
+    const actual = strings.map(str => camelCase(camelCase(str)));
+    const expected = strings.map(() => 'fooBar');
+    expect(actual).toEqual(expected);
+  });
+
   it('should remove contraction apostrophes', () => {
     const apostrophes = ["'", '\u2019'];
     const postfixes = ['d', 'll', 'm', 're', 's', 't', 've'];
--- a/src/compat/string/camelCase.spec.ts
+++ b/src/compat/string/camelCase.spec.ts
@@ -46,12 +46,6 @@ describe('camelCase', () => {
     expect(camelCase('Москва')).toBe('москва');
   });
 
-  it('should handle double-converting strings', () => {
-    const actual = strings.map(str => camelCase(camelCase(str)));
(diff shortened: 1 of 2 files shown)


src/compat/string/camelCase.spec.ts
```diff
@@ -46,12 +46,6 @@ describe('camelCase', () => {
     expect(camelCase('Москва')).toBe('москва');
   });
 
-  it('should handle double-converting strings', () => {
```
raon0211, line 49: These test cases are originally from the lodash library. To maintain 100% compatibility, I'd prefer to keep them as-is.

src/compat/string/capitalize.spec.ts
```diff
@@ -3,10 +3,23 @@ import type { capitalize as capitalizeLodash } from 'lodash';
 import { capitalize } from './capitalize';
 
 describe('capitalize', () => {
-  it('should capitalize the first character of a string', () => {
+  it('should capitalize the first character', () => {
```
raon0211: Same goes for the test descriptions — I'd rather not modify the originals. Feel free to add more, though.

src/compat/string/camelCase.spec.ts
BangDori: I initially thought this test was unnecessary because it calls camelCase twice, but I understand now that it comes from the original Lodash tests and helps preserve compatibility.

Thanks for the clarification! I’ve restored it.

[test(camelCase): restore lodash double conversion case](https://github.com/toss/es-toolkit/pull/1799/commits/f2b45f7f0a68cd83dda0acb9ecdc9abcf0daf8a5)

src/compat/string/capitalize.spec.ts
BangDori: I restored the original test description and kept the additional cases separate.

Thank you!

[test(capitalize): preserve original test description](https://github.com/toss/es-toolkit/pull/1799/commits/1155f017a2c02f9fe31c8504dfe67633d69d98f9)

raon0211 (approved):

PR #1808, opened 2026-06-26
fix(compat): don't treat an own literal dotted key as a deep path in unset (match #1775)

mahirhir:
`unset` (`compat/object/unset.ts`) is the unfixed sibling of #1775, which added the `&& !Object.hasOwn(object, path)` guard to `get` and `pick`. `unset` still has the bare `if (obj?.[path] === undefined && isDeepKey(path))`, so unsetting an own literal dotted key like `'a.b'` whose value is `undefined` wrongly deletes the nested `a.b` and leaves the literal key, diverging from lodash. `omit` delegates to `unset` and inherits this.

The fix is the one-token mirror of #1775's guard (`Object.hasOwn` is already used across 23 source files). Verified against the lodash 4.17.21 oracle: the old code mismatches on `{'a.b': undefined, a:{b:99}}` and `{'x.y': undefined}` (2 cases), the fix matches on all, and the non-undefined-literal and pure-deep-path cases are unchanged in both states.

### DIFF
--- a/src/compat/object/unset.spec.ts
+++ b/src/compat/object/unset.spec.ts
@@ -168,9 +168,6 @@ describe('unset', () => {
   });
 
   it('should prioritize a literal key over a path when the literal value is undefined', () => {
-    // Mirrors the literal-key-over-path fix applied to `get`/`pick` (#1775).
-    // When `'a.b'` exists as an own key whose value is `undefined`, lodash deletes
-    // that literal key and leaves the nested `a.b` untouched.
     const object: Record<string, unknown> = { 'a.b': undefined, a: { b: 99 } };
 
     expect(unset(object, 'a.b')).toBe(true);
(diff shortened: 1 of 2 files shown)


src/compat/object/unset.spec.ts
```diff
@@ -167,6 +167,24 @@ describe('unset', () => {
     expect(object).toEqual({ b: null });
   });
 
+  it('should prioritize a literal key over a path when the literal value is undefined', () => {
+    // Mirrors the literal-key-over-path fix applied to `get`/`pick` (#1775).
+    // When `'a.b'` exists as an own key whose value is `undefined`, lodash deletes
+    // that literal key and leaves the nested `a.b` untouched.
```
raon0211:
```suggestion
```

raon0211 (approved):
Thanks!

PR #1810, opened 2026-06-26
fix[has, hasIn]: prioritize nullish literal keys over paths

BangDori:
## Summary

Fixes `has` and `hasIn` to match Lodash when a literal key that looks like a deep path exists with a nullish value.

For string paths like `'a.b'`, the current implementation falls back to deep-path resolution when `object['a.b']` is `null` or `undefined`. This makes existing literal keys with nullish values behave as if they were missing.

Lodash decides whether a path is a literal key based on key existence, not on the value:

- `_.has({ 'a.b': undefined }, 'a.b') // true`
- `_.hasIn({ 'a.b': null }, 'a.b') // true`

## Changes

- Updated `has` to preserve nullish literal keys before falling back to deep path resolution.
- Updated `hasIn` with the same behavior, including inherited literal keys.
- Added regression tests for `undefined` and `null` literal keys that look like paths.

## References

- https://github.com/lodash/lodash/blob/4.17.21-npm/_isKey.js#L16-L27
- https://github.com/lodash/lodash/blob/4.17.21-npm/has.js#L1-L31
- https://github.com/lodash/lodash/blob/4.17.21-npm/hasIn.js#L1-L30
- https://github.com/lodash/lodash/blob/4.17.21-npm/_baseHas.js#L1-L16
- https://github.com/lodash/lodash/blob/4.17.21-npm/_baseHasIn.js#L1-L12

## Test results

- `yarn vitest run src/compat/object/has.spec.ts src/compat/object/hasIn.spec.ts`
- `yarn typecheck`


raon0211 (approved):
Hello, thanks for your contribution! I simplified the condition a little bit.

BangDori:
Thank you !

PR #1811, opened 2026-06-27
fix[add]: avoid mutating typed parameters during coercion

ramong26:
## Summary

Refactors the compat `add` implementation to avoid reassigning typed parameters during value coercion.

## Changes

* Introduced local `left` and `right` variables for internal coercion.
* Replaced inline `as any` assertions with internal `any` variables.
* Preserved the existing Lodash-compatible runtime behavior.

## Test results

* Not run (refactoring only).

### DIFF
--- a/src/compat/math/add.ts
+++ b/src/compat/math/add.ts
@@ -17,18 +17,21 @@ import { toString } from '../util/toString.ts';
  * const result3 = add(NaN, 10); // result3 will be NaN
  */
 export function add(value: number, other: number): number {
-  if (value === undefined && other === undefined) {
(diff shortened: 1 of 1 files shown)


raon0211:
Thanks for the PR! es-toolkit/compat is feature-complete and in maintenance mode — we only accept fixes for behavioral/bug inconsistencies with lodash, not refactoring-only changes. Since this PR preserves the existing behavior, I'll close it. If you find an actual behavioral mismatch in compat/add, please file an issue with a repro and we'll gladly look into it. 🙏

PR #1818, opened 2026-06-28
feat(types): add es-toolkit/types module with compile-time type utilities

dayongkr:
## Overview

Adds a declaration-only `es-toolkit/types` entrypoint for compile-time type utilities. v1 ships nine everyday helpers TypeScript lacks natively:

| Type | What it does |
|---|---|
| `ValueOf<T>` | Union of an object's value types (value-side `keyof`) |
| `Simplify<T>` | Flattens an intersection/mapped type into one readable object |
| `Merge<T, U>` | Merges two objects; `U` overrides overlapping keys |
| `SetOptional<T, K>` | Makes specific keys optional (scoped `Partial`) |
| `SetRequired<T, K>` | Makes specific keys required (scoped `Required`) |
| `Writable<T>` | Removes `readonly` from all properties (inverse of `Readonly`) |
| `NonEmptyArray<T>` | Array with at least one element (`[T, ...T[]]`) |
| `DeepPartial<T>` | Recursive `Partial` for nested object patches; plain objects, arrays/tuples (non-sparse), and `Map`/`Set` contents recurse; `Date`/`RegExp`/functions pass through |
| `DeepReadonly<T>` | Recursive `Readonly` for immutable state; arrays/tuples become readonly, `Map`/`Set` become `ReadonlyMap`/`ReadonlySet` |

## Selection rules

"Standard" = TypeScript native types (`lib.*.d.ts`).

1. **Native already provides it → skip** (`Partial`, `Omit`, `NonNullable`, `Awaited`, `Uppercase`, …).
2. **No stricter/gap-filling variants** — if something similar exists natively, use it (`Except`, `StrictOmit`, `Exact` skipped). Only fill what's genuinely missing.
3. **Stay consistent with the native spec** for related utilities (`ValueOf` mirrors `keyof`).

> Provide it only when TypeScript has nothing comparable, and shape it like the native types.

## Packaging

Declaration-only: `publishConfig.exports` exposes `./types` via the `types` condition (no runtime `default`), and `postbuild` strips the empty JS — ships only `.d.ts`/`.d.mts`. Use with `import type`.

### Deep utilities: where the special-casing stops

Deep utilities recurse into data containers (plain objects, arrays/tuples, `Map`/`Set`) and pass through atomic built-ins that appear in real data shapes (functions, `Date`, `RegExp`). Nothing else is special-cased (`WeakMap`, typed arrays, `Blob`, DOM types, class instances fall through to the object branch): observed demand for them is ~0, the fallout is mild, and adding a case later is non-breaking.

## Tests

Co-located `*.spec.ts` with Vitest `expectTypeOf` (existing convention). 9 files, 22 tests, passing under `--typecheck`. Specs assert intended design decisions only (including negative cases via `@ts-expect-error`, e.g. sparse arrays stay illegal under `DeepPartial`).

## Remaining TODO

- [x] Review the test code
- [x] Review the JSDoc descriptions
- [ ] Add the next high-priority types (by measured demand): `JsonValue` family → `UnionToIntersection` → `LiteralUnion` → `Brand` → `DeepRequired`

Note: the `isNonEmpty` guard (pairs with `NonEmptyArray`) belongs in `predicate` — separate PR.


dayongkr:
Delete

- Merge<T, U>
- SetOptional<T, K>
- SetRequired<T, K>

PR #1824, opened 2026-06-28
fix(compat/find): align implementation param names with overload signatures

Antoliny0919:
## Summary 

For consistency, how about using the same parameter names as the overload signatures? 👍


raon0211:
Thanks for the contribution! Some of the parameter renames here look good (e.g. `source` → `collection`), but we'd like to keep `doesMatch` as-is. It was intentionally named that way since we feel `predicate` is a bit too abstract. Could you revert just the `doesMatch` part and keep the rest? That would make this ready to merge!

raon0211 (approved):

PR #1827, opened 2026-06-28
fix(compat/findIndex): rename doesMatch parameter to predicate for consistency

Antoliny0919:
## Summary

I think predicate would be a more appropriate name than doesMatch. It would also make the parameter name consistent with find, findLast, and findLastIndex, which would improve consistency across the API.


raon0211:
Thanks for the suggestion! The `doesMatch` name was actually intentional. We felt that `predicate` is a bit too abstract for this context, since `doesMatch` more clearly communicates what the parameter actually does (i.e., a function that checks whether an element matches a condition). We'll keep the current naming for now. Appreciate the contribution!

PR #1841, opened 2026-07-05
feat (object/omit-deep): add 'omitDeep' function to omit fields deeply

kevaldonga:
## Summary

- Add `omitDeep` function that removes deeply nested object fields using dot-separated paths (e.g. `'user.profile.email'`)
- Supports top-level keys, nested objects, and field omission across array items
- Strong TypeScript inference via `Paths<T>` for valid paths and `OmitDeep<T, P>` for the return type
- 17 test cases including nested objects, arrays, entire field omission, primitives, and type checks
- Docs in all 4 languages (en, ko, ja, zh_hans)

## Test Plan

- [x] Top-level property omission
- [x] Nested path omission (`'b.x'`, `'user.profile.email'`)
- [x] Multiple paths at once
- [x] Entire nested object omitted (`'user.profile'`, `'b'`)
- [x] Property omitted from objects inside arrays (`'users.secret'`)
- [x] Entire array field omitted (`'users'`)
- [x] Nested arrays (`'items.y'`)
- [x] Empty paths, all keys omitted, original object not mutated
- [x] Empty objects/arrays and non-existent nested paths (runtime no-op, type error)
- [x] Primitives (`null`, `undefined`, numbers, strings, booleans) unchanged
- [x] Type tests with `expectTypeOf().toExtend()`
- [x] `yarn vitest run src/object/omitDeep --typecheck` passes (17 tests, no type errors)

### DIFF
--- a/docs/ja/reference/object/omitDeep.md
+++ b/docs/ja/reference/object/omitDeep.md
@@ -1,0 +1,74 @@
+# omitDeep
+
+指定されたネストされたパスを除外した新しいオブジェクトを返します。
+
+```typescript
+const result = omitDeep(object, paths);
+```
+
+## 使用法
+
+### `omitDeep(object, paths)`
+
+オブジェクトから特定のネストされたプロパティを除外したい時に`omitDeep`を使用してください。ドット(`.`)で区切られたパスに対応するプロパティを削除した新しいオブジェクトを返します。ネストされたオブジェクトと配列内のオブジェクトも再帰的に処理されます。
+
+```typescript
+import { omitDeep } from 'es-toolkit/object';
+
+// ネストされたプロパティを除外
+const obj = { a: 1, b: { x: 2, y: 3 }, c: 4 };
+const result = omitDeep(obj, ['b.x']);
+// resultは{ a: 1, b: { y: 3 }, c: 4 }になります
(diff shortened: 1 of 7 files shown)


kevaldonga:
@raon0211 @dayongkr 
Please review this PR, suggest me any improvements on this feature. 
would like to have feedback for this.

dayongkr:
Thanks for the contribution!

es-toolkit asks for a discussion before a new function lands:

> **Open a discussion first and wait until it is accepted.** A pull request that adds a function without one will be closed, even when the implementation itself is good.

([CONTRIBUTING §4.1](https://github.com/toss/es-toolkit/blob/main/.github/CONTRIBUTING.md#adding-a-new-function))

There isn't one for `omitDeep`, so I'm closing this. It's a process call, not a verdict on the code.

To be upfront: adding new functions isn't a high priority for us at the moment, so a discussion may sit for a while before we can take it up. If you'd still like to open one, [Ideas](https://github.com/toss/es-toolkit/discussions/new?category=ideas) is the place. Worth addressing there: `compat`'s `omit` already accepts deep paths (`omit(obj, [['a', 'b']])`), so what `omitDeep` adds beyond it.

PR #1844, opened 2026-07-06
refactor(compat/every): rename 'source' param to 'collection'

Antoliny0919:


Antoliny0919:
```ts
export function every<T>(
  source: ArrayLike<T> | Record<any, any> | null | undefined,
  doesMatch?: ((item: T, index: number, arr: any) => unknown) | Partial<T> | [keyof T, unknown] | PropertyKey,
  guard?: unknown
):
```
스크린샷 2026-07-06 오후 9 46 25

The parameter name used in the doesMatch documentation doesn't match the one used in the implementation signature. Would it make sense to standardize this as well (`arr` -> `collection`)?

raon0211:
Agreed, I think it makes sense to standardize this as you suggested — using `collection` and `doesMatch` consistently across the signature and docs.

Antoliny0919:
Thank you @raon0211

> Agreed, I think it makes sense to standardize this as you suggested — using collection and doesMatch consistently across the signature and docs.

I agree as well, it seems like the docs should switch to `doesMatch` too.

Also, to add to what I asked earlier, I don't think my intent came across clearly, so let me clarify:

my-think

For one of the `doesMatch` types, `((item: T, index: number, collection: any) => unknown)`, the docs use `collection` for the third parameter, but the code uses `arr`.
What I was actually asking was whether we should unify this to `collection` :)

raon0211 (approved):

PR #1846, opened 2026-07-06
fix(compat/difference, compat/differenceBy): normalize -0 to 0 in results

Antoliny0919:
## Summary

Lodash returns 0 instead of -0, regardless of whether the value comes from removing differences. We had a compatibility difference in this behavior.

### difference

```ts
import { difference as loDifference } from 'lodash';
import { difference as esDifference } from '../../src/compat/array/difference';

console.log(loDifference([-0, 1], [1])) // [0]
console.log(esDifference([-0, 1], [1])) // [-0]
```

### differenceBy

```ts
import { differenceBy as loDifferenceBy } from 'lodash';
import { differenceBy as esDifferenceBy } from '../../src/compat/array/differenceBy';

console.log(loDifferenceBy([-0, 1], [1]));
console.log(esDifferenceBy([-0, 1], [1]));
```


raon0211:
Closing because this was fixed in https://github.com/toss/es-toolkit/pull/2007.

PR #1847, opened 2026-07-07
fix(compat/flattenDepth): move flattening logic from flatten to flattenDepth

Antoliny0919:
## Summary

Previously, `flatten` behaved the same as `flattenDepth`.
Remove the depth parameter from `flatten` and delegate its behavior to `flattenDepth`.

### flatten

- Remove the `depth` parameter.
- Remove the internal implementation and delegate all behavior to `flattenDepth`.

### flattenDeep

- Rename the parameter to array.( for consistency)

### flattenDepth

It previously acted as a wrapper around `flatten` 😂
- Move the actual flattening logic from `flatten` into `flattenDepth`.

### DIFF
--- a/docs/ko/compat/reference/array/flatten.md
+++ b/docs/ko/compat/reference/array/flatten.md
@@ -11,14 +11,14 @@
 배열을 한 단계 평탄화해요.
 
 ```typescript
-const result = flatten(array, depth);
+const result = flatten(array);
 ```
 
 ## 사용법
 
-### `flatten(value, depth)`
+### `flatten(array)`
 
-중첩 배열을 지정된 깊이만큼 평탄화해요. 기본적으로 한 단계만 평탄화하며, Arguments 객체나 Symbol.isConcatSpreadable을 가진 객체도 지원해요.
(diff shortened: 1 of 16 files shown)


docs/ko/compat/reference/array/flatten.md
```diff
@@ -11,14 +11,14 @@
...
+const result = flatten(array);
 ```
 
 ## 사용법
 
-### `flatten(value, depth)`
+### `flatten(array)`
 
-중첩 배열을 지정된 깊이만큼 평탄화해요. 기본적으로 한 단계만 평탄화하며, Arguments 객체나 Symbol.isConcatSpreadable을 가진 객체도 지원해요.
+중첩 배열을 한 단계 평탄화해요.
```
Antoliny0919, line 21: Do you think the previous version was better? It feels like this is completely duplicating what's already covered here, which gives me a bit of pause. I think it would be better if the message were a bit more general and extensible.

Image

raon0211 (approved):
Thanks!

PR #1849, opened 2026-07-07
refactor(compat/flatMap): delegate to flatMapDepth

Antoliny0919:
## Summary

I think `flatMap` can delegate directly to `flatMapDepth` without handling nil or performing the mapping itself.


src/compat/array/flatMap.ts
```diff
@@ -114,14 +111,6 @@ export function flatMap(collection: object | null | undefined, iteratee: object)
...
-  if (isNil(collection)) {
-    return [];
-  }
-
-  // eslint-disable-next-line
-  // @ts-ignore
-  const mapped = isNil(iteratee) ? map(collection) : map(collection, iteratee);
-
-  return flattenDepth(mapped, 1) as R[];
+export function flatMap<R = any>(collection: object | null | undefined, iteratee?: any = identity): R[] {
```
raon0211:
```suggestion
export function flatMap<R = any>(collection: object | null | undefined, iteratee: any = identity): R[] {
```

raon0211 (dismissed):

src/compat/array/flatMap.ts
```diff
@@ -115,13 +112,5 @@ export function flatMap(collection: object | null | undefined, iteratee: object)
  * // => [1, 2, 2, 4]
  */
 export function flatMap<R = any>(collection: object | null | undefined, iteratee: any = identity): R[] {
```
raon0211:
```suggestion
export function flatMap<R = any>(collection: object | null | undefined, iteratee? any): R[] {
```

raon0211 (approved):

PR #1850, opened 2026-07-07
docs(compat/flatMap): add default iteratee value across all languages

Antoliny0919:
## Summary

The behavior looks correct, but it seems like the default value for iteratee is missing 😅

스크린샷 2026-07-07 오후 6 57 34

### DIFF
--- a/docs/compat/reference/array/flatMap.md
+++ b/docs/compat/reference/array/flatMap.md
@@ -70,7 +70,7 @@ flatMap(users, { active: false });
 #### Parameters
 
 - `collection` (`object | null | undefined`): The collection to iterate over. Can be an array, object, or string.
-- `iteratee` (`ListIterator | ObjectIterator | string | object`, optional): The iteratee to apply to each element. Can be a function, property name, or partial object. Default is `identity`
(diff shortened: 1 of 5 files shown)


docs/compat/reference/array/flatMap.md
```diff
@@ -70,7 +70,7 @@ flatMap(users, { active: false });
 #### Parameters
 
 - `collection` (`object | null | undefined`): The collection to iterate over. Can be an array, object, or string.
-- `iteratee` (`ListIterator | ObjectIterator | string | object`, optional): The iteratee to apply to each element. Can be a function, property name, or partial object.
+- `iteratee` (`ListIterator | ObjectIterator | string | object`, optional): The iteratee to apply to each element. Can be a function, property name, or partial object. Default is `identity`
```
raon0211:
```suggestion
- `iteratee` (`ListIterator | ObjectIterator | string | object`, optional): The iteratee to apply to each element. Can be a function, property name, or partial object. Defaults to `identity`.
```

raon0211 (approved):

docs/compat/reference/array/flatMap.md
Antoliny0919: oops .. Thank you 🙏

PR #1852, opened 2026-07-08
fix(compat): count code points when padding multi-byte strings

spokodev:
`pad`, `padStart` and `padEnd` delegate to the native `String.prototype.pad*` methods and `String#length`, which count UTF-16 code units. For strings with astral (surrogate-pair) characters this diverges from lodash: the pad width is wrong, and padding *with* an astral character splits its surrogate pair into a lone surrogate.

```js
pad('ab', 8, '😀')      // '😀\ud83dab😀\ud83d'  (expected '😀😀😀ab😀😀😀')
padStart('abc', 6, '😀') // '😀\ud83dabc'          (expected '😀😀😀abc')
```

Count and slice padding by Unicode code points via a shared `createPadding` helper, matching lodash and the existing `truncate` handling (same `regexMultiByte` guard). Strings without multi-byte characters keep the `String#length` fast path.


raon0211 (approved):
Thanks!

PR #1853, opened 2026-07-08
fix(compat): count string size by code points for multi-byte characters

spokodev:
`size` returns `String#length` for strings — the number of UTF-16 code units — so a string with astral characters is over-counted:

```js
size('😀😁😂') // 6  (expected 3)
size('a😀b')   // 4  (expected 3)
```

Count strings by Unicode code points, matching lodash and the existing `truncate` handling. Strings without multi-byte characters keep using `length` to avoid the extra allocation.


raon0211 (approved):
Thanks!

PR #1869, opened 2026-07-10
fix(compat/forEach): remove redundant array check in key resolution

Antoliny0919:
Since arrays and array like objects are handled the same way, I think `isArrayLike` alone should be sufficient.

### DIFF
--- a/src/compat/array/forEach.ts
+++ b/src/compat/array/forEach.ts
@@ -152,8 +152,7 @@ export function forEach<T>(
     return collection;
   }
 
-  const keys: PropertyKey[] =
-    isArrayLike(collection) || Array.isArray(collection) ? range(0, collection.length) : Object.keys(collection);
+  const keys: PropertyKey[] = isArrayLike(collection) ? range(0, collection.length) : Object.keys(collection);
 
   for (let i = 0; i < keys.length; i++) {
     const key = keys[i];


src/compat/array/forEach.ts
```diff
@@ -152,8 +152,7 @@ export function forEach<T>(
     return collection;
   }
```
Antoliny0919, line 153: It looks like this is only intended to handle `null` and `undefined`.

With the current implementation, it would also treat other falsy values as empty, which seems broader than intended.

Would it make sense to explicitly handle only null and undefined instead?

```diff --git a/src/compat/array/forEach.ts b/src/compat/array/forEach.ts
index f0f6fb48..8aa3bbcf 100644
--- a/src/compat/array/forEach.ts
+++ b/src/compat/array/forEach.ts
@@ -144,7 +144,7 @@ export function forEach<T>(
   collection: ArrayLike<T> | Record<any, any> | string | null | undefined,
   callback: (item: any, index: any, arr: any) => unknown = identity
 ): ArrayLike<T> | Record<any, any> | string | null | undefined {
-  if (!collection) {
+  if (collection == null) {
     return collection;
   }
```

One thing I'm curious about: do you have a preference between using a raw `collection == null` check and using isNil for this kind of logic?

Personally, I think using `isNil` is perfectly reasonable. In fact, I'd be happy to standardize on `isNil` for `null/undefined` checks going forward.

```diff --git a/src/compat/array/forEach.ts b/src/compat/array/forEach.ts
index f0f6fb48..a1a93f68 100644
--- a/src/compat/array/forEach.ts
+++ b/src/compat/array/forEach.ts
@@ -1,5 +1,6 @@
 import { identity } from '../../function/identity.ts';
 import { range } from '../../math/range.ts';
+import { isNil } from '../../predicate/isNil.ts';
 import { ArrayIterator } from '../_internal/ArrayIterator.ts';
 import { ListIterator } from '../_internal/ListIterator.ts';
 import { ObjectIterator } from '../_internal/ObjectIterator.ts';
@@ -144,7 +145,7 @@ export function forEach<T>(
   collection: ArrayLike<T> | Record<any, any> | string | null | undefined,
   callback: (item: any, index: any, arr: any) => unknown = identity
 ): ArrayLike<T> | Record<any, any> | string | null | undefined {
-  if (!collection) {
+  if (isNil(collection)) {
     return collection;
   }
```

Using `isNil` does introduce a dependency on that utility, but I think there are both pros and cons to that. Since `isNil` is such a fundamental utility that's unlikely to change, I also feel it makes the intent more explicit.

PR #1870, opened 2026-07-10
fix(compat/forEach): fix outdated JSDoc param name and description

Antoliny0919:


src/compat/array/forEach.ts
```diff
@@ -130,12 +130,8 @@ export function forEach<T extends object>(
  * Iterates over each element of the object invoking the provided callback function for each property.
  *
  * @template T - The type of object.
- * @param object - The object to iterate over.
- * @param [callback] - The function invoked for each property.
- * The callback function receives three arguments:
- *  - 'value': The current property being processed in the object.
- *  - 'key': The key of the current property being processed in the object.
- *  - 'object': The object 'forEach' was called upon.
```
Antoliny0919, line 138: I'm not sure this explanation is necessary 🤔
I think the signature alone is enough to make the behavior clear.

Antoliny0919:
Since the JSDoc updates don't seem urgent at the moment, I'll close this for now. Thanks!

PR #1871, opened 2026-07-10
docs: point alias docs to canonical function

Antoliny0919:
## Summary

The documentation for aliased functions is currently quite inconsistent. It would be better if both documents used the same description rather than providing different explanations.


docs/compat/reference/array/each.md
```diff
@@ -1,58 +1,3 @@
...
-
-#### Parameters
-
-- `collection` (`ArrayLike<T> | Record<any, any> | string | null | undefined`): The collection to iterate over.
-- `iteratee` (`(item: any, index: any, collection: any) => unknown`, optional): The function to execute for each element. Default is the `identity` function.
-
-#### Returns
-
-(`ArrayLike<T> | Record<any, any> | string | null | undefined`): Returns the original collection.
+`each` is an alias of [`forEach`](./forEach.md). See the [`forEach`](./forEach.md) documentation for details.
```
Antoliny0919, line 7: I removed the detailed documentation from the aliased function.

Using `each` and `forEach` as an example, they provide the same functionality, so I think their documentation should remain consistent as well. If both documents contain detailed explanations, any future changes will need to be made in both places.

If one of them is updated without the other being reviewed, we could easily end up in a situation where two aliases provide different documentation for the same behavior.

For that reason, I removed the detailed description from `each` and had it refer to the `forEach` documentation instead.

Previously, both functions had detailed documentation. Was there a particular reason for that?

docs/compat/reference/array/forEach.md
```diff
@@ -76,11 +76,15 @@ forEach(undefined, value => {
...
+#### Alias
+
+`each`
+
 #### Parameters
 
-- `collection` (`ArrayLike<T> | Record<string, unknown> | null | undefined`): The array or object to iterate over.
-- `callback` (`(value: T, index: number | string, collection: any) => void | false`): The function to execute for each element. Returns `false` to stop iteration.
+- `collection` (`ArrayLike<T> | Record<any, any> | string | null | undefined`): The array or object to iterate over.
+- `callback` (`(item: any, index: any, collection: any) => unknown`, optional): The function to execute for each element. Default is the `identity` function.
```
Antoliny0919, line 86: noticed that the types documented in `each` were written correctly, so I applied that to `forEach`. Going forward, when consolidating docs for other aliases beyond `each` <-> `forEach`, what do you think about just taking whichever version is more accurate between the two and consolidating them into one, rather than keeping both?

docs/compat/reference/array/forEach.md
```diff
@@ -76,11 +76,15 @@ forEach(undefined, value => {
 });
 ```
 
+#### Alias
+
+`each`
+
```
Antoliny0919, line 82: I also added an Alias section to the `forEach` documentation to make it clear that each is an alias of `forEach`.
(Lodash includes a similar section, which I think helps with discoverability as well.)

I'm not sure whether this is the best place for the section. If you think there's a more appropriate location, please let me know!

Antoliny0919 (commented):
I'm planning to consolidate the documentation for functions that have aliases.

You can find the reasoning behind this in the comment below.

I've put together one example of what the consolidated documentation would look like. Would you be willing to help with the remaining alias functions once we agree on the approach?

### Functions that require documentation consolidation

- each, forEach
- eachRight, forEachRight
- first, head
- extend, assignIn
- extendWith, assignInWith
- entries, toPairs
- entriesIn, toPairsIn

(Just as a side note, it looks like `entries` and `entriesIn` aren't implemented yet 🥲)

raon0211 (dismissed):
Thanks!

raon0211 (approved):

PR #1873, opened 2026-07-10
test(compat/each): replace each duplicate tests with an alias check against forEach

Antoliny0919:
## Summary

We could delegate the specific test cases for the alias each to `forEach`. Since `forEach` is already thoroughly tested and `each` is simply an alias of it, that should be sufficient.


src/compat/array/each.spec.ts
```diff
@@ -1,246 +1,9 @@
...
-    const values = [];
-
-    func(object, (value, other) => {
-      values.push(Array.isArray(value) ? other : value);
-      return false;
-    });
-
-    expect(values.length).toBe(1);
+  it('should be an alias of forEach', () => {
+    expect(each).toBe(forEach);
```
Antoliny0919, line 7: I noticed that alias functions didn't have dedicated test files before. How about keeping the test file, as it is now, but only adding a test that verifies the alias references the same implementation as the original function?

raon0211 (approved):
Thanks!

src/compat/array/each.spec.ts
raon0211: As far as I understand, that's actually already what this PR does.

`each.spec.ts` stays right where it is; we just trimmed it down to a single test confirming `each` is the same implementation as `forEach`:

```ts
describe('each', () => {
  it('should be an alias of forEach', () => {
    expect(each).toBe(forEach);
  });
});
```

Only the duplicated behavioral cases were removed, since `forEach.spec.ts` already covers those. Let me know if you meant something a bit different, happy to adjust!

Antoliny0919: > As far as I understand, that's actually already what this PR does.
> 
> `each.spec.ts` stays right where it is; we just trimmed it down to a single test confirming `each` is the same implementation as `forEach`:
> 
> ```ts
> describe('each', () => {
>   it('should be an alias of forEach', () => {
>     expect(each).toBe(forEach);
>   });
> });
> ```
> 
> Only the duplicated behavioral cases were removed, since `forEach.spec.ts` already covers those. Let me know if you meant something a bit different, happy to adjust!

I actually meant asking whether we should add the same kind of test for the other aliases too, like `first`, `extend` ...
It looks like this is already being worked on :) -> https://github.com/toss/es-toolkit/pull/1883

PR #1875, opened 2026-07-11
fix(compat/forEachRight): fix outdated JSDoc param name and description

Antoliny0919:


raon0211:
Thanks for the cleanup!

Quick question, in what way was this outdated? The collection rename looks right, but the callback argument description (value/key/object) also got removed, and that part still seems accurate. 

Also, just FYI, JSDoc-only fixes aren't a priority for us right now.

Antoliny0919:
> Thanks for the cleanup!
> 
> Quick question, in what way was this outdated? The collection rename looks right, but the callback argument description (value/key/object) also got removed, and that part still seems accurate.

The "outdated" part was about the parameter name (`object` → `collection`). I removed the `callback` detail because I hadn't seen other JSDoc comments include that kind of parameter specific detail (especially around `iteratee`).
My main intent was consistency, and I also wondered whether that level of detail was really necessary, since I felt it could already be inferred from the types.

> Also, just FYI, JSDoc-only fixes aren't a priority for us right now.

Agreed. Unless something is clearly wrong, I'll focus more on the parts that matter most :)

PR #1877, opened 2026-07-11
fix(compat/groupBy): rename parameters to match docs for consistency

Antoliny0919:
## Summay

스크린샷 2026-07-11 오후 1 47 45

I think we should align the parameter name with the documentation for consistency :)


raon0211:
Thanks for opening this PR! However, we don't currently consider argument name changes a high priority. We'll batch these kinds of changes together and address them all at once later.

I'll go ahead and close this PR for now. Thanks for understanding.

PR #1878, opened 2026-07-11
docs(compat/groupBy): remove incorrect default value from iteratee jsdoc

Antoliny0919:


raon0211:
Looking at the implementation, the iteratee effectively does default to identity:

```ts
const getKeyFromItem = createIteratee(_getKeyFromItem ?? identity);
```

So groupBy([6.1, 4.2, 6.3]) behaves the same as groupBy([6.1, 4.2, 6.3], identity), which is what [iteratee=identity] was describing. Could you share what made you consider this change? I'd like to understand the reasoning — was there a case where the JSDoc seemed misleading?

Antoliny0919:
> Looking at the implementation, the iteratee effectively does default to identity:
> 
> ```ts
> const getKeyFromItem = createIteratee(_getKeyFromItem ?? identity);
> ```
> 
> So groupBy([6.1, 4.2, 6.3]) behaves the same as groupBy([6.1, 4.2, 6.3], identity), which is what [iteratee=identity] was describing. Could you share what made you consider this change? I'd like to understand the reasoning — was there a case where the JSDoc seemed misleading?

When looking only at the signature, my intention was to use `[]` for optional parameters and `[x=some]` when a default value exists.
However, after thinking about it again, I feel that `[iteratee=identity]` provides more useful information to users, since when this overload is matched, the runtime default for `iteratee` is `identity` anyway.

Also, every currently uses the same form as `groupBy` did before my current commit, while `filter` and `reject` use the same form as the change in my current commit. The lack of consistency here also made it difficult for me to decide which direction to take.

### every

https://github.com/toss/es-toolkit/blob/171d8a8a5941671d3b8e99ba617bfb27cf2c71b3/src/compat/array/every.ts#L10-L40

### filter

https://github.com/toss/es-toolkit/blob/171d8a8a5941671d3b8e99ba617bfb27cf2c71b3/src/compat/array/filter.ts#L13-L21

https://github.com/toss/es-toolkit/blob/171d8a8a5941671d3b8e99ba617bfb27cf2c71b3/src/compat/array/filter.ts#L116-L119

Thank you for pointing this out @raon0211 !

Considering that you mentioned in another PR that JSDoc-related work is a lower priority, I think it would be better to first review and improve other areas related to compatibility.

It also would have been better if I had explained my intention in the description. Thank you, as always, for your reviews! I've been really happy lately because I get to enjoy both improving and learning along the way.

PR #1883, opened 2026-07-12
test(compat): add missing type and alias tests

bel1c10ud:
## Summary

Adds missing type and alias tests for `es-toolkit/compat`.

## Changes

- Added alias tests for `eachRight`, `first`, `extend`, and `extendWith`.
- Added missing Lodash type compatibility checks for `eachRight`, `fill`, `first`, `remove`, `reverse`, `extend`, and `extendWith`.
- Added missing comments explaining why some compat exports are not checked with `expectTypeOf`.

### DIFF
--- a/src/compat/array/first.spec.ts
+++ b/src/compat/array/first.spec.ts
@@ -1,0 +1,9 @@
+import { describe, expect, it } from 'vitest';
+import { first } from './first';
+import { head } from './head';
+
+describe('first', () => {
+  it('should be an alias of head', () => {
+    expect(first).toBe(head);
+  });
+});
--- a/src/compat/array/eachRight.spec.ts
+++ b/src/compat/array/eachRight.spec.ts
@@ -1,0 +1,9 @@
+import { describe, expect, it } from 'vitest';
+import { eachRight } from './eachRight';
+import { forEachRight } from './forEachRight';
+
+describe('eachRight', () => {
+  it('should be an alias of forEachRight', () => {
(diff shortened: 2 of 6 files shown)


src/compat/array/first.spec.ts
```diff
@@ -0,0 +1,9 @@
+import { describe, expect, it } from 'vitest';
+import { first } from './first';
+import { head } from './head';
+
+describe('first', () => {
+  it('should be an alias of head', () => {
+    expect(first).toBe(head);
+  });
+});
```
Antoliny0919, line 9: https://github.com/toss/es-toolkit/blob/171d8a8a5941671d3b8e99ba617bfb27cf2c71b3/src/compat/array/head.spec.ts#L26-L28

It looks like the same test already exists in the head tests as well. What do you think about removing it from head?

Also, it looks like the head tests are testing first too 😅 -> https://github.com/toss/es-toolkit/pull/1884

Antoliny0919 (commented):
Welcome @bel1c10ud  😏

I left small suggestion !

src/compat/array/first.spec.ts
bel1c10ud: Hello @Antoliny0919 👋

Oh, I missed an important point there. 😮 

I agree that we should remove the duplicate alias test from `head.spec.ts`.

I also checked out #1884. Thanks for addressing that! 👍

raon0211 (approved):

PR #1884, opened 2026-07-12
test(compat/head): replace first alias usage with head in tests

Antoliny0919:
## Summary

The `head` tests are currently using its alias, `first`, to test the behavior. :(
I think we should test head directly, while `first` only needs a test to verify that it behaves the same as `head`.


raon0211 (approved):
Thanks!

PR #1885, opened 2026-07-13
fix(isEqual): correctly compare Maps with structurally-equal object keys

### LINKED ISSUE #1881: isEqual: false negative on equal Maps/Sets with duplicate-shaped keys (deviates from lodash)
`isEqual` returns `false` for genuinely equal Maps when multiple keys are
structurally equal — and deviates from lodash, which returns `true`:

```ts
import { isEqual } from 'es-toolkit';
import lodashIsEqual from 'lodash/isEqual';

const A = new Map([[{}, 1], [{}, 2]]);
const B = new Map([[{}, 2], [{}, 1]]);
// Both maps hold the same multiset of entries: ({} → 1) and ({} → 2).

isEqual(A, B);       // false ❌
lodashIsEqual(A, B); // true  ✅
```

The cause is greedy key matching: each key of `A` is paired with the first
structurally-equal key of `B` and never reconsidered, so when that first
pairing has mismatched values the comparison fails even though a valid
pairing exists. Sets with duplicate-shaped elements hit the same path.

In an 8,000-pair differential fuzz over random Map/Set/object/array data
(adjudicated by an exhaustive backtracking matcher), roughly 10% of
genuinely-equal collection pairs came back `false` from this class — it
fires whenever a collection contains two entries whose keys are structurally
equal but whose values differ per pairing.

Deciding these cases correctly requires backtracking (or grouping equal keys
and matching values as a multiset) rather than first-match-wins. Happy to
provide the fuzz corpus or more failing cases if useful.

xianjianlf2:
## Summary

Closes #1881

`isEqual` could return `false` for equal `Map` instances when their keys were different object references but structurally equal. This also affected `es-toolkit/compat`, which reuses the same implementation and should match Lodash for this behavior.

## Changes

- Compare `Map` entries by deep equality for both keys and values instead of relying on `Map.has()` / `Map.get()` reference lookup.
- Remove each matched entry from the remaining candidates so Maps with duplicate-shaped object keys can still match in a different insertion order.
- Add regression coverage for structurally equal object keys, duplicate-shaped keys with reordered values, and mismatched values.
- Update the `isEqual` bundle-size snapshot for the implementation change.

## Test results

- `yarn vitest run src/predicate/isEqual.spec.ts src/compat/predicate/isEqual.spec.ts`
- `yarn workspace benchmarks vitest run bundle-size/isEqual.spec.ts`

### DIFF
--- a/src/predicate/isEqual.spec.ts
+++ b/src/predicate/isEqual.spec.ts
@@ -112,4 +112,50 @@ describe('isEqual', () => {
 
     expect(isEqual(buffer3, buffer4)).toBe(false);
   });
+
+  it('should return true for equal Maps with structurally-equal object keys', () => {
+    const map1 = new Map<object, number>([
+      [{ a: 1 }, 1],
+      [{ b: 2 }, 2],
+    ]);
+    const map2 = new Map<object, number>([
+      [{ a: 1 }, 1],
+      [{ b: 2 }, 2],
+    ]);
+
+    expect(isEqual(map1, map2)).toBe(true);
+  });
+
+  it('should return true for equal Maps with duplicate-shaped keys but differing values per pairing', () => {
+    const map1 = new Map<object, number>([
+      [{}, 1],
+      [{}, 2],
+    ]);
+    const map2 = new Map<object, number>([
+      [{}, 2],
+      [{}, 1],
+    ]);
+
+    expect(isEqual(map1, map2)).toBe(true);
+  });
+
+  it('should return false for Maps with structurally-equal keys but mismatched values', () => {
+    const map1 = new Map<object, number>([
+      [{}, 1],
+      [{}, 2],
+    ]);
+    const map2 = new Map<object, number>([
+      [{}, 1],
+      [{}, 3],
+    ]);
+
+    expect(isEqual(map1, map2)).toBe(false);
+  });
+
+  it('should return true for equal Sets with duplicate-shaped elements in different order', () => {
+    const set1 = new Set<object>([{ a: 1 }, { a: 1, b: 2 }]);
+    const set2 = new Set<object>([{ a: 1, b: 2 }, { a: 1 }]);
+
+    expect(isEqual(set1, set2)).toBe(true);
+  });
 });
--- a/src/predicate/isEqual.spec.ts
+++ b/src/predicate/isEqual.spec.ts
@@ -151,5 +151,4 @@ describe('isEqual', () => {
 
     expect(isEqual(map1, map2)).toBe(false);
   });
-
 });
--- a/src/predicate/isEqual.spec.ts
+++ b/src/predicate/isEqual.spec.ts
@@ -126,7 +126,7 @@ describe('isEqual', () => {
     expect(isEqual(map1, map2)).toBe(true);
   });
 
-  it('should return true for equal Maps with duplicate-shaped keys but differing values per pairing', () => {
+  it('should return true for equal Maps with duplicate-shaped elements in different order', () => {
     const map1 = new Map<object, number>([
       [{}, 1],
       [{}, 2],
--- a/src/predicate/isEqual.spec.ts
+++ b/src/predicate/isEqual.spec.ts
@@ -152,10 +152,4 @@ describe('isEqual', () => {
     expect(isEqual(map1, map2)).toBe(false);
   });
 
-  it('should return true for equal Sets with duplicate-shaped elements in different order', () => {
-    const set1 = new Set<object>([{ a: 1 }, { a: 1, b: 2 }]);
-    const set2 = new Set<object>([{ a: 1, b: 2 }, { a: 1 }]);
-
-    expect(isEqual(set1, set2)).toBe(true);
-  });
 });
--- a/benchmarks/bundle-size/isEqual.spec.ts
+++ b/benchmarks/bundle-size/isEqual.spec.ts
@@ -9,11 +9,11 @@ describe('isEqual bundle size', () => {
 
   it('es-toolkit', async () => {
     const bundleSize = await getBundleSize('es-toolkit', 'isEqual');
-    expect(bundleSize).toMatchInlineSnapshot(`3477`);
+    expect(bundleSize).toMatchInlineSnapshot(`3725`);
   });
 
   it('es-toolkit/compat', async () => {
     const bundleSize = await getBundleSize('es-toolkit/compat', 'isEqual');
-    expect(bundleSize).toMatchInlineSnapshot(`3477`);
+    expect(bundleSize).toMatchInlineSnapshot(`3725`);
   });
 });
--- a/benchmarks/bundle-size/isEqual.spec.ts
+++ b/benchmarks/bundle-size/isEqual.spec.ts
@@ -9,11 +9,11 @@ describe('isEqual bundle size', () => {
 
   it('es-toolkit', async () => {
     const bundleSize = await getBundleSize('es-toolkit', 'isEqual');
-    expect(bundleSize).toMatchInlineSnapshot(`3357`);
+    expect(bundleSize).toMatchInlineSnapshot(`3477`);
   });
 
   it('es-toolkit/compat', async () => {
     const bundleSize = await getBundleSize('es-toolkit/compat', 'isEqual');
-    expect(bundleSize).toMatchInlineSnapshot(`3357`);
+    expect(bundleSize).toMatchInlineSnapshot(`3477`);
   });
 });
--- a/src/compat/predicate/isEqual.spec.ts
+++ b/src/compat/predicate/isEqual.spec.ts
@@ -574,6 +574,36 @@ describe('isEqual', () => {
       map1.clear();
       map2.clear();
     });
+
+    const map1 = new Map<object, number>([
+      [{ a: 1 }, 1],
+      [{ b: 2 }, 2],
+    ]);
+    const map2 = new Map<object, number>([
+      [{ a: 1 }, 1],
+      [{ b: 2 }, 2],
+    ]);
+    expect(isEqual(map1, map2)).toBe(true);
+
+    const map3 = new Map<object, number>([
+      [{}, 1],
+      [{}, 2],
+    ]);
+    const map4 = new Map<object, number>([
+      [{}, 2],
+      [{}, 1],
+    ]);
+    expect(isEqual(map3, map4)).toBe(true);
+
+    const map5 = new Map<object, number>([
+      [{}, 1],
+      [{}, 2],
+    ]);
+    const map6 = new Map<object, number>([
+      [{}, 1],
+      [{}, 3],
+    ]);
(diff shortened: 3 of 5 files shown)


src/predicate/isEqual.spec.ts
```diff
@@ -112,4 +112,50 @@ describe('isEqual', () => {
 
     expect(isEqual(buffer3, buffer4)).toBe(false);
   });
+
+  it('should return true for equal Maps with structurally-equal object keys', () => {
```
Antoliny0919, line 116: Since compat/isEqual uses es-toolkit's isEqual implementation directly, this bug also affected compatibility between compat and Lodash.

How about adding map related test to compat/isEqual as well? Since this behavior should be guaranteed there too, I think it makes sense to cover it with a compatibility test.

Antoliny0919 (approved):
Thank you for work on this @xianjianlf2 ⭐️

I had a hard time understanding the issue when I first read the PR description. I think the description might not accurately explain the problem.

Could you take another look at it?

LGTM, I left a small suggestion :)

src/predicate/isEqual.spec.ts
```diff
@@ -112,4 +112,50 @@ describe('isEqual', () => {
...
+    ]);
+    const map2 = new Map<object, number>([
+      [{}, 1],
+      [{}, 3],
+    ]);
+
+    expect(isEqual(map1, map2)).toBe(false);
+  });
+
+  it('should return true for equal Sets with duplicate-shaped elements in different order', () => {
```
Antoliny0919: It looks like these test cases were added independently of the changes in this PR.

I do agree that adding test cases for `Set` is reasonable, and that both the expected and failing cases should eventually be covered.

That said, I'm not sure whether those changes belong in this PR 🤔

src/predicate/isEqual.spec.ts
```diff
@@ -112,4 +112,50 @@ describe('isEqual', () => {
...
+    ]);
+    const map2 = new Map<object, number>([
+      [{ a: 1 }, 1],
+      [{ b: 2 }, 2],
+    ]);
+
+    expect(isEqual(map1, map2)).toBe(true);
+  });
+
+  it('should return true for equal Maps with duplicate-shaped keys but differing values per pairing', () => {
```
Antoliny0919: This looks very similar to the `Set` test below. Would you consider renaming the test to follow the same naming pattern? Personally, I think the `Set` test name is a bit more intuitive because it explicitly mentions that the elements are in a different order.

```suggestion
  it('should return true for equal Maps with duplicate-shaped elements in different order, () => {
```

Hprogram:
Nice fix — I was independently digging into #1881 and reached the same root cause: since `Map.has`/`get` use SameValueZero, even a *single* structurally-equal object key failed on main (`isEqual(new Map([[{}, 1]]), new Map([[{}, 1]]))` → `false`), not only the duplicate-shaped case from the issue. The entry-by-entry multiset matching here resolves that class of false negatives correctly. This might also help sharpen the PR description that was mentioned in review.

One thing worth considering before merge: this removes the O(1) native lookup entirely, so **primitive-keyed Maps — the overwhelmingly common case — go from O(n) to O(n²)**. Measured on this branch vs main (vitest, avg of 5 runs after warmup):

| Case (string keys) | main | this PR |
|---|---|---|
| 5,000 entries, same insertion order | 0.76 ms | 4.75 ms (~6x) |
| 5,000 entries, reversed insertion order | 0.34 ms | 116 ms (~345x) |

Lodash is also O(n²) here, so this still "matches" lodash — but main previously beat it, and differing insertion order is realistic (e.g. Maps rebuilt from serialized or re-fetched data). The absolute numbers are machine-dependent, but the growth is not: scaling n from 1,000 to 5,000 in the reversed case grows the time ~19x (6 ms → 116 ms), which is the quadratic curve regardless of hardware.

<details>
<summary>Benchmark code (runs in the repo)</summary>

```ts
import { isEqual } from 'es-toolkit';

function timeIt(fn: () => void, runs: number): number {
  fn(); fn(); // warmup
  const start = performance.now();
  for (let i = 0; i < runs; i++) fn();
  return (performance.now() - start) / runs;
}

for (const n of [1000, 5000]) {
  const a = new Map<string, number>();
  const bSame = new Map<string, number>();
  const bRev = new Map<string, number>();
  for (let i = 0; i < n; i++) {
    a.set(`key-${i}`, i);
    bSame.set(`key-${i}`, i);
  }
  for (let i = n - 1; i >= 0; i--) bRev.set(`key-${i}`, i);

  console.log(`same-order n=${n}:`, timeIt(() => isEqual(a, bSame), 5), 'ms');
  console.log(`reversed   n=${n}:`, timeIt(() => isEqual(a, bRev), 5), 'ms');
}
```

Compared once on `main` and once on this branch. (Note if you run it via the `benchmarks` workspace: clear `benchmarks/node_modules/.vite` between switches — vite pre-bundles `es-toolkit` and will happily serve the cached implementation.)

</details>

A hybrid pass would keep both properties:

1. First pass: try `b.has(aKey)` (SameValueZero). If the key hits and the values are deep-equal, consume that entry. If the key hits but values differ, don't fail yet — defer the entry to a remainder pool, since it might still match a different structurally-equal key.
2. Second pass: run the current `findIndex` multiset matching only on the unmatched remainder.

Primitive-keyed Maps stay O(n) (the remainder is empty), and object-keyed Maps behave exactly as in this PR. Happy to help with this if useful — either way, the correctness fix is the important part. 🙏

xianjianlf2:
Addressed the primitive-key performance concern in f7febfc9 by adding a native exact-match fast path for Map keys and Set values, while preserving the structural fallback for unmatched entries. The fast path still routes exact Map keys through the customizer path so `isEqualWith` semantics are preserved.\n\nLocal checks:\n- `yarn vitest run src/predicate/isEqualWith.spec.ts`\n- `yarn typecheck`

xianjianlf2:
Follow-up for the Bundle Size check: 9490da63 updates the bundle-size snapshots after trimming the fast-path implementation. The isEqual snapshot now moves from 3477 to 3725 bytes (+248) instead of the larger first attempt.\n\nAdditional local check:\n- `yarn workspace benchmarks check-bundle-size`

xianjianlf2:
Rebased onto the latest `main` and pushed the updated branch (`543c2509`).

Local checks:
- `yarn vitest run src/predicate/isEqualWith.spec.ts` (40 tests passing)
- `yarn typecheck`
- `yarn workspace benchmarks check-bundle-size` (100 files / 173 tests passing)

Note: I left the unrelated local untracked `.agents/` directory untouched.

src/predicate/isEqual.spec.ts
xianjianlf2: Added this in the existing `src/compat/predicate/isEqual.spec.ts` map coverage. It now covers structurally-equal object keys, duplicate-shaped entries in different order, and the mismatched value case through the compat import path as well.

xianjianlf2:
Updated this branch with latest `main` using a normal merge commit (no history rewrite), so it is no longer behind the base branch.

Local checks after the merge:
- `yarn vitest run src/predicate/isEqualWith.spec.ts`
- `yarn typecheck`

Hprogram:
I re-ran the benchmark against the current head (`55997ebe`) to check whether the fast path landed the way we hoped. Short version: the regression I raised is gone, and there is a second one worth a look.

**The reversed primitive-key case is linear now.** On this branch, comparing two reversed primitive-key Maps scales with `n`, not `n²`:

| n | this branch |
|---|---|
| 1,000 | 0.045 ms |
| 5,000 | 0.257 ms |
| 10,000 | 0.564 ms |

5x the input costs 5.8x the time, then 2x costs 2.2x. The O(n²) blowup is no longer there — thanks for taking that on.

**A side benefit that is not in the PR description: this also fixes `Set`.** The `setTag` path on `main` is already O(n²) for reversed primitive members, and this branch makes it linear:

| reversed primitive Set, n = 10,000 | |
|---|---|
| `main` | 204.8 ms |
| this branch | 0.215 ms |

That is a pre-existing bug this PR happens to remove. It seems worth calling out in the description — reviewers weighing the bundle-size increase would probably want to know it buys a `Set` fix as well.

**What is left: a constant ~3.3–3.7x on every Map comparison.** It does not grow with `n`, so it is a constant factor rather than a complexity problem, but it lands on the most common case — primitive keys:

| scenario, n = 10,000 | `main` | this branch | ratio |
|---|---|---|---|
| Map, primitive keys, same order | 0.175 ms | 0.646 ms | 3.69x |
| Map, primitive keys, reversed | 0.172 ms | 0.562 ms | 3.27x |
| Map, shared object-key refs, reversed | 0.176 ms | 0.600 ms | 3.40x |

Reading the mapTag branch, three things look like they account for it:

1. `new Map(b)` copies the whole of `b` on every call, before we know whether any entry will miss.
2. `remainingBEntries.get(aKey)` followed by `remainingBEntries.has(aKey)` hashes the key twice.
3. `isEqualWithImpl(aKey, aKey, ...)` compares each key against itself. For plain `isEqual` this is always `true`, so it is pure overhead on the hot path. I assume the intent is to let a customizer reject a key — if so, a comment would help, since it reads as accidental.

I sketched a variant with those three removed to check whether the constant is inherent to the hybrid approach or just incidental. It is not inherent: the sketch runs **3.9–4.4x faster than this branch** on the three Map scenarios above, while producing identical results on the full suite (4,552 tests, no new failures) and on 15 edge cases I diffed against both this branch and Lodash — NaN keys, `-0` vs `0` keys and values, nested Maps, Maps used as keys, cyclic values, and duplicate structurally-equal keys. I am not proposing a patch here — this is your PR and the constant is a separate concern from the correctness fix. If it would help, I am happy to open a follow-up PR for it once this lands.

**One thing the fast path cannot help with, by construction.** It hits when `b` has the same key by SameValueZero, which is a hash lookup — so it always hits for primitive keys regardless of order, and always misses for structurally-equal keys that are distinct references. That is exactly the case this PR exists to fix, so that path stays fully O(n²): 20x the entries costs ~400x the time, and a single comparison of two 2,000-entry Maps takes ~780 ms on my machine.

That matches Lodash, so it is not a regression — but "addressed the primitive-key performance concern" could be read as the whole performance story being settled, and it is not. Stating the object-key cost in the description would set the right expectation, and it is bounded by user-controlled input, which is worth someone knowing before it shows up in a hot path.

<details>
<summary>Method</summary>

Node v22.14.0, macOS, M5 Pro. Each implementation runs in its own process — an in-process harness gave me a 1.4x spread between byte-identical copies of the same code, so those numbers are unusable. Each figure is the median of 9 trials, repeated in two rounds with the run order reversed. As a control I measured a byte-identical copy of `main` alongside `main`: it came out at 0.99–1.03x across all four scenarios, so differences above ~1.1x are signal.

Ratios for structurally-equal object keys against `main` are omitted on purpose: `main` returns `false` immediately there (the bug), so the two are not doing comparable work.
</details>

PR #1886, opened 2026-07-14
fix(compat/includes): exclude 'length' property from array like value matching

Antoliny0919:
## Summary

When `includes` is used with an array like object, the value of the `length` property itself is also included in the search.
This causes a compatibility issue with Lodash, which only checks the elements and does not consider the `length` property itself.

```js
    function includes(collection, value, fromIndex, guard) {
      collection = isArrayLike(collection) ? collection : values(collection);
      fromIndex = (fromIndex && !guard) ? toInteger(fromIndex) : 0;

      var length = collection.length;
      if (fromIndex < 0) {
        fromIndex = nativeMax(length + fromIndex, 0);
      }
      return isString(collection)
        ? (fromIndex <= length && collection.indexOf(value, fromIndex) > -1)
        : (!!length && baseIndexOf(collection, value, fromIndex) > -1);
    }
```

-> Lodash treats array like objects the same way as arrays, while es-toolkit compat treats them as regular objects.

## Reproduce

```ts
import { includes as loIncludes } from 'lodash';
import { includes as esIncludes } from './src/compat/array/includes';

const arrayLike = { length: 2, 0: 'a', 1: 'b' };

console.log(loIncludes.includes(arrayLike, 2)); // false
console.log(esIncludes.includes(arrayLike, 2)); // true
```

### DIFF
--- a/src/compat/array/includes.spec.ts
+++ b/src/compat/array/includes.spec.ts
@@ -12,6 +12,13 @@ describe('includes', () => {
     expect(includes(obj, 'value')).toBe(false);
   });
 
+  it(`should not match arrayLike 'length' property value`, () => {
+    const arrayLike = { 0: 'a', 1: 'b', length: 2 };
+    expect(includes(arrayLike, 2)).toBeFalsy();
+    expect(includes(arrayLike, 'a')).toBeTruthy();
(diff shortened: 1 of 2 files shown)


raon0211 (approved):
Thanks!

PR #1888, opened 2026-07-14
docs(compat/includes): mark target, fromIndex parameter as optional in signature

Antoliny0919:
## Summary

It looks like the optional indicator is missing :(


raon0211 (approved):
Thanks!

PR #1889, opened 2026-07-14
docs(compat/head): remove duplicate undefined return note from summary

Antoliny0919:
## Summary

head-duplicate

It looks like there are some duplicate messages. I think we can remove the one above!


raon0211 (approved):
Thanks!

PR #1893, opened 2026-07-17
fix(compat/maxBy, compat/minBy): restore lodash compatible comparison behavior

### LINKED ISSUE #1890: maxBy/minBy return the first element for non-numeric values (regression in 1.48.0)
## Description

Since v1.48.0, `maxBy` and `minBy` (both in `es-toolkit` and `es-toolkit/compat`) silently return the **first element** of the array whenever the values extracted by the iteratee are not numbers (e.g. strings), instead of the actual max/min element. v1.47.0 behaved correctly, and lodash (which `es-toolkit/compat` aims to match) handles strings fine.

## Reproduction

```js
import { maxBy, minBy } from 'es-toolkit/compat'; // same with 'es-toolkit'

maxBy([{ v: 'a' }, { v: 'b' }], (o) => o.v);
// v1.47.0 / lodash: { v: 'b' } ✅
// v1.48.0+:         { v: 'a' } ❌ (first element, regardless of order)

minBy([{ v: 'b' }, { v: 'a' }], (o) => o.v);
// v1.47.0 / lodash: { v: 'a' } ✅
// v1.48.0+:         { v: 'b' } ❌
```

Still reproduces on `1.49.0` and on the latest dev build I tried (`1.49.0-dev.1949`).

## Cause

The v1.48.0 rewrite initializes the accumulator to a number and compares against it:

```js
let max = -Infinity;
...
if (value > max) { max = value; maxElement = element; }
```

For a string value, `'a' > -Infinity` is always `false` (relational comparison between a non-numeric string and a number yields `false`), so the loop never picks any element and the function falls back to `items[0]`. Same pattern with `Infinity` in `minBy`. Dates still work only because they coerce numerically.

The pre-1.48 implementation compared elements against each other (`getValue(element) > getValue(maxElement)`), which worked for any comparable values.

## Expected behavior

`maxBy`/`minBy` should return the element with the max/min extracted value for any values comparable with `<`/`>` (strings included), as in v1.47.0 and lodash. If the new implementation is intentionally number-only, that's a breaking change that at least deserves a major bump and a documented migration note — but for `es-toolkit/compat` it breaks lodash parity, which seems contrary to its goal.

## Impact

This is silent data corruption for existing code: no error is thrown, the wrong element is just returned. In our codebase this regressed logic like `maxBy(versions, v => v.version)` where versions are strings — the bug was only caught by a property-based test.

Antoliny0919:
## Summary

Closes: https://github.com/toss/es-toolkit/issues/1890

regression in https://github.com/toss/es-toolkit/commit/e319b3b2826a7101d423ba4dc27bf802ad05b508 

This change fixes the issue where comparisons weren't being performed correctly for string values.
Additionally, it updates the behavior so that NaN values are skipped when present during comparison, and undefined is returned when the array contains only NaN values.
These were additional compatibility issues.

### e911de8

```ts
console.log(lodash.maxBy([{ v: 'a' }, { v: 'b' }], item => item.v)); // { v: 'b' }
console.log(maxBy([{ v: 'a' }, { v: 'b' }], item => item.v)); // { v: 'a' }
console.log(lodash.minBy([{ v: 'b' }, { v: 'a' }], item => item.v)); // { v: 'a' }
console.log(minBy([{ v: 'b' }, { v: 'a' }], item => item.v)); // { v: 'b' }
```

### aeaad75

```ts
console.log(lodash.maxBy([3, NaN, 1, 2], x => x)); // 3
console.log(maxBy([3, NaN, 1, 2], x => x)); // NaN
console.log(lodash.minBy([3, NaN, 1, 2], x => x)); // 1
console.log(minBy([3, NaN, 1, 2], x => x)); // NaN

console.log(lodash.maxBy([NaN, NaN], x => x)); // undefined
console.log(maxBy([NaN, NaN], x => x)); // MaN
console.log(lodash.minBy([NaN, NaN], x => x)); // undefined
console.log(minBy([NaN, NaN], x => x)); // NaN
```

### DIFF
--- a/docs/compat/reference/math/maxBy.md
+++ b/docs/compat/reference/math/maxBy.md
@@ -16,7 +16,7 @@ const maxItem = maxBy(array, iteratee);
 
 ## Usage
 
-### `maxBy(array, iteratee)`
+### `maxBy(array, iteratee?)`
 
 Finds the element in an array that has the largest value when computed by a function.
 
@@ -107,7 +107,7 @@ maxBy(undefined);
 #### Parameters
 
 - `array` (`ArrayLike<T> | null | undefined`): The array to search.
-- `iteratee` (`ValueIteratee<T>`, optional): The function, property name, or condition to apply to each element.
+- `iteratee` (`ValueIteratee<T>`, optional): The function, property name, or condition to apply to each element. Default is `identity`.
 
 #### Returns
 
--- a/docs/compat/reference/math/minBy.md
+++ b/docs/compat/reference/math/minBy.md
@@ -16,7 +16,7 @@ const minItem = minBy(array, iteratee);
 
 ## Usage
 
-### `minBy(array, iteratee)`
+### `minBy(array, iteratee?)`
 
 Finds the element in an array that has the smallest value when computed by a function.
 
@@ -107,7 +107,7 @@ minBy(undefined);
 #### Parameters
 
 - `array` (`ArrayLike<T> | null | undefined`): The array to search.
-- `iteratee` (`ValueIteratee<T>`, optional): The function, property name, or condition to apply to each element.
+- `iteratee` (`ValueIteratee<T>`, optional): The function, property name, or condition to apply to each element. Default is `identity`.
 
 #### Returns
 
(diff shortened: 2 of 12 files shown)


Antoliny0919:
### [35d57a4](https://github.com/toss/es-toolkit/pull/1893/commits/35d57a448efcb16b699395a51d564807406d9146)

minBy-default

raon0211 (approved):
Thanks!

PR #1898, opened 2026-07-18
fix: Can't scroll playground content

### LINKED ISSUE #1897: Playground content is cut off when the code has many lines
When the code has many lines, the content below is not fully visible in the playground. The screenshot below shows an example with 31 lines of code.

Screenshot of playground.

I think adding a `height: 0` style to the `.playground-sandpack` element resolves the issue.

Screenshot of fixed playground

blro0319:
Fixes #1897

### DIFF
--- a/docs/.vitepress/components/Playground.vue
+++ b/docs/.vitepress/components/Playground.vue
@@ -576,6 +576,7 @@ function resetCode() {
 .playground-sandpack {
   padding: 0;
   flex: 1;
+  height: 0;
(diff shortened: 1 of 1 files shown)


Antoliny0919 (approved):
Good catch 👍 Thank you @blro0319 

## After

https://github.com/user-attachments/assets/155f5d71-827f-4302-8d55-4ee18ad8803e

raon0211 (approved):

PR #1901, opened 2026-07-19
fix(retry): throw the last attempt's error without applying a delay

ssi02014:
## Summary

`retry` applied a delay on **every** failed attempt, including the final one. When the last attempt fails, `retry` should throw the error immediately — waiting for a delay when no further attempt follows is pure wasted time, and not waiting is the natural behavior.

This PR skips the delay after the final attempt so the error is thrown immediately.

## Problem

In the previous implementation, the delay ran unconditionally inside the `catch` block:

```ts
} catch (err) {
  error = err;

  if (!shouldRetry(err, attempts)) {
    throw err;
  }

  // Always awaited a delay, even on the last attempt
  const currentDelay = typeof delay === 'function' ? delay(attempts) : delay;
  await delayToolkit(currentDelay);
}
```

So with `retry(func, { delay, retries: 3 })` where `func` always fails, the timeline was:

```
attempt 0 (fail) → delay → attempt 1 (fail) → delay → attempt 2 (fail) → delay → attempt 3 (fail) → delay → throw
                                                                                                   ^^^^^^^ unnecessary
```

The delay after the last attempt adds latency without any benefit.

## Solution

Guard the delay with `attempts < retries` so it only runs when a retry actually follows:

```ts
} catch (err) {
  error = err;

  if (!shouldRetry(err, attempts)) {
    throw err;
  }

  if (attempts < retries) {
    const currentDelay = typeof delay === 'function' ? delay(attempts) : delay;
    await delayToolkit(currentDelay);
  }
}
```

New timeline:

```
attempt 0 (fail) → delay → attempt 1 (fail) → delay → attempt 2 (fail) → delay → attempt 3 (fail) → throw
                                                                                                   ^^^^^ immediate
```

## Changes

- `src/function/retry.ts` — Only apply the delay when another retry follows (`attempts < retries`).
- `src/function/retry.spec.ts` — Add a test verifying the last attempt throws without a delay.

## Test

```ts
it('should throw the error from the last attempt without applying a delay', async () => {
  const func = vi.fn().mockRejectedValue(new Error('failure'));
  const retries = 3;
  const delay = vi.fn((attempt: number) => attempt);

  await expect(retry(func, { delay, retries })).rejects.toThrow('failure');

  expect(func).toHaveBeenCalledTimes(retries + 1); // 4 attempts (1 initial + 3 retries)
  expect(delay).toHaveBeenCalledTimes(retries);    // delay called 3 times, NOT on the last attempt
});
```

The key assertion is `func` is called `retries + 1` times while `delay` is called only `retries` times — proving the delay is skipped on the final failing attempt.

## Notes

- No public API change; behavior only removes an unnecessary wait before the final throw.


raon0211 (approved):
Thanks!

PR #1909, opened 2026-07-20
feat: capitalCase (Start Case)

### LINKED ISSUE #306: Support for `capitalCase`
nice to have:

```typescript
capitalCase("hello world")
// output: Hello World
```

`es-toolkit` already supports some *case.
**not the same as [capitalize](https://es-toolkit.slash.page/reference/string/capitalize.html)**

hazrid93:
Closes #306.

Add capitalCase(str) that capitalises the first letter of each word and lower cases the rest, joining words with a single space: "hello world" to "Hello World", "camelCase" to "Camel Case", "hyphen text" to "Hyphen Text". Distinct from capitalize (which only capitalises the first letter of the whole string). Reuses the existing words and capitalize helpers.

### DIFF
--- a/src/string/capitalCase.ts
+++ b/src/string/capitalCase.ts
@@ -1,0 +1,26 @@
+import { capitalize } from './capitalize.ts';
+import { words as getWords } from './words.ts';
+
+/**
+ * Converts a string to capital case (also known as "start case"), in which
+ * the first letter of each word is capitalised and the rest are lower-cased,
+ * with words separated by a single space. Not to be confused with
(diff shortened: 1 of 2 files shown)


raon0211:
Thank you for the contribution!

However, this implementation is functionally identical to the existing [`startCase`](https://es-toolkit.dev/reference/string/startCase.html) — both split the input with `words()`, capitalize each word, and join with a single space. Adding `capitalCase` would just be a duplicate under a different name.

It also doesn't address the concern raised in #306 about special characters being stripped, since it relies on the same `words()` helper.

I'll close this PR along with the issue. Thanks again for your time!

PR #1910, opened 2026-07-20
feat: Async/Promise memoization

### LINKED ISSUE #1739: Feature request: Async/Promise memoization
Currently only synchronous calls are supported in `memoize` function. Would it make sense to also add support or dedicated function for async/Promise memoization?

Caching promises instead of their final results can lead to [potential memory leaks](https://x.com/matteocollina/status/1502585707256242176) so having option of tool which prevents this by proper memoization could help in preventing this problem.

Sync implementation is pretty similar to [basic async memoization](https://x.com/izs/status/1502506002926628867) so it could be easily adapted.

hazrid93:
Closes #1739.

Add memoizePromise(fn, options) that caches the in flight Promise (so concurrent calls with the same argument share a single execution) and deletes the cache entry on rejection so a failed call is retried on the next invocation. Supports the same options as memoize (cache, getCacheKey) and exposes a .cache property. Existing memoize is unchanged.

### DIFF
--- a/src/function/memoizePromise.ts
+++ b/src/function/memoizePromise.ts
@@ -1,0 +1,64 @@
+import type { MemoizeCache } from './memoize.ts';
+
+/**
+ * Creates a memoized version of an async (Promise-returning) function.
+ *
+ * Unlike {@link memoize}, which caches the resolved value, `memoizePromise`
+ * caches the **in-flight Promise** itself, so concurrent calls with the same
+ * argument share a single execution. Crucially, when a Promise rejects the
+ * cache entry is deleted, so a failed call is retried on the next invocation
+ * instead of caching the rejection forever (which would leak memory and
+ * freeze the cache — see issue #1739).
+ *
+ * Like {@link memoize}, this works with functions that take zero or just one
+ * argument; provide a `getCacheKey` for non-primitive arguments.
+ *
+ * @template F - The async function to memoize.
+ * @param fn - The async function to memoize.
+ * @param [options={}] - Optional configuration.
+ * @param [options.cache] - The cache object. Defaults to a new `Map`.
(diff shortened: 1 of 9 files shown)


niksy:
I think this doesn’t cover important part for Promise memoization: caching return value to prevent memory leaks.

Please see original issue for links on explanation why this is important.

src/function/memoizePromise.ts
```diff
@@ -1,64 +1,83 @@
...
-  const { cache = new Map<unknown, ReturnType<F>>(), getCacheKey } = options;
+): MemoizedPromise<F> {
+  const cache: MemoizeCache<any, Awaited<ReturnType<F>>> = options.cache ?? new Map<unknown, Awaited<ReturnType<F>>>();
+  const { getCacheKey } = options;
+  const promiseCache = new Map<unknown, ReturnType<F>>();
 
   const memoizedFn = function (this: unknown, arg: Parameters<F>[0]): ReturnType<F> {
     const key = getCacheKey ? getCacheKey(arg) : arg;
 
+    if (promiseCache.has(key)) {
```
niksy: Do we actually need Promise cache? Why not use final value cache for in-flight request and final value (as in https://x.com/izs/status/1502506002926628867)?

I suppose this is because final value isn’t always thenable, but we can store Promise value to cache with `Promise.resolve`.

hazrid93: Updated in c41639a1. `memoizePromise` now uses a single cache: it stores the in-flight Promise, wraps cache hits with `Promise.resolve`, replaces the Promise with the resolved value, and deletes rejected entries.

raon0211:

niksy:
@raon0211 what is the proposed way forward here? Issue for this is already opened (#1739) so I suppose this can act as start of discussion?

PR #1914, opened 2026-07-20
feat: support for throttleAsync function

### LINKED ISSUE #976: Support for `throttleAsync` function
What if we added a `throttleAsync` function to throttle asynchronous functions?

### Suggesting

I suggest implementing a promise-based throttling function that only allows a new invocation when the previous one has completed.

```typescript
export function throttleAsync<F extends (...args: any[]) => Promise<any>>(
  func: F,
  { signal }: { signal?: AbortSignal } = {}
): ((...args: Parameters<F>) => Promise<ReturnType<F> | undefined>) & { cancel: () => void } {
  let isRunning = false;
  let isCancelled = false;
  
  if (signal) {
    signal.addEventListener('abort', () => {
      isCancelled = true;
    }, { once: true });
  }
  
  function cancel(): void {
    isCancelled = true;
  }
  
  const throttled = async function(...args: Parameters<F>): Promise<ReturnType<F> | undefined> {
    if (isRunning || isCancelled) {
      return undefined;
    }
    
    isRunning = true;
    
    try {
      const result = await func(...args);
      return result;
    } finally {
      isRunning = false;
    }
  } as ((...args: Parameters<F>) => Promise<ReturnType<F> | undefined>) & { cancel: () => void };
  
  throttled.cancel = cancel;
  
  return throttled;
}
```

### Why this is needed

The current `throttle` function is time-based, which works well for UI events but has limitations with asynchronous operations. Consider this common scenario:

```typescript
// Current approach with time-based throttle
const throttledFetch = throttle(fetchData, 1000);

throttledFetch('/api/users'); // Executes and takes 2 seconds to complete
// 1.5 seconds later
throttledFetch('/api/users'); // Executes again because 1 second has passed
                              // But the first request hasn't finished yet!
```

This can lead to:
1. Multiple simultaneous API calls
2. Race conditions
3. Duplicate form submissions

With the proposed `throttleAsync`:

```typescript
// Using promise-based throttleAsync
const throttledFetch = throttleAsync(fetchData);

throttledFetch('/api/users'); // Executes and takes 2 seconds
throttledFetch('/api/users'); // Ignored because previous call is still in progress
// 2 seconds later when first call completes
throttledFetch('/api/users'); // Now allowed to execute
```

#### Preventing duplicate form submissions
```typescript
const form = document.querySelector('form');
const submitForm = async (event) => {
  event.preventDefault();
  const data = new FormData(form);
  return fetch('/api/submit', { method: 'POST', body: data });
};

// Prevent multiple submissions while one is in progress
const throttledSubmit = throttleAsync(submitForm);
form.addEventListener('submit', throttledSubmit);
```

#### Avoiding duplicate API calls
```typescript
function UserProfile({ userId }) {
  const [user, setUser] = useState(null);
  
  // Create throttled fetch that prevents duplicate calls
  const fetchUserData = useCallback(
    throttleAsync(async (id) => {
      const response = await fetch(`/api/users/${id}`);
      const data = await response.json();
      setUser(data);
    }),
    []
  );
  
  // Even if this effect runs multiple times in quick succession,
  // only one API call will be made at a time
  useEffect(() => {
    fetchUserData(userId);
  }, [userId, fetchUserData]);
}
```

I think this would be a useful addition to complement the existing `throttle` function, as it addresses a different but common use case in modern web applications.

### References
- [[Discussion about throttleAsync in PR #414](https://github.com/toss/es-toolkit/pull/414)](https://github.com/toss/es-toolkit/pull/414)
- [[Handling race conditions in React](https://maxrozen.com/race-conditions-fetching-data-react-with-useeffect)](https://maxrozen.com/race-conditions-fetching-data-react-with-useeffect)

hazrid93:
Closes #976.

Add throttleAsync(func, {signal}) that only allows a new invocation when the previous one has completed. While a call is in flight, further calls resolve to null immediately (dropped, classic async throttle cooldown). Returns the in flight Promise for the running call.

### DIFF
--- a/src/function/index.ts
+++ b/src/function/index.ts
@@ -18,4 +18,5 @@ export { rest } from './rest.ts';
 export { retry } from './retry.ts';
 export { spread } from './spread.ts';
 export { throttle, type ThrottledFunction, type ThrottleOptions } from './throttle.ts';
+export { throttleAsync } from './throttleAsync.ts';
 export { unary } from './unary.ts';
--- a/src/function/throttleAsync.ts
+++ b/src/function/throttleAsync.ts
@@ -1,0 +1,63 @@
+interface ThrottleAsyncOptions {
+  /**
+   * An optional AbortSignal; when it aborts, the throttle lock is released so
+   * the next call can run (the in-flight call is not cancelled).
+   */
+  signal?: AbortSignal;
+}
+
+/**
+ * Creates a throttled version of an async function that only allows a new
+ * invocation when the previous one has completed. While a call is in flight,
+ * further calls resolve to `null` immediately (they are dropped, the classic
+ * throttle "cooldown" behaviour for async). See issue #976.
+ *
+ * The returned function exposes a `.cancel()` method that releases the
+ * throttle lock so subsequent calls can run even if a previous call is still
+ * in flight (it does not cancel the in-flight call).
+ *
+ * @template F - The async function to throttle.
+ * @param func - The async function to throttle.
+ * @param options - Optional configuration.
+ * @returns A throttled async function with a `.cancel` method.
+ *
+ * @example
+ * const throttled = throttleAsync(async () => await fetch('/data'));
+ * await throttled(); // runs fetch
+ * await throttled(); // while the previous is in flight, resolves to null
+ */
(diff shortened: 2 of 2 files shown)


raon0211:

PR #1917, opened 2026-07-20
feat: iterator chain / concatenation (lazy concat of iterables)

### LINKED ISSUE #771: feature: iterator chain / concatenation

hazrid93:
Closes #771.

Add chain(...iterables): a lazy generator that yields every element of every iterable, in order, without an intermediate array. Works with any iterable: arrays, strings, Sets, generators. Returns an IterableIterator, so callers can `Array.from(chain(...))`, spread it, or use `.next()`.

### DIFF
--- a/src/array/chain.ts
+++ b/src/array/chain.ts
@@ -1,0 +1,28 @@
+/**
+ * Chains multiple iterables together into a single lazy iterator (Python's
+ * `itertools.chain`). Like `Array.prototype.concat`, but lazy — it yields
+ * elements from each iterable in order without building an intermediate array,
+ * and works with any iterable (not only `Symbol.isConcatSpreadable` ones).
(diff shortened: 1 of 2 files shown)


raon0211:

PR #1921, opened 2026-07-20
feat: coalesce: single flight async call deduplication

### LINKED ISSUE #1608: feat: coalesce (single-flight) utility
Hello, I would like a coalition utility to group identical concurrent asynchronous calls into a single execution.

Most implementations have a re-entrancy bug where the factory runs before the promise is stored. A robust version for the toolkit would be:

```typescript
const inflight = new Map<string, Promise<any>>();

export async function coalesce<T>(key: string, fn: () => T | PromiseLike<T>): Promise<T> {
  const active = inflight.get(key);
  if (active) return active;

  const promise = Promise.resolve()
    .then(() => fn())
    .finally(() => inflight.delete(key));

  inflight.set(key, promise);
  return promise;
}
```

hazrid93:
Closes #1608.

Add coalesce(key, fn): if a call with `key` is already in flight, return its promise; otherwise run `fn()` and store the promise so concurrent callers await the same result. The entry is removed once the promise settles (later calls re run fn). fn is invoked on a microtask AFTER the entry is stored, so a re entrant coalesce(key, …) inside fn returns the stored promise instead of re running fn and racing: the robustness the issue asks for.

### DIFF
--- a/src/function/coalesce.ts
+++ b/src/function/coalesce.ts
@@ -1,0 +1,50 @@
+/**
+ * Coalesces concurrent calls that share a `key` into a single in-flight
+ * execution. If a call with `key` is already in flight, returns that promise;
+ * otherwise invokes `fn()` and stores its promise so concurrent callers await
+ * the same result. The entry is removed once the promise settles, so a later
+ * call re-runs `fn`.
+ *
+ * `fn` is invoked on a microtask AFTER the entry is stored, which avoids the
+ * re-entrancy bug present in naive implementations (where the factory re-runs
+ * before its promise is stored). See issue #1608.
+ *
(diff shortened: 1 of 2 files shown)


seungrodotlee:
I believe this implementation is quite risky.

Since `in-flight` state is shared at the module scope, using it incorrectly in a server environment—particularly in a Next.js BFF—could cause data to leak across requests. What makes this even more concerning is that the module-level sharing is effectively a black box to consumers of this API.

The current PR also lacks documentation and tests covering this behavior.

That said, I also see a need for this functionality, so I’ve opened a separate PR with an alternative approach. Please take a look #1994.

raon0211:

PR #1922, opened 2026-07-20
feat: isTreeData: validate tree structured data

### LINKED ISSUE #1598: Feature request: support isValidTreeNode,isLeafNode and isTreeData
Supporting tree-structured data could be beneficial for framework development and many other use cases.If acceptable, I can add predicate  functions isValidTreeNode,isLeafNode and isTreeData.

hazrid93:
Closes #1598.

Add isTreeData(value, { childrenKey? = 'children' }): returns true when value is a plain object whose optional childrenKey array contains further valid tree nodes, OR an array (a forest) of such nodes. A leaf is any plain object without the children key.

### DIFF
--- a/src/predicate/index.ts
+++ b/src/predicate/index.ts
@@ -27,6 +27,7 @@ export { isRegExp } from './isRegExp.ts';
 export { isSet } from './isSet.ts';
 export { isString } from './isString.ts';
 export { isSymbol } from './isSymbol.ts';
+export { isTreeData } from './isTreeData.ts';
 export { isTypedArray } from './isTypedArray.ts';
 export { isUndefined } from './isUndefined.ts';
 export { isWeakMap } from './isWeakMap.ts';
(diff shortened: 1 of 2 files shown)


raon0211:
Thank you for the suggestion, and for taking the time to open this issue!

Unfortunately, as described in our [contributing guide](https://github.com/toss/es-toolkit/blob/main/.github/CONTRIBUTING.md#41-what-we-accept), we are currently accepting new feature contributions on a limited basis. To keep the library focused and maintainable, new functions need broad community demand and a clear, widely applicable use case before being accepted.

Validating tree-structured data tends to be fairly application-specific — the shape of tree nodes (children key, leaf definition, etc.) varies from project to project, so a general-purpose isTreeData would be hard to design in a way that fits most users' needs. For this reason, we'd like to respectfully decline this proposal for now.

Thank you again for your interest in es-toolkit, and we hope to see you around!

PR #1923, opened 2026-07-20
feat: flattenObject: return a Map instead of a Record

### LINKED ISSUE #1388: Feature Request: `flattenObject` into `Map` object
I need to convert a JSON locales file into a flattened Map. This is what I'm currently having to do:

```ts
import { flattenObject } from 'es-toolkit'

let { default: contents } = await import('@locales/en.json')
let translations = new Map(Object.entries(flattenObject(contents)))
```

would be nice if flattenObject took a target object that responds to `set`, i.e. Map:

```ts
import { flattenObject } from 'es-toolkit'

let { default: contents } = await import('@locales/en.json')
let translations = flattenObject(contents, new Map())
```

It's a bit cleaner, and it would be faster (es-toolkit would set keys into the map direct, rather than the Object.entries approach I'm using now)

hazrid93:
Closes #1388.

Add an optional `{ map?: boolean }` (default false, unchanged behavior): when true, flattenObject returns a `Map<string, any>` (built via `new Map(Object.entries(flat))`) instead of a plain object. TypeScript overloads ensure the return type is `Map<string, any>` when `map: true` and `Record<string, any>` otherwise. Honours the `delimiter` option alongside `map`.

### DIFF
--- a/src/object/flattenObject.ts
+++ b/src/object/flattenObject.ts
@@ -6,6 +6,11 @@ interface FlattenObjectOptions {
    * @default '.'
    */
   delimiter?: string;
+  /**
+   * When `true`, return a `Map<string, any>` instead of a plain object. See issue #1388.
+   * @default false
+   */
+  map?: boolean;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


raon0211:
Thanks for the pull request!

es-toolkit favors keeping each function small and single-purpose, and we don't want to add overloads + a union return type to flattenObject for a need that hasn't shown broad demand. The existing composition already covers the use case:

```typescript
const map = new Map(Object.entries(flattenObject(contents)))
```

Appreciate the pull request — closing to keep the API minimal. 🙏

PR #1924, opened 2026-07-20
feat: groupByMap, keyByMap, countByMap: Map returning variants

### LINKED ISSUE #1278: It'd be neat if Array functions countBy, groupBy, keyBy could return a Map if specified
Right now the functions countBy, groupBy, keyBy return an object/dict, which is good. But there are times I want a more complex key to be indexed, and a Map type would be preferrable. It'd be neat if that were a possible output, either via another function, or parameter specification

hazrid93:
Closes #1278.

Add groupByMap, keyByMap, countByMap: Map returning variants whose keys can be of any type (including object references), since a Map preserves key identity. groupByMap groups items into Map<K, T[]>; keyByMap maps each key to an item (last wins); countByMap counts occurrences per key into Map<K, number>. The existing Record returning functions are unchanged.


raon0211:
Thank you for the contribution! However, I don't think we can accept this PR, for two reasons:

First, es-toolkit already provides Map-returning variants of these functions in the `es-toolkit/map` module (available since v1.44.0):

- [keyBy](https://es-toolkit.dev/reference/map/keyBy.html) — indexes entries by a computed key, returning a `Map`
- [countBy](https://es-toolkit.dev/reference/map/countBy.html) — counts entries by a computed key, returning a `Map`

Second, for `groupBy`, modern JavaScript already has a built-in: [Map.groupBy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/groupBy) (ES2024) does exactly what `groupByMap` proposes:

```ts
Map.groupBy(items, item => item.someKey); // Map<K, T[]>
```

One of es-toolkit's design principles is to not implement functions that are replaceable by modern JavaScript built-ins, so `groupByMap` falls outside our scope.

Thanks again for taking the time to contribute — closing this in favor of the existing `es-toolkit/map` functions and `Map.groupBy`.

PR #1925, opened 2026-07-20
feat: medianBigint: bigint support for median()

### LINKED ISSUE #1588: Feature request: support `bigint` in `median()`

hazrid93:
Closes #1588.

Add medianBigint(nums: bigint[]): bigint: the middle value for odd length arrays; for even length, the truncated average `(a + b) / 2n` (bigint has no fractional type). Sorts using bigint comparison. Throws RangeError for an empty array (a bigint cannot represent NaN).

### DIFF
--- a/src/math/index.ts
+++ b/src/math/index.ts
@@ -3,6 +3,7 @@ export { inRange } from './inRange.ts';
 export { mean } from './mean.ts';
 export { meanBy } from './meanBy.ts';
 export { median } from './median.ts';
+export { medianBigint } from './medianBigint.ts';
 export { medianBy } from './medianBy.ts';
 export { percentile } from './percentile.ts';
 export { random } from './random.ts';
--- a/src/math/medianBigint.ts
+++ b/src/math/medianBigint.ts
@@ -1,0 +1,42 @@
+/**
+ * Calculates the median of an array of `bigint`s.
+ *
+ * The median is the middle value of a sorted array. If the array has an odd
+ * number of elements, the median is the middle value. If the array has an
+ * even number of elements, it returns the truncated average of the two middle
+ * values (`(a + b) / 2n`), since `bigint` has no fractional type. See issue
+ * #1588.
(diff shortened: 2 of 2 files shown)


Antoliny0919 (commented):
I suspect these PRs were generated by AI. They add features, but none of them include documentation or tests. Also, many of the issues are already assigned to other contributors.

I'm not sure how the maintainers feel about AI generated PRs, but submitting PRs indiscriminately like this doesn't seem like a good approach.

I think it would be a good idea to read the contribution [guidelines](https://github.com/toss/es-toolkit/blob/main/.github/CONTRIBUTING.md) again.

++ If AI was used during the development process, it would also be a good idea to clearly mention that in the PR description or include it in the commit messages.

raon0211:
Hi @hazrid93, thank you for taking the time to work on this!

Since this PR was opened, bigint support for `median` has landed through the new `es-toolkit/bigint` entry point in #1966 — see [`src/bigint/median.ts`](https://github.com/toss/es-toolkit/blob/main/src/bigint/median.ts). Instead of a separate `medianBigint` function in `es-toolkit/math`, we decided to provide bigint variants of math functions under their own dedicated entry point:

```ts
import { median } from 'es-toolkit/bigint';

median([1n, 2n, 3n, 4n]); // 2.5
```

Since the functionality this PR adds is already available on `main`, I'll close this one. Thanks again for your contribution, and we'd be happy to see more PRs from you!

PR #1927, opened 2026-07-20
fix(compat): skip null/undefined iteratee values in maxBy/minBy

Kropiunig:
## Summary

`compat`'s `maxBy`/`minBy` diverge from lodash (and from their own sibling `max`/`min`) when the iteratee yields `null` or `undefined` for elements — e.g. a missing/optional key.

They only skip `NaN`:

```ts
if (Number.isNaN(current)) {
  continue;
}
```

but `max`/`min` skip `current == null` (and symbols) as well. Two problems follow:

1. `max`/`min` is initialised to `undefined` **and** reused as the "not yet seen" sentinel (`max === undefined`). An `undefined` iteratee result sets the sentinel back to `undefined`, so it never flips and every later element overwrites the result — the function returns the **last** element instead of `undefined`.
2. A `null` result is coerced to `0` by the `>` / `<` comparison, so it can be picked as the extremum.

### Reproduction

```ts
import { maxBy, minBy } from 'es-toolkit/compat';
import _ from 'lodash';

maxBy([{ a: 1 }, { a: 2 }], 'b');        // es-toolkit: { a: 2 }   lodash: undefined
maxBy([{ a: null }, { a: null }], 'a');  // es-toolkit: { a: null } lodash: undefined
minBy([5, undefined, 3, null], x => x);  // es-toolkit: null        lodash: 3
```

## Fix

Mirror the guard already used by `max`/`min` so `null`, `undefined`, and symbol values are skipped. With those values skipped, the existing `=== undefined` sentinel becomes reliable again (it can only ever be set from a real comparable value).

```ts
if (current == null || Number.isNaN(current) || typeof current === 'symbol') {
  continue;
}
```

## Verification

- All existing `maxBy`/`minBy` specs still pass (Date, ±Infinity, string iteratees, NaN skipping, single/large arrays).
- Added regression tests covering missing keys and mixed `null`/`undefined`/numeric values; they fail on `main` and pass with the fix.
- Differentially fuzz-tested the patched functions against lodash over several thousand combinations of `undefined`/`null`/`NaN`/`±Infinity`/number/string values (property, function, and identity iteratees) with zero divergences.

### DIFF
--- a/src/compat/math/maxBy.spec.ts
+++ b/src/compat/math/maxBy.spec.ts
@@ -68,9 +68,20 @@ describe('maxBy', () => {
     expect(maxBy([NaN, NaN], x => x)).toBeUndefined();
   });
 
+  it('should skip symbol values', () => {
+    const sym = Symbol('a');
+    expect(maxBy([sym, 1, 3, 2], x => x)).toBe(3);
+    expect(maxBy([1, sym, 3, 2], x => x)).toBe(3);
+  });
+
+  it('should return undefined when every value is a symbol', () => {
+    expect(maxBy([Symbol('a'), Symbol('b')], x => x)).toBeUndefined();
+  });
+
   it('should skip null and undefined values, matching lodash', () => {
-    expect(maxBy([{ a: undefined }, { a: 5 }, { a: null }], 'a')).toEqual({ a: 5 });
-    expect(maxBy([5, undefined, 3, null], x => x)).toBe(5);
+    // Negative values so `null` (coerced to 0) would wrongly win as the max on the old code.
+    expect(maxBy([{ a: undefined }, { a: -5 }, { a: null }], 'a')).toEqual({ a: -5 });
+    expect(maxBy([-5, undefined, -3, null], x => x)).toBe(-3);
   });
 
   it('should return undefined when the iteratee yields no comparable value', () => {
--- a/src/compat/math/maxBy.spec.ts
+++ b/src/compat/math/maxBy.spec.ts
@@ -67,4 +67,16 @@ describe('maxBy', () => {
   it('should return undefined when every value is NaN', () => {
     expect(maxBy([NaN, NaN], x => x)).toBeUndefined();
   });
+
+  it('should skip null and undefined values, matching lodash', () => {
+    expect(maxBy([{ a: undefined }, { a: 5 }, { a: null }], 'a')).toEqual({ a: 5 });
+    expect(maxBy([5, undefined, 3, null], x => x)).toBe(5);
+  });
+
+  it('should return undefined when the iteratee yields no comparable value', () => {
+    // A missing key makes the iteratee return `undefined` for every element.
+    expect(maxBy([{ a: 1 }, { a: 2 }], 'b')).toBeUndefined();
+    expect(maxBy([{ a: undefined }, { a: undefined }], 'a')).toBeUndefined();
+    expect(maxBy([{ a: null }, { a: null }], 'a')).toBeUndefined();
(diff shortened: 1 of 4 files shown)


src/compat/math/maxBy.ts
```diff
@@ -51,7 +51,7 @@ export function maxBy<T>(items: ArrayLike<T> | null | undefined, iteratee: Value
     const element = array[i];
     const current = getValue(element, i, array);
 
-    if (Number.isNaN(current)) {
+    if (current == null || Number.isNaN(current) || typeof current === 'symbol') {
```
Antoliny0919, line 54: In Lodash, an error is only thrown when a Symbol appears after the first element. My guess is that this is actually an implementation bug in Lodash.
```
const sym = Symbol('x');
console.log(lodash.maxBy([{ a: 1 }, { a: sym }], 'a'));  // Error !!!
```

While compat should match Lodash's behavior as closely as possible, do you think it would be reasonable to make an exception for this case 🤔

src/compat/math/minBy.ts
```diff
@@ -51,7 +51,7 @@ export function minBy<T>(items: ArrayLike<T> | null | undefined, iteratee: Value
     const element = array[i];
     const current = getValue(element, i, array);
 
-    if (Number.isNaN(current)) {
+    if (current == null || Number.isNaN(current) || typeof current === 'symbol') {
```
Antoliny0919, line 54: It would be great to add some test cases for `Symbol` as well! (same as maxBy)

src/compat/math/maxBy.spec.ts
```diff
@@ -67,4 +67,16 @@ describe('maxBy', () => {
   it('should return undefined when every value is NaN', () => {
     expect(maxBy([NaN, NaN], x => x)).toBeUndefined();
   });
+
+  it('should skip null and undefined values, matching lodash', () => {
+    expect(maxBy([{ a: undefined }, { a: 5 }, { a: null }], 'a')).toEqual({ a: 5 });
```
Antoliny0919:
```suggestion
    expect(maxBy([{ a: undefined }, { a: -5 }, { a: null }], 'a')).toEqual({ a: -5 });
```
This test case covers the current change well.
However, it also passes on the previous implementation.

I think that's because it compares positive numbers with `null` and `undefined`. 
What do you think about using negative numbers instead? 
With negative values, the previous implementation would behave differently, so selecting the negative number would verify that the fix is actually working.

Antoliny0919 (commented):
Nice catch @Kropiunig !

It looks like there were a few compatibility behaviors that were missing!

The current implementation already looks really good, but I'd like to suggest a few improvements.

Kropiunig:
Thanks for the careful review!

Applied your negative-value suggestion to the `maxBy` test — with a negative maximum, `null` (coerced to `0`) would wrongly win on the old code, so the assertion now genuinely fails before the fix and passes after. For `minBy` I kept positive values, since that is the mirror case where `0` wrongly wins as the minimum; that test already failed on the old code. I also added Symbol coverage to both specs (I confirmed each new test fails on the pre-fix guard and passes now).

On the Symbol design question: I left the current skip-and-do-not-throw behavior in place. lodash only throws when a Symbol appears after the first element, which is position-dependent and looks like a bug, and never throwing feels safer for `compat` consumers who pass in mixed data. That said, this is your call — happy to switch to strict lodash parity and replicate the throw if you would rather match it exactly.

raon0211 (approved):
Thanks for the catch! Let's stick with the current implementation for simplicity.

PR #1928, opened 2026-07-21
fix(compat/map): make iteratee optional with identity default

kojesung:
## Summary

- Simplified the code by replacing the approach of providing the identity default value via the Nullish Coalescing Operator with explicitly specifying the identity default value in the _iteratee implementation signature.
- Applied the implementation signature to the map doc's type signature, and specified that _iteratee is optional along with its default value.

### DIFF
--- a/src/compat/array/map.ts
+++ b/src/compat/array/map.ts
@@ -135,8 +135,7 @@ export function map(
     return [];
   }
 
-  const keys: PropertyKey[] =
-    isArrayLike(collection) || Array.isArray(collection) ? range(0, collection.length) : Object.keys(collection);
+  const keys: PropertyKey[] = isArrayLike(collection) ? range(0, collection.length) : Object.keys(collection);
 
   const iteratee = iterateeToolkit(_iteratee);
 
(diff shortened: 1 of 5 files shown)


src/compat/array/map.ts
```diff
@@ -138,7 +138,7 @@ export function map(
   const keys: PropertyKey[] =
     isArrayLike(collection) || Array.isArray(collection) ? range(0, collection.length) : Object.keys(collection);
```
Antoliny0919: I think there's an additional optimization we can make here.
Do we need to check whether it's an array using `Array.isArray`? It seems like `isArrayLike` should be sufficient. 👍

Antoliny0919 (approved):
Thank you for work on this @kojesung ⭐️

LGTM !! I left small suggestion.

src/compat/array/map.ts
kojesung: As you mentioned, since isArrayLike can already detect arrays, and looking at isArrayLike's `@example` (which includes an array case), it's reasonable to interpret that covering arrays was an intended part of its design — so I concluded that isArrayLike alone is sufficient and applied the change.

raon0211 (approved):

PR #1930, opened 2026-07-21
test(compat/array): verify forEach skips non-enumerable properties

kojesung:
## Summary

The docs for each state that "it iterates through enumerable properties," but forEach.spec.ts doesn't have a test case covering this.

The current implementation is correct as-is since Object.keys(collection) only extracts enumerable keys, but this assumption could silently break if someone later changes the implementation to a method that also iterates non-enumerable properties. So I added a test case that creates a non-enumerable property with Object.defineProperty and verifies that forEach skips it.


raon0211 (approved):
Thanks!

PR #1933, opened 2026-07-21
refactor(compat/intersection): use shared toArray helper instead of Array.from

Antoliny0919:
## Change

Replace `Array.from` with `toArray`. 
Also, if the value is already an array, it would be more efficient to use it as is rather than applying `Array.from`.

### DIFF
--- a/src/compat/array/intersection.ts
+++ b/src/compat/array/intersection.ts
@@ -1,5 +1,6 @@
 import { intersection as intersectionToolkit } from '../../array/intersection.ts';
 import { uniq } from '../../array/uniq.ts';
+import { toArray } from '../_internal/toArray.ts';
 import { isArrayLikeObject } from '../predicate/isArrayLikeObject.ts';
 
 /**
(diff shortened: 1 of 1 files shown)


src/compat/array/intersection.ts
```diff
@@ -28,7 +29,7 @@ export function intersection<T>(...arrays: Array<ArrayLike<T> | null | undefined
     return [];
   }
 
-  let result: T[] = uniq(Array.from(arrays[0]));
+  let result: T[] = uniq(toArray(arrays[0]));
 
   for (let i = 1; i < arrays.length; i++) {
```
Antoliny0919, line 34:
```suggestion
  for (let i = 1; i < arrays.length && result.length > 0; i++) {
```
What do you think about adding a `result.length > 0` check? 
Once there's nothing left to compare, we can skip the rest of the loop. 
This is more efficient for that edge case, while in the general case it adds only a single extra check, which doesn't seem like it would have any meaningful performance impact.

raon0211:
Could you add a benchmark to check whether this doesn't hurt the original performance? You can run it with a command like `yarn bench intersection.bench.ts`

Antoliny0919:
> Could you add a benchmark to check whether this doesn't hurt the original performance? You can run it with a command like `yarn bench intersection.bench.ts`

## Before

스크린샷 2026-07-26 오후 4 15 54

## After

스크린샷 2026-07-26 오후 4 15 38

src/compat/array/intersection.ts
raon0211: We would like to keep the implementation simple and straightforward if possible, so I would like to stick with the current approach.

raon0211 (approved):

PR #1934, opened 2026-07-21
test(compat/intersectionBy): use intersectionBy instead of intersection in tests

Antoliny0919:
## Summary

The `intersectionBy` tests are testing `intersection` 🥲


raon0211 (approved):
Thanks!

Issue #1931, opened 2026-07-21
`trimStart` thowing on multi-character string `chars` is a breaking change and a Lodash compatibility break


foucdeg:
Hello,

Just some feedback about the fix https://github.com/toss/es-toolkit/pull/1615 introduced in v1.49 (`trimStart` throwing an error when `chars` is a multi-character string).
While I understand and agree with the intention, transforming an undeterminate behaviour into a thrown error should have been considered a breaking change. It did break something on our side.
IMO, the better way to introduce this in a minor release should have been a deprecation warning.

I also suggest editing the [docs](https://es-toolkit.dev/reference/string/trimStart.html#trimstart-str-chars) to make this constraint more explicit. While the examples only show one-character strings, the spec for the argument does not specify that it should be a one-character string. This is especially important for people migrating from Lodash (as Lodash's equivalent [does support](https://lodash.com/docs/4.18.1#trimStart) multi-character strings - this breaks the 1-1 compatibility promise).

Thanks for the hard work!

Hprogram:
Some findings that might help triage this:

**The compat package is not affected** — `es-toolkit/compat`'s `trimStart`/`trimEnd` split a string `chars` into a character array before delegating (`chars.toString().split('')`), so they keep lodash's "set of characters" semantics and never throw. The existing compat tests cover multi-character strings (e.g. `trimStart('-_-abc-_-', '_-')`). So the 1-1 lodash compatibility promise still holds for `es-toolkit/compat` — the throw only affects the main `es-toolkit` package (introduced in #1615 for `trimStart`; `trimEnd` has the same validation, and `trim` inherits it from both).

**The docs were inconsistent about it** — `trimEnd`'s reference docs already document the constraint with an `#### Errors` section in all four languages, but `trimStart` and `trim` didn't mention it in any language.

Whether the v1.49 behavior change should have been treated as breaking is of course a maintainer call — but regardless of that outcome, the docs should match the implementation, so I've submitted #1936 adding the missing `Errors` sections to `trimStart`/`trim` across the four languages, mirroring the existing `trimEnd` wording.

foucdeg:
Thanks, I didn't know about `es-toolkit/compat` - makes more sense now.

heeoneie:
A note on where this stands, since part of it moved after the report.

[...]

PR #1936, opened 2026-07-22
docs(trimStart, trim): document error thrown for multi-character string chars

Hprogram:
## Summary

Related to #1931.

`trimStart` and `trim` throw when `chars` is a string whose length is not 1 (validation added in #1615 for `trimStart`; `trimEnd` has the same validation and `trim` delegates to both), but only `trimEnd`'s reference docs document this with an `#### Errors` section. This adds the missing `Errors` sections to `trimStart` and `trim`, mirroring the existing `trimEnd` wording in each language.

## Changes

- Add `#### Errors` section to `trimStart.md` and `trim.md` in all four languages (en / ko / ja / zh_hans, 8 files), placed after `#### Returns` to match the structure of `trimEnd.md`

## Test results

- Docs-only change; wording, section placement, and punctuation mirror the existing `trimEnd.md` sections per language
- `yarn prettier --check` passes on all touched files


raon0211 (approved):
Thanks!

PR #1937, opened 2026-07-22
fix(partition): accept any predicate return value like Array.prototype.filter

### LINKED ISSUE #1926: Predicate types for `partition` (and other functions?)
While investigating switching from Lodash to es-toolkit, I encountered a TypeScript error with code similar to the following:

```ts
interface Value {
  name: string;
  good?: boolean;
}
function doSomething(values: Value[]) {
  const [good, bad] = partition(values, i => i.good);
}
```

Lodash's `partition` accepts this, because it only requires that its callback returns an `unknown`. es-toolkit's `partition` rejects it, because it strictly requires that its callback returns a `boolean`.

After working with similar tradeoffs in my own projects, I prefer the `unknown` approach: There are many cases where making a decision based on whether a value is truthy or falsy (not just `true` or `false`) is a common and expected idiom, and I've found very few cases where strictly requiring a `boolean` prevents bugs. (The main one, I suppose, would be catching the scenario of mistakenly using an async function - i.e., accepting a function returning `Promise<boolean>` when it should be a `boolean`.) And, perhaps more importantly, accepting `unknown` is compatible with the TS definitions for JavaScript functions like `Array.filter` - intuitively, I would expect a utility toolkit's array functions to accept the same predicates as JavaScript's built-in array functions.

Could `partition` and similar be updated to accept `unknown`?

(I realize I could use `es-toolkit/compat` in these cases, but I'm hoping to not rely on `compat` long term.)

Thank you.

Hprogram:
## Summary

Closes #1926.

`partition`'s fallback overload required the predicate to return a strict `boolean`, so common truthiness predicates like `partition(values, i => i.good)` (returning `boolean | undefined`) failed to compile — even though the implementation already branches on truthiness. This widens the fallback overload's return type from `boolean` to `unknown`, matching:

- TypeScript's own `Array.prototype.filter` signature (`predicate: (value, index, array) => unknown`), and
- `es-toolkit/compat`'s `partition`, which already types its callback as `(value: T) => unknown`.

The type-guard overload (`value is U`) is unchanged, so narrowing behavior is identical. The change is source-compatible (it strictly accepts more programs) and has no runtime effect.

## Changes

- `src/array/partition.ts`: fallback overload + implementation signature `=> boolean` → `=> unknown`, with JSDoc wording adjusted to "truthy/falsy value"
- `src/array/partition.spec.ts`: regression test using an optional-boolean property predicate (compile + runtime partitioning + `expectTypeOf`)
- `docs/**/array/partition.md` (en / ko / ja / zh_hans): parameter type updated to `=> unknown`

## Scope note

The issue title mentions "(and other functions?)" — other predicate-style callbacks (`dropWhile`/`dropRightWhile`, `takeWhile`/`takeRightWhile`, `remove`, `findKey`, `omitBy`/`pickBy`) have the same strict `boolean` requirement. This PR is scoped to `partition` per the issue's primary ask; happy to follow up on the others if you'd like the same treatment.

## Test results

- [x] `yarn vitest run src/array/partition.spec.ts` — 6 passed, including the new regression test
- [x] `yarn tsc --noEmit`
- [x] `yarn eslint src/array/partition.ts src/array/partition.spec.ts`
- [x] `yarn prettier --check` on all touched files

### DIFF
--- a/docs/ja/reference/array/partition.md
+++ b/docs/ja/reference/array/partition.md
@@ -45,7 +45,7 @@ const [truthy, falsy] = partition([], x => x > 0);
 #### パラメータ
 
 - `arr` (`T[]`): 2つのグループに分ける配列です。
-- `isInTruthy` (`(value: T, index: number, array: readonly T[]) => boolean`): 各要素が最初の配列(truthy)に含まれるか、2番目の配列(falsy)に含まれるかを決定する条件関数です。値、インデックス、配列全体と共に呼び出されます。
+- `isInTruthy` (`(value: T, index: number, array: readonly T[]) => unknown`): 各要素が最初の配列(truthy)に含まれるか、2番目の配列(falsy)に含まれるかを決定する条件関数です。値、インデックス、配列全体と共に呼び出されます。
 
 #### 戻り値
 
--- a/docs/ko/reference/array/partition.md
+++ b/docs/ko/reference/array/partition.md
@@ -45,7 +45,7 @@ const [truthy, falsy] = partition([], x => x > 0);
 #### 파라미터
 
 - `arr` (`T[]`): 두 그룹으로 나눌 배열이에요.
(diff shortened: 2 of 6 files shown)


raon0211 (approved):
Thanks!

PR #1938, opened 2026-07-22
fix(compat/findIndex): convert fromIndex to an integer like lodash

racgoo:
## Summary

`compat/findIndex` did not convert `fromIndex` with `toInteger` like lodash.
`Array.prototype.slice` truncates a float start index, but the relative index
was added back with the original float, so the result could be non-integer
(e.g. `2.5`).

```ts
findIndex([1, 2, 3, 4, 5], x => x >= 3, 1.5);
// before: 2.5
// lodash / after: 2


raon0211 (approved):
Thanks!

PR #1935, opened 2026-07-21
fix(compat/intersectionBy): resolve Lodash compatibility issues

Antoliny0919:
## Summary

fix Lodash compatibility issues in `intersectionBy`

### fb9ef1b

```ts
const array = [1, 2, 2, 3, 3];

console.log(lodash.intersectionBy(array)); // [1, 2, 3]
console.log(intersectionBy(array)); // [1, 2, 2, 3, 3]
```

### d78d963

```ts
const array = [2.1, 2.9, 1.2];

console.log(lodash.intersectionBy(array, Math.floor)); // [2.1, 1.2]
console.log(intersectionBy(array, Math.floor)); // [2.1, 2.9, 1.2]
```

PR #1939, opened 2026-07-22
 fix(compat/fill): truncate negative fractional positions like Lodash

gwagjiug:
## Summary                                                                                                                                                                                      
                                                                                                                                                                                              
compat/fill converted start and end with Math.floor. While this produces the expected result for positive fractional values, it rounds negative fractions toward negative infinity. For example, -1.5 became -2, whereas Lodash truncates it to -1.                                                                                                                                  
                                                                                                                                                                                              
This change uses the existing compat toInteger helper so that negative fractional positions are truncated toward zero, matching Lodash’s behavior and the integer-coercion pattern used by other compat array utilities.                                                                                                                                                                
                                                                                                                                                                                               #Changes                                                                                                                                                                                      
                                                                                                                                                                                              
- Replaced Math.floor with the shared toInteger helper for start and end.                                                                                                                    
- Added regression coverage for negative fractional start and end values.                                                                                                                    
- Kept the public API, mutation behavior, and return types unchanged.                                                                                                                        
                                                                                                                                                                                              
For example:                                                                                                                                                                                 
                                                                                                                                                                                              
 ```ts                                                                                                                                                                                        
   fill([1, 2, 3, 4], 0, -1.5);                                                                                                                                                               
 ```                                                                                                                                                                                          
                                                                                                                                                                                              
Before:                                                                                                                                                                                      
                                                                                                                                                                                              
 ```ts                                                                                                                                                                                        
   [1, 2, 0, 0]                                                                                                                                                                               
 ```                                                                                                                                                                                          
                                                                                                                                                                                              
 After, matching Lodash:                                                                                                                                                                      
                                                                                                                                                                                              
 ```ts                                                                                                                                                                                        
   [1, 2, 3, 0]                                                                                                                                                                               
 ```                                                                                                                                                                                          
                                                                                                                                                                                             Test results                                                                                                                                                                                 
                                                                                                                                                                                              
 - yarn vitest run src/compat/array/fill.spec.ts                                                                                                                                              
   - 1 test file passed                                                                                                                                                                       
   - 19 tests passed                                                                                                                                                                          
 - Compared 12 boundary cases directly against Lodash 4.18.1                                                                                                                                  
   - 0 mismatches after the change                                                                                                                                                            
 - yarn typecheck                                                                                                                                                                             
 - yarn eslint --config eslint.config.mjs src/compat/array/fill.ts src/compat/array/fill.spec.ts                                                                                              
   - 0 errors                                                                                                                                                                                 
   - 4 existing no-explicit-any warnings                                                                                                                                                      
 - yarn prettier --check src/compat/array/fill.ts src/compat/array/fill.spec.ts                                                                                                               
                                                                                                                                                                                              
Repository-wide checks currently report failures outside this diff:                                                                                                                          
                                                                                                                                                                                              
 - yarn vitest run                                                                                                                                                                            
   - 639 test files and 4,431 tests passed                                                                                                                                                    
   - src/compat/array/intersectionBy.spec.ts failed on duplicate removal, which is being addressed separately in #1935                                                                        
   - tests/check-dist.spec.ts could not pack the project because tsdown failed to import unrun                                                                                                
 - yarn lint                                                                                                                                                                                  
   - Failed on 8 existing curly errors in docs/.vitepress/components/FpLazySimulation.vue

### DIFF
--- a/src/compat/array/fill.spec.ts
+++ b/src/compat/array/fill.spec.ts
@@ -115,6 +115,11 @@ describe('fill', () => {
     ]);
   });
 
+  it('should truncate negative fractional `start` and `end` toward zero', () => {
+    expect(fill([1, 2, 3], 'a', -1.5)).toEqual([1, 2, 'a']);
+    expect(fill([1, 2, 3], 'a', 0, -1.5)).toEqual(['a', 'a', 3]);
+  });
+
   it('should work as an iteratee for methods like `_.map`', () => {
     const array = [
       [1, 2],
--- a/src/compat/array/fill.ts
+++ b/src/compat/array/fill.ts
@@ -3,6 +3,7 @@ import type { MutableList } from '../_internal/MutableList.d.ts';
 import type { RejectReadonly } from '../_internal/RejectReadonly.d.ts';
 import { isArrayLike } from '../predicate/isArrayLike.ts';
 import { isString } from '../predicate/isString.ts';
+import { toInteger } from '../util/toInteger.ts';
 
 /**
  * Fills an array with a value.
@@ -104,8 +105,8 @@ export function fill<T, U>(
     // prevent TypeError: Cannot assign to read only property of string
     return array;
   }
-  start = Math.floor(start);
-  end = Math.floor(end);
+  start = toInteger(start);
+  end = toInteger(end);
 
   if (!start) {
     start = 0;


gwagjiug:
## PR Change Explanation

> `es-toolkit/compat` preserves Lodash-compatible behavior for users migrating away from Lodash. This PR corrects one boundary-normalization edge case in `fill`: negative fractional `start` and `end` values previously shifted the effective boundary one index too far toward the beginning of the array. The public signatures and mutation model stay unchanged; only the conversion of range positions changes, with two regression assertions defining the new contract.

### Background

#### Background for readers new to this codebase

`fill(array, value, start?, end?)` mutates an array or array-like object by replacing the elements in a half-open range: `start` is included, while `end` is excluded. If the positions are omitted, the range is the whole array. Negative positions are resolved from the end of the array, so on a three-element array `start = -1` points to index `2`, while `end = -1` stops before index `2`.

The compatibility entry point exposes this function through `src/compat/compat.ts`, which is re-exported from `src/compat/index.ts`. The default compatibility toolkit receives the same export through `Object.assign(toolkit, compat)` in `src/compat/toolkit.ts`, so both the named `fill` export and `toolkit.fill` follow the same implementation.

The normal execution path has two layers:

1. `src/compat/array/fill.ts` handles compatibility concerns: invalid inputs, array-like objects, read-only strings, default positions, and position coercion.
2. It delegates to `src/array/fill.ts` as `fillToolkit`. That lower-level function resolves negative positions against `array.length`, clamps the range, and writes `value` for every index from `finalStart` up to—but not including—`finalEnd`.

#### Background specific to this change

Before this PR, the compatibility wrapper normalized both positions with `Math.floor`:

```ts
start = Math.floor(start);
end = Math.floor(end);
```

That happened before the lower-level implementation interpreted negative positions relative to the array length. `Math.floor` appears to work for the positive fractional cases already covered by the test suite: `0.1` becomes `0`, and `1.6` becomes `1`. Negative fractions expose a different rule, however: `Math.floor(-1.5)` is `-2`, because flooring always moves toward negative infinity. Removing the fractional portion should instead produce `-1` for Lodash-compatible position coercion.

The repository already has that coercion in `src/compat/util/toInteger.ts`. It converts the input to a finite number, computes `finite % 1`, and subtracts that remainder:

```ts
const remainder = finite % 1;
return remainder ? finite - remainder : finite;
```

For `-1.5`, the remainder is `-0.5`, so the result is `-1.5 - (-0.5) = -1`. Other compatibility range helpers, including `src/compat/array/slice.ts`, already use `toInteger` for `start` and `end`.

### Intuition

The key idea is to remove the fractional portion without moving a negative position farther left on the number line. A negative boundary is converted to an integer first and only then resolved relative to the array length, so changing `-1.5` into `-2` versus `-1` selects a different array boundary.

#### A minimal example

Consider a three-element array and a replacement value of `'a'`:

```ts
const array = [1, 2, 3];
```

For `start = -1.5`, the old and new paths resolve as follows:

```mermaid
flowchart LR
    I["Caller supplies start = -1.5"]
    I --> O["Before: floor to -2"]
    O --> OR["Resolve from length 3: 3 + -2 = index 1"]
    I --> N["After: remove the fraction to -1"]
    N --> NR["Resolve from length 3: 3 + -1 = index 2"]
```

Because filling continues from `start` to the end, the old path replaced two elements, while the new path replaces only the last element.

| Range input | Before normalization and result | After normalization and result |
| --- | --- | --- |
| `start = -1.5` | `-2` → start at index `1` → `[1, 'a', 'a']` | `-1` → start at index `2` → `[1, 2, 'a']` |
| `start = 0`, `end = -1.5` | `-2` → stop at index `1` → `['a', 2, 3]` | `-1` → stop at index `2` → `['a', 'a', 3]` |

The `end` example follows the same conversion rule, but the resolved boundary remains exclusive. An `end` of `-1` on a three-element array resolves to index `2`, so only indices `0` and `1` are replaced.

#### Before-and-after flow

- **Before:** validate the input → apply `Math.floor` → resolve negative positions relative to the array length → mutate the resolved range.
- **After:** validate the input → apply compatibility-layer `toInteger` coercion → resolve negative positions relative to the array length → mutate the resolved range.

Everything after position coercion is unchanged. The fix changes which integer boundary reaches the existing range-resolution logic.

### Code walkthrough

#### 1. Normalize both range boundaries with the shared compatibility rule

`src/compat/array/fill.ts` now imports `toInteger` and applies it to both `start` and `end`:

```ts
start = toInteger(start);
end = toInteger(end);
```

The surrounding guards remain in the same order. Non-array-like inputs still return `[]`, strings are still returned unchanged to avoid writing to read-only indices, and omitted positions still default to `0` and `array.length`. After coercion, the existing falsey-value normalization remains in place, and the wrapper delegates to `fillToolkit` exactly as before.

`src/array/fill.ts` is unchanged. It still computes negative boundaries as `length + position`, clamps them to the array bounds, and mutates the half-open range with a `for` loop. The behavioral difference therefore comes entirely from passing `-1` instead of `-2` into that established flow for a value such as `-1.5`.

The function overloads are also unchanged. Arrays, mutable array-like objects, `null`, and `undefined` keep the same TypeScript surface and return types; callers receive the corrected boundary behavior without adopting a new API.

#### 2. Keep every public access path on the same implementation

No export or caller wiring changes are required. `src/compat/compat.ts` continues to export this `fill`, `src/compat/index.ts` continues to expose it from `es-toolkit/compat`, and the default toolkit continues to attach the same function object. The normalization change therefore applies uniformly whether a caller uses the named export or accesses `fill` through the default compatibility toolkit.

#### 3. Define the regression contract for negative fractions

`src/compat/array/fill.spec.ts` adds one focused test with two assertions:

- `fill([1, 2, 3], 'a', -1.5)` must produce `[1, 2, 'a']`, proving that a negative fractional `start` truncates to `-1` before being resolved from the end.
- `fill([1, 2, 3], 'a', 0, -1.5)` must produce `['a', 'a', 3]`, proving the same rule for the exclusive `end` boundary.

These assertions sit next to the existing coercion coverage for positive fractions, numeric strings, and `NaN`, as well as the existing negative-integer boundary tests. Together, they distinguish the two relevant operations: positive inputs happen to produce the same integer under flooring and truncation, while negative fractional inputs establish why `toInteger` is required.

Antoliny0919 (approved):
Thank you for work on this @gwagjiug .
LGTM 👍

raon0211 (approved):

PR #1940, opened 2026-07-22
fix(compat/every): treat falsy doesMatch as property shorthand, not identity

kojesung:
## Summary

When determining the `predicate`, I changed the code that previously overwrote `doesMatch` with `identity` via an `if (!doesMatch)` condition — before going through the switch-case that determined it based on doesMatch's type — when doesMatch was a `falsy` value, to instead reuse the `iteratee()` util used by the same logic elsewhere (map, ...).

The previous problem, where `falsy` values such as `0, false, NaN, '', etc`. passed as `doesMatch` were all overwritten with `identity`, was designed such that it ignored the `falsy` value that should have been checked in a case like the one below, and unconditionally returned true, causing a problem where it returned a different result from lodash.

```tsx
// true
compat.every(
    [
      { 0: '', b: 1 },
      { 0: 'x', b: 1 },
    ],
    0
  )
```
```tsx
// false
lodash.every(
    [
      { 0: '', b: 1 },
      { 0: 'x', b: 1 },
    ],
    0
  )
```
In the `iteratee()` util, this problem can be resolved with the following code, which returns `identity` only when value is exactly null:
```tsx
if (value == null) {
    return identity;
}
```

[@3b6aa71](https://github.com/toss/es-toolkit/pull/1940/changes/3b6aa71421232beed07d1411b9281e4c5358ed02)
Added test cases to verify the scenario where, as in the code prior to this change, using a falsy value as the key would get caught by the wrong branch.

### DIFF
--- a/src/compat/array/every.ts
+++ b/src/compat/array/every.ts
@@ -2,10 +2,8 @@ import { identity } from '../../function/identity.ts';
 import { isIterateeCall } from '../_internal/isIterateeCall.ts';
 import { ListIterateeCustom } from '../_internal/ListIterateeCustom.ts';
 import { ObjectIterateeCustom } from '../_internal/ObjectIteratee.ts';
-import { property } from '../object/property.ts';
+import { iteratee } from '../compat.ts';
 import { isArrayLike } from '../predicate/isArrayLike.ts';
-import { matches } from '../predicate/matches.ts';
-import { matchesProperty } from '../predicate/matchesProperty.ts';
 
 /**
  * Checks if all elements in a collection pass the predicate check.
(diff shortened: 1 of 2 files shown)


src/compat/array/every.ts
```diff
@@ -102,33 +100,7 @@ export function every<T>(
...
-      }
-      break;
-    }
-    case 'symbol':
-    case 'number':
-    case 'string': {
-      predicate = property(doesMatch);
-    }
-  }
+  const predicate = iteratee(doesMatch ?? identity);
```
Antoliny0919, line 103: How about adding test cases for falsy values?
Since this is now behavior we guarantee for Lodash compatibility, I think it's worth adding explicit test coverage.

Antoliny0919 (commented):
Great work @kojesung 👍

Nice simplification by delegating the behavior to `iteratee`!
One additional thing: there's still a compatibility issue with the falsy value false.

```ts
console.log(lodash.every([{ false: 1 }, { false: 0 }], false)); // false
console.log(compatEvery([{ false: 1 }, { false: 0 }], false)); // error
```

kojesung:
> Great work @kojesung 👍
> 
> Nice simplification by delegating the behavior to `iteratee`! One additional thing: there's still a compatibility issue with the falsy value false.
> 
> ```ts
> console.log(lodash.every([{ false: 1 }, { false: 0 }], false)); // false
> console.log(compatEvery([{ false: 1 }, { false: 0 }], false)); // error
> ```

While working on this, I was able to confirm that it currently doesn't behave in a way that's fully compatible with lodash for boolean values either!

I analyzed that this is because a boolean case hasn't been added to the switch-case in iteratee, which is reused in many places across the codebase. Since that change is likely to have a much wider blast radius, I'm planning to tackle it in a separate branch. Thanks for the hint! @Antoliny0919

src/compat/array/every.ts
kojesung: I agree! I'll add tests for the newly added cover cases.

raon0211 (approved):
Thanks!

PR #1941, opened 2026-07-23
fix(compat/iteratee): treat boolean as a property shorthand, matching lodash

kojesung:
## Summary

- In the implementation of iteratee(), I added a test case to reveal the problem that, unlike string, symbol, and number types which were already being used as property name values, the boolean type was not being checked.
- To make that case pass, I changed the previous approach — defining separate cases for string, symbol, and number each — to using the default keyword instead. This is the same way lodash's baseIteratee handles the return when a value doesn't pass any of the conditionals.
- find and findLast use the iteratee fixed in this PR as is. However, there was a test in find and findLast that had asserted "an error should be thrown when a boolean predicate is used." But checking against real lodash, it returns undefined instead of an error. So I removed the existing test asserting "it should throw an error," and modified it to test that a boolean predicate behaves the same way as lodash.

### DIFF
--- a/benchmarks/bundle-size/find.spec.ts
+++ b/benchmarks/bundle-size/find.spec.ts
@@ -9,6 +9,6 @@ describe('find bundle size', () => {
 
   it('es-toolkit/compat', async () => {
     const bundleSize = await getBundleSize('es-toolkit/compat', 'find');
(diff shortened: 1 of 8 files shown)


dayongkr:
Thanks!!!

dayongkr (dismissed):

raon0211 (approved):

PR #1950, opened 2026-07-24
fix(compat/forEach): delete unnecessary array check

DongEun02:
## Summary 

Both `isArrayLike` and `Array.isArray` are being checked, but the `Array.isArray` check is redundant because arrays already satisfy the conditions of `isArrayLike`.
Additionally, unnecessary checks for objects have been removed to improve performance.

### DIFF
--- a/src/compat/array/forEach.ts
+++ b/src/compat/array/forEach.ts
@@ -152,8 +152,7 @@ export function forEach<T>(
     return collection;
   }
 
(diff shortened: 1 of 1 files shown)


raon0211 (approved):

Antoliny0919:
Thank you @DongEun02 and Welcome onboard ⭐️

PR #1955, opened 2026-07-26
docs(compat/lastIndexOf): correct native method reference and mark fromIndex optional

Antoliny0919:
## Summary

It looks like the prototype and optional notation are missing !!


raon0211 (approved):
Thanks!

PR #1956, opened 2026-07-26
test(compat/lastIndexOf): add falsey fromIndex handling test case

Antoliny0919:
## Summary

It looks like we're missing the Lodash test case that covers a falsey fromIndex.

https://github.com/lodash/lodash/blob/6a2cc1dfcf7634fea70d1bc5bd22db453df67b42/test/findLastIndex-and-lastIndexOf.spec.js#L54


raon0211 (approved):
Thanks!

PR #1957, opened 2026-07-26
refactor(compat/lastIndexOf): use internal toArray helper for array like conversion

Antoliny0919:
## Summary

Use toArray to avoid unnecessary conversions for arrays.

### Before

스크린샷 2026-07-26 오후 6 01 40

### After

스크린샷 2026-07-26 오후 6 02 00

### DIFF
--- a/benchmarks/bundle-size/lastIndexOf.spec.ts
+++ b/benchmarks/bundle-size/lastIndexOf.spec.ts
@@ -9,6 +9,6 @@ describe('lastIndexOf bundle size', () => {
 
   it('es-toolkit/compat', async () => {
     const bundleSize = await getBundleSize('es-toolkit/compat', 'lastIndexOf');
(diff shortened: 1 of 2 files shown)


dayongkr:
I got about 2x too on array input.

One thing though — this changes what happens with sparse arrays (the ones with empty slots). `toArray` gives the array back as it is, so the native `lastIndexOf` runs on the original and skips the empty slots. `Array.from` used to fill them with `undefined`:

```js
lastIndexOf([1, , 1], undefined, 2)
// lodash: 1   main: 1   this PR: -1
```

(I pass `fromIndex` here because lodash returns `length` when you leave it out and look for `undefined`)

Maybe there's another way to make it faster — lodash never turns the input into an array at all, it just goes backwards by index, so empty slots never matter. Either way, a test for this case would help.

raon0211 (approved):

PR #1958, opened 2026-07-26
fix/compat array find array like length

kojesung:
## Summary

- The existing `find` used the `if (!Array.isArray(source))` statement to treat anything that wasn't an array as an object and process it accordingly. However, this branch handling, when an object has a readable property called `length`, uses the property name `length` as a key to perform the search. But this differs from lodash's intended behavior. When lodash encounters a `length` property — such as the `length` that exists on real arrays — it does not use it as a key and ignores it instead.
```tsx
const arrayLike = { 0: 'a', 1: 'b', length: 2 };
console.log(findCompat(arrayLike, v => v === 2));
// 2

console.log(lodash.find(arrayLike, v => v === 2));
// undefined
```
- To follow this behavior of lodash, instead of `Array.isArray`, I applied the implementation approach of `.map`, which uses the `isArrayLike` function to extract keys excluding the length property when it exists, and treats only the extracted keys as the iteration target. Through this, I implemented it so that, just like lodash's implementation, when a length property exists it is not used as a key and is ignored.

- In this process, I also removed the structure where the array branch delegated to native code via `.slice(fromIndex).find(doesMatch);` as a side effect, this also fixed the `index/collection` arguments passed to the predicate when using `fromIndex` so that they are now correctly passed based on the original array (rather than the sliced array).

- I added a test case that can verify this work.

- As the implementation approach changed, the shape of the error message that the `'should throw error when boolean predicate is used'` test was verifying ended up changing. Because the entity that throws the error changed, I modified the assertion message so the test would pass. For reference, actual lodash does not throw an error for a boolean predicate and instead returns `undefined;` this behavior itself (throwing an error) was judged to be the domain of the `iteratee` function, so it was not included in the scope of this fix.


raon0211 (approved):
Thanks!

raon0211 (dismissed):

raon0211 (approved):

PR #1969, opened 2026-07-28
refactor(compat/findIndex): reuse iteratee() instead of duplicating shorthand dispatch logic

kojesung:
## Summary

The logic that branched via a switch-case based on the type of findIndex's `doesMatch` was duplicated with `iteratee.ts`, so it has been unified into a way that utilizes a single `iteratee()` function.


src/compat/array/findIndex.ts
```diff
@@ -1,8 +1,6 @@
 import { ListIterateeCustom } from '../_internal/ListIterateeCustom.ts';
+import { iteratee as iterateeToolkit } from '../compat.ts';
```
raon0211:
```suggestion
import { iteratee as iterateeToolkit } from '../util/iteratee.ts';
```

raon0211 (approved):

PR #1970, opened 2026-07-29
docs(compat/orderBy): update documentation to clarify 'orders' parameter options

DongEun02:
## Summary

The `orderBy` documentation incorrectly stated that `true` sorts in
ascending order and `false` sorts in descending order.

Internally, each order value is converted to a string, and `compareValues`
uses descending order only when that value is exactly `'desc'`. All other
values follow the ascending-order branch. As a result, both `true` and
`false` sort in ascending order.

Lodash also documents `orders` as a string array using `'asc'` and `'desc'`:
https://lodash.com/docs/#orderBy

## Changes

- Removed the misleading `true` and `false` sort-order descriptions.
- Kept `'asc'` and `'desc'` as the documented sort-order options.
- Updated the English, Korean, Japanese, and Simplified Chinese documentation.


dayongkr (approved):
Thanks for your contribution!

PR #1980, opened 2026-08-01
fix[without]: Narrow return type for literal exclusions

### LINKED ISSUE #1503: `without` should `Exclude` from return type.

SR0725:
## Summary

Closes #1503

`without` currently returns `T[]` even when a readonly literal input and literal exclusions make a narrower element type known. This adds a const rest-tuple overload so `without(['a', 'b', 'c'] as const, 'a')` returns `Array<'b' | 'c'>`.

Calls with no exclusions keep the input element type. A spread dynamic exclusion array also keeps the input type instead of narrowing to `never[]`.

## Changes

- infer literal exclusion values in the public `without` type
- add type tests for literal, empty, and dynamic exclusion lists
- document the inferred return type in all four reference languages

## Test results

- `yarn vitest run --reporter=dot` — 660 files passed; 4,552 tests passed and 2 skipped
- `yarn typecheck`
- `yarn build`
- `yarn workspace docs docs:build`
- targeted ESLint and Prettier checks for the changed files

Full `yarn lint` still reports eight existing `curly` errors in `docs/.vitepress/components/FpLazySimulation.vue`; neither that file nor the lint configuration changes here.

_OpenAI Codex assisted with this contribution. The type behavior, generated declarations, tests, and documentation build were verified locally._

### DIFF
--- a/docs/ja/reference/array/without.md
+++ b/docs/ja/reference/array/without.md
@@ -33,6 +33,16 @@ without([1, NaN, 3, NaN, 5], NaN);
 // Returns: [1, 3, 5]
 ```
 
+入力が読み取り専用のリテラル配列の場合、TypeScript は戻り値の要素型からも指定したリテラル値を除外します。
+
+```typescript
+import { without } from 'es-toolkit/array';
+
+const letters = ['a', 'b', 'c'] as const;
+const filtered = without(letters, 'a');
+// Type: Array<'b' | 'c'>
+```
+
 #### パラメータ
 
 - `arr` (`readonly T[]`): 値を削除する配列です。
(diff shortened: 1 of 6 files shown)


raon0211:
Thanks so much for taking the time to work on this, including the type tests and docs in all four languages! 

After careful consideration, we've decided not to move forward with this change. The type declaration becomes significantly more complex, while the narrowing only benefits readonly literal (as const) arrays with inline literal values. As described in [CONTRIBUTING.md](https://github.com/toss/es-toolkit/blob/main/.github/CONTRIBUTING.md), es-toolkit focuses on providing the simplest interface for the common 85% of use cases, and we feel this trade-off isn't the right fit.

Closing this PR, but again, thank you for the contribution — we'd love to see you around for other issues!

PR #1983, opened 2026-08-01
docs[usage]: document browser support

### LINKED ISSUE #800: docs: browsers support
Hi! 

I really like this library, but I can't find out which browsers it supports and which it doesn't. My guess is that support for older browsers is not a primary goal here, and I understand that. But it would be great to have this information somewhere in the docs – a list of supported browsers or a note warning about it.

xianjianlf2:
## Summary
Fixes #800

Document supported browser expectations for es-toolkit usage.

## Changes
- Add an ES2015 (ES6) browser support note to the browser usage docs.
- Mention transpilation and polyfills for older browser targets.
- Keep the English, Korean, Japanese, and Simplified Chinese usage pages in sync.

## Test results
- `git diff --check`
- Not run: test suite (docs only).

### DIFF
--- a/docs/ja/usage.md
+++ b/docs/ja/usage.md
@@ -70,6 +70,8 @@ bun add es-toolkit
 
 [jsdelivr](https://www.jsdelivr.com)や[unpkg](https://unpkg.com)などのCDNからes-toolkitを使用できます。Lodashと同様に、`_`変数にすべての関数が含まれています。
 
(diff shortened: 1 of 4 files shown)


dayongkr:
Thanks for the PR! Documenting this is the right call — #800 is a fair ask.

That said, an ES2015 note alone still feels too vague for users to actually feel safe about what runs where. Let me think a bit more about how to phrase it — and ideas are welcome in the meantime!

dayongkr:
es-toolkit targets modern browsers, so the syntax level we ship can rise over time. The current build already emits ES2022 syntax (class fields in `dist/promise/semaphore.mjs`, `??` in `dist/array/countBy.mjs`), so pinning a version here would go stale.

For browser use we recommend downleveling es-toolkit in your own build and adding polyfills for the APIs the functions you import rely on.

We'll think about how to document this properly. Holding off on merging for now.

PR #1989, opened 2026-08-03
fix[mean, meanBy, sum, sumBy]: Update to take `ArrayLike`

### LINKED ISSUE #1988: Permit `ArrayLike<number>` with `mean` and `sum`
### es-toolkit version

1.50.0

### Entry point

es-toolkit

### Reproduction

```typescript
import { sum } from 'es-toolkit';
const x = new Float32Array([21, 31]);
sum(x);
```

Current: TypeScript compilation error

```
Argument of type 'Float32Array<ArrayBuffer>' is not assignable to parameter of type 'readonly number[]'.
  Type 'Float32Array<ArrayBuffer>' is missing the following properties from type 'readonly number[]': concat, flatMap, flat, toSpliced, [Symbol.unscopables]
```

Expected: Success

Nothing in `sum` or `mean`'s implementation requires an actual array, and support for JS TypedArrays would be valuable and appreciated. Can they be updated to take `ArrayLike<number>` instead?

### Validations

joshkel:
## Summary

Update `mean`, `meanBy`, `sum`, `sumBy` to take `ArrayLike`.

Fixes #1988

## Changes

* Broaden TypeScript types
* Add tests to verify TypedArray compatibility in particular

### DIFF
--- a/src/math/mean.spec.ts
+++ b/src/math/mean.spec.ts
@@ -9,4 +9,10 @@ describe('mean', () => {
   it('returns NaN for empty arrays', () => {
     expect(mean([])).toEqual(NaN);
   });
+
+  it('works with TypedArrays', () => {
+    const typedArray = new Float32Array([1.5, 2.5]);
+    const result = mean(typedArray);
+    expect(result).toBe(2.0);
+  });
 });
--- a/src/math/mean.ts
+++ b/src/math/mean.ts
@@ -5,7 +5,7 @@ import { sum } from './sum.ts';
  *
  * If the array is empty, this function returns `NaN`.
  *
- * @param nums - An array of numbers to calculate the average.
+ * @param nums - An array-like of numbers to calculate the average.
  * @returns The average of all the numbers in the array.
  *
  * @example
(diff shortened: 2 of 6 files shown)


raon0211:
Thank you for the pull request and for taking the time to write tests and update the JSDoc as well. The implementation itself is clean, and we appreciate that you filed #1988 first to explain the motivation.

That said, we've decided not to take this change. Almost every function in the main `es-toolkit` package takes `readonly T[]`, and broadening only `mean`, `meanBy`, `sum`, and `sumBy` to `ArrayLike<T>` would make these four functions inconsistent with the rest of the library. Keeping the API surface uniform is something we weigh heavily, so we're going to keep these signatures as they are.

The good news is there are two easy ways to get what you need today:

```ts
// Option 1: es-toolkit/compat already accepts ArrayLike
import { sum, mean } from 'es-toolkit/compat';
sum(new Float32Array([21, 31])); // works

// Option 2: convert at the call site
import { sum } from 'es-toolkit';
sum(Array.from(float32Array));
```

Thanks again for contributing!

PR #1991, opened 2026-08-03
fix[omit, pick]: Allow union types

### LINKED ISSUE #1990: `omit` / `pick` with union types
### es-toolkit version

1.50.0

### Entry point

es-toolkit

### Reproduction

```typescript
import { omit, pick } from 'es-toolkit';
interface A { type: 'A'; a: number };
interface B { type: 'B'; b: number };
function test(x: A | B) {
  pick(x, ['a']);
  omit(x, ['b']);
}
```

Current: TypeScript errors:

```
Type '"a"' is not assignable to type '"type"'.
Type '"b"' is not assignable to type '"type"'.
```

Expected: Ideally, es-toolkit's types would handle type unions by accepting keys that satisfied any of the component types.

### Validations

joshkel:
## Summary

Revise types for omit and pick to allow union types.

Note: This changes the return type from, e.g., `Omit<A|B, K>` to a distributed union. This may technically be a breaking change but should be unlikely to cause issues in practice.

Fixes #1990

## Changes

- Change pick and omit to use new DistributedPick and DistributedOmit types so that they can support union types. Previously, they only accepted keys present on all union components; now they, accept keys of any union component.
- Introduce an internal helper type KeysOfUnion

### DIFF
--- a/src/_internal/KeysOfUnion.ts
+++ b/src/_internal/KeysOfUnion.ts
@@ -1,0 +1,8 @@
+/**
+ * The keys of every member of `T`.
+ *
+ * Unlike `keyof T`, which for a union only yields the keys shared by all
+ * members, this distributes over each member, so a key that exists on some
+ * members is included.
+ */
+export type KeysOfUnion<T> = T extends unknown ? keyof T : never;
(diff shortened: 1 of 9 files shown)


raon0211:
Thank you for the PR! That said, I actually think this is a limitation of TypeScript itself. For example:

```ts
type A = { type: 'a'; a: number };
type B = { type: 'b'; b: string };

const value: A | B = { type: 'a', a: 1 };

value.b; // type error
```

Here, `value.b` cannot be referenced safely — TypeScript only allows accessing properties shared by all members of a union.

This is a relatively minor use case, and as CONTRIBUTING.md describes, we aim for simple implementations optimized for the major ~80% of use cases. So I'm going to close this PR.

PR #1992, opened 2026-08-03
fix(compat): render nested nullish values in toString like lodash

mariazuheros:
## Summary

`compat/toString` renders nested nullish array values differently from lodash.

The array branch maps over the elements using the public `toString`, whose top guard turns `null` and `undefined` into `''`. lodash only applies that nullish guard at the top level; internally its `baseToString` renders nested `null` as `"null"` and `undefined` as `"undefined"`.

```ts
toString([1, null, 3])
// es-toolkit: "1,,3"
// lodash:     "1,null,3"

toString([null, undefined])
// es-toolkit: ","
// lodash:     "null,undefined"
```

## Fix

Split the public nullish guard from the recursive conversion by adding a private `baseToString`, so the top level still returns `''` for `null`/`undefined` while nested values are stringified. This mirrors lodash's `toString`/`baseToString` split.

## Tests

Added regression cases for nested nullish values in arrays and nested arrays. All existing cases (nullish top level, sign of `-0`, symbols) still pass. Verified against lodash 4.17.21.

## Bundle size snapshots

Adding the private `baseToString` helper slightly increases the compiled size of `toString`, so the bundle-size snapshots of the compat functions that import it (`add`, `camelCase`, `escape`, `escapeRegExp`, `find`, `findKey`, `mapKeys`, `mapValues`, `omit`, `zipObjectDeep`) were regenerated to match.

### DIFF
--- a/benchmarks/bundle-size/add.spec.ts
+++ b/benchmarks/bundle-size/add.spec.ts
@@ -9,6 +9,6 @@ describe('add bundle size', () => {
 
   it('es-toolkit/compat', async () => {
     const bundleSize = await getBundleSize('es-toolkit/compat', 'add');
-    expect(bundleSize).toMatchInlineSnapshot(`455`);
+    expect(bundleSize).toMatchInlineSnapshot(`472`);
   });
 });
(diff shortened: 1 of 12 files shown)


mariazuheros:
Sorry for the noise on the history here. The earlier commits were accidentally attributed to the wrong GitHub account, so I have corrected the author and they now show under my account. Nothing else changed, the fix and the snapshots are the same.

raon0211 (approved):
Thanks!

PR #1994, opened 2026-08-04
feat(SingleFlight, KeyedSingleFlight): share in-flight async work with duplicate callers

seungrodotlee:
## Summary

Adds `SingleFlight` and `KeyedSingleFlight` to `es-toolkit/promise`: two concurrency primitives that let duplicate callers join an operation that is already running instead of starting a second one.

Refs #1608.

When several call sites independently ask for the same thing at the same time — a config fetch on app boot, a token refresh from three parallel requests, the same user record from two components — the work runs N times. These share one Promise instead.

```ts
import { SingleFlight, KeyedSingleFlight } from 'es-toolkit';

const config = new SingleFlight();
config.run(() => fetch('/api/config')); // runs
config.run(() => fetch('/api/config')); // joins the first — same Promise

const users = new KeyedSingleFlight<string>();
users.run('user:1', fetchUser1); // runs
users.run('user:1', fetchUser1); // joins
users.run('user:2', fetchUser2); // separate key, runs concurrently
```

They join `Mutex` and `Semaphore` as class-based primitives in `es-toolkit/promise`.

## Changes

- `src/promise/singleFlight.ts` — `SingleFlight` with a single `run(task)` method.
- `src/promise/keyedSingleFlight.ts` — `KeyedSingleFlight<K>` with `run(key, task)`, backed by a `Map<K, SingleFlight>`.
- Re-exported from `src/promise/index.ts` (so also from the `es-toolkit` root and the `es-toolkit/promise` subpath).
- Tests for both, and reference docs in all four languages.

## Design notes

**Results are not cached.** Once a flight settles, the next call starts fresh. This is deduplication of concurrent work, not memoization — `memoize` already covers caching, and conflating the two would mean every user needs a TTL knob.

**The Promise is stored before the task runs.** The naive implementation calls the factory first and stores the result, which leaves a window where a reentrant call — a task that calls `run` on the same instance — sees an empty slot and starts a second flight. Here the Promise is created with an explicit resolver pair and assigned to the instance *before* `task()` is invoked, so a reentrant caller joins the flight it is already inside. #1608 raised this exact bug; `singleFlight.spec.ts` covers it.

Note that storing first makes the microtask hop unnecessary: `task()` is called synchronously, so the underlying request is dispatched on the current tick rather than one tick later.

**A synchronous throw becomes a rejection.** `run(() => { throw e })` returns a rejected Promise rather than throwing at the call site, so callers have one error channel.

**Rejections are shared, then the slot clears.** Every caller that joined a failed flight sees the same error; the next call after settlement retries. Clearing is guarded by identity (`if (this.currentPromise === promise)`), so a settling old flight cannot evict a newer one.

Cleanup uses `promise.then(onFulfilled, onRejected)` rather than `promise.finally(...)` on purpose. `finally` returns a derived Promise that re-propagates the original rejection, and nothing handles that derived Promise — which surfaces as a spurious `unhandledRejection` even when the caller correctly catches the error it received. The two-callback form returns a Promise that fulfills, so no such warning is produced.

**`KeyedSingleFlight` deletes its Map entry on settle**, also identity-guarded, so the Map does not grow without bound over the life of the instance. Keys use `Map` equality, which means object keys compare by reference.

**State is per instance, not module-scope.** A module-level registry makes the key namespace global: two unrelated features that both pick `'user'` collide, tests cannot be isolated from one another, and on a server where the module is shared across requests, two concurrent requests with the same key receive each other's result. Scoping to an instance means callers decide the sharing boundary — one instance per app for a genuinely global resource, one per request where that is what isolation requires.

## Relationship to #1921

#1921 implements the same idea from #1608 as `coalesce(key, fn)` with a module-scope `Map`. The core semantics agree with this PR — dedupe concurrent calls, no result caching, identity-guarded cleanup, reentrancy safety. I have left a comment there with the details; two behaviors are worth repeating here, since they are what motivated the shape of this API.

The module-scope registry shares state across every consumer in the process:

```ts
// two independent callers, one global Map
const a = coalesce('current-user', () => Promise.resolve({ user: 'alice' }));
const b = coalesce('current-user', () => Promise.resolve({ user: 'bob' }));
await b; // { user: 'alice' } — the second caller's own fn never runs
```

That is the intended dedupe behavior when the two calls really are the same operation, but there is no way to express that they are not. On a server rendering two requests concurrently, the second request receives the first request's data.

And `finally`-based cleanup emits an `unhandledRejection` even when the caller handles the error:

```ts
try { await coalesce('k', () => Promise.reject(new Error('boom'))); }
catch (e) { /* handled */ }
// caller caught properly: boom
// !!! UNHANDLED REJECTION: boom
```

Both reproduce on Node; `KeyedSingleFlight` returns the per-caller result in the first case and emits nothing in the second.

On naming: `coalesce` collides with SQL's `COALESCE` (first non-null argument), which is a different and more widely known operation. `SingleFlight` follows `golang.org/x/sync/singleflight`, where this pattern is named, and reads consistently next to `Mutex` and `Semaphore`. Splitting the keyed and unkeyed cases also means the common single-resource case does not have to invent a key, and the keyed case is not restricted to `string`.

Happy to fold this into #1921 instead if the maintainers prefer that path — the two PRs should not both land.

## Note on the discussion requirement

#1608 is still open and labelled `p3: discussion` rather than accepted, so per CONTRIBUTING §4.1 this may be premature. Opening it anyway because #1921 is an in-progress implementation of the same request and the two defects above seemed worth surfacing with a working alternative rather than as review comments alone. If the proposal is not wanted, feel free to close — no objection.

No benchmark section: these have no Lodash counterpart to compare against, same as `Mutex` and `Semaphore`.

### DIFF
--- a/docs/ja/reference/promise/KeyedSingleFlight.md
+++ b/docs/ja/reference/promise/KeyedSingleFlight.md
@@ -1,0 +1,47 @@
+# KeyedSingleFlight
+
+キーごとに非同期タスクを1つだけ実行し、重複した呼び出しに実行中のPromiseを共有します。
+
+```typescript
+const flight = new KeyedSingleFlight<string>();
+```
+
+## 使用法
+
+### `KeyedSingleFlight<K>()`
+
+キーごとにタスクを重複排除し、異なるキーのタスクは並行して実行したい場合に`KeyedSingleFlight`を使用します。
+
+```typescript
+import { KeyedSingleFlight } from 'es-toolkit';
+
+const flight = new KeyedSingleFlight<string>();
+
+const first = flight.run('user:1', async () => {
+  const response = await fetch('/api/users/1');
+  return response.json();
+});
+
+const second = flight.run('user:1', async () => {
+  const response = await fetch('/api/users/1');
+  return response.json();
+});
+
+const other = flight.run('user:2', async () => {
+  const response = await fetch('/api/users/2');
+  return response.json();
+});
+
+console.log(first === second); // true
+console.log(first === other); // false
+```
+
+タスクが成功した場合も失敗した場合も同じPromiseを共有します。タスクが完了すると結果はキャッシュされず、同じキーの次の呼び出しでは新しいタスクを実行します。キーは`Map`と同じ規則で比較されるため、オブジェクトキーは参照が同じ場合のみ一致します。
+
+#### パラメータ
+
+- `K`（`unknown`）：タスクを識別するキーの型です。
+
+#### メソッド
+
+- `run`（`(key: K, task: () => Promise<T>) => Promise<T>`）：タスクを実行するか、そのキーで実行中のタスクのPromiseを返します。
--- a/docs/ja/reference/promise/SingleFlight.md
+++ b/docs/ja/reference/promise/SingleFlight.md
@@ -1,0 +1,37 @@
+# SingleFlight
+
+同時に実行される非同期タスクを1つに制限し、重複した呼び出しに実行中のPromiseを共有します。
+
+```typescript
+const flight = new SingleFlight();
+```
+
+## 使用法
+
+### `SingleFlight()`
+
+複数の呼び出し元が同じ処理を同時に開始する可能性があり、実際の処理は1回だけ実行したい場合に`SingleFlight`を使用します。
+
+```typescript
+import { SingleFlight } from 'es-toolkit';
+
+const flight = new SingleFlight();
+
+const first = flight.run(async () => {
+  const response = await fetch('/api/config');
+  return response.json();
+});
+
+const second = flight.run(async () => {
+  const response = await fetch('/api/config');
+  return response.json();
+});
+
+console.log(first === second); // true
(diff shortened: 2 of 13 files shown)


dayongkr:
Thanks for digging up an older issue and thinking it through before opening this!

That said, #1608 is still labeled `p3: discussion` with no maintainer response, so the proposal hasn't been accepted yet. Our contributing guide asks for that before a new function lands:

> **Open a discussion first and wait until it is accepted.** A pull request that adds a function without one will be closed, even when the implementation itself is good.

([CONTRIBUTING §4.1](https://github.com/toss/es-toolkit/blob/main/.github/CONTRIBUTING.md#adding-a-new-function))

So I'm closing this PR, but the idea stays open in #1608.

For context on how we decide: we mostly look at how much traction the issue gets, and whether the pattern is already widely used across OSS projects and production services. Adding new functions isn't a high priority for us right now either, so it would help to see some of that first before we take the discussion further.

PR #1999, opened 2026-08-06
fix(compat): convert non-array-like iterables in toArray like lodash

mariazuheros:
## Summary

`compat/toArray` ignores a non-array-like object's `Symbol.iterator` and falls straight through to `Object.values`, so iterables are not converted the way lodash converts them.

```ts
toArray(new Map([['a', 1]]).entries())
// es-toolkit: []
// lodash:     [['a', 1]]

toArray(new Set([1, 2]).values())
// es-toolkit: []
// lodash:     [1, 2]

const iterable = { *[Symbol.iterator]() { yield 1; yield 2; } };
toArray(iterable)
// es-toolkit: []
// lodash:     [1, 2]
```

lodash's `toArray` handles the iterator (`symIterator && value[symIterator]`) before falling back to `values`.

## Fix

Convert iterables with `Array.from` before the `Object.values` fallback, mirroring lodash's ordering. Array-likes, `Map` and `Set` are still handled first, and plain non-iterable objects still use `Object.values`.

## Tests

Added a regression case for non-array-like iterables (a generator object, `Map` entries and `Set` values). All existing cases still pass. Verified against lodash 4.17.21.

### DIFF
--- a/src/compat/util/toArray.ts
+++ b/src/compat/util/toArray.ts
@@ -60,6 +60,10 @@ export function toArray(value?: unknown): any[] {
   }
 
   if (typeof value === 'object') {
+    if (typeof (value as Record<symbol, unknown>)[Symbol.iterator] === 'function') {
+      return Array.from(value as Iterable<unknown>);
+    }
+
     return Object.values(value);
   }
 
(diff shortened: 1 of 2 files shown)


src/compat/util/toArray.ts
```diff
@@ -60,6 +60,10 @@ export function toArray(value?: unknown): any[] {
   }
 
   if (typeof value === 'object') {
+    if (typeof (value as Record<symbol, unknown>)[Symbol.iterator] === 'function') {
```
raon0211: You could use [isIterable](https://es-toolkit.dev/reference/predicate/isIterable.html#isiterable) in our library.

raon0211 (approved):

PR #2025, opened 2026-08-15
docs(zh_hans): remove stale lodash compatibility sections from spread and throttle

D-Sketon:


dayongkr:
Thanks for you contribution!

PR #2037, opened 2026-08-19
fix[omitBy, pickBy]: type numeric callback keys as strings

### LINKED ISSUE #2036: `omitBy` / `pickBy` with number keys
### es-toolkit version

1.50.0

### Entry point

es-toolkit

### Reproduction

```typescript

const foo = {
  [2]: 'two',
};
omitBy(foo, (_value, key: number) => {
  console.log(JSON.stringify(key));
  return true;
});
```

The type definitions for `omitBy` and `pickBy` state that the `key` parameter through to their callback is of type `keyof T`. However, this is incorrect for `Record<number, unknown>` and any similar number-keyed objects, since JavaScript coerces any numeric object keys to strings.

I believe that the actual callback should always be a subset of string. (It seems impractical for `omitBy` / `pickBy` to handle converting back to a numeric type themselves.)

### Validations

ProgWon:
## Summary

Fixes #2036.

`omitBy` and `pickBy` use `Object.keys`, so numeric object keys are passed to predicates as strings at runtime. For number-keyed objects, their types currently expose those keys as numbers. This change aligns the callback key type with the runtime behavior.

## Changes

- Use `StringKeyOf<T>` for predicate keys, preserving string literal keys and converting numeric keys to strings.
- Add runtime and type-level regression tests.
- Update documentation in all four supported languages.

### DIFF
--- a/docs/ja/reference/object/omitBy.md
+++ b/docs/ja/reference/object/omitBy.md
@@ -34,7 +34,7 @@ const nonAdmins = omitBy(data, (value, key) => key.startsWith('admin'));
 #### パラメータ
 
 - `obj` (`T extends Record<string, any>`): プロパティをフィルタリングするオブジェクトです。
-- `shouldOmit` (`(value: T[keyof T], key: ObjectKeys<T>) => boolean`): プロパティを除外するかを決定する条件関数です。値とキーを受け取り、除外する場合は`true`、維持する場合は`false`を返します。数値キーは条件関数に文字列として渡されます。
+- `shouldOmit` (`(value: T[keyof T], key: ObjectKeys<T>) => boolean`): プロパティを除外するかを決定する条件関数です。値とキーを受け取り、除外する場合は`true`、維持する場合は`false`を返します。数値キーは条件関数に文字列として渡されます([`ObjectKeys<T>`](/ja/types/reference/ObjectKeys)を参照)。
 
 #### 戻り値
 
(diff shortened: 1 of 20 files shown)


raon0211:
Hi @ProgWon, thank you for the contribution — the diagnosis in #2036 was spot on, and the PR came complete with runtime tests, type-level tests, and docs in all four languages. Much appreciated!

I've pushed two commits on top of your branch (86f8937e, 11464ffa) that keep your fix intact but reshape the supporting type. Since I changed someone else's work, I owe you an explanation:

- **`StringKeyOf` (internal) → `ObjectKeys` (public, `es-toolkit/types`).** The type you introduced is useful well beyond `omitBy`/`pickBy` — typing the result of `Object.keys(obj)` (which TypeScript widens to `string[]`) is a common enough need that libraries like ts-extras exist largely for it. Promoting it to `es-toolkit/types` makes it the key-side counterpart to our existing `ValueOf`, and it also fixes a docs problem: the parameter docs referenced `StringKeyOf<T>`, an unexported internal name readers couldn't look up. Now they link to a documented public type.

- **Simpler definition.** The conditional type is now a one-liner, `` `${Exclude<keyof T, symbol>}` ``, which mirrors the runtime semantics of `Object.keys` exactly: numeric keys stringified, symbol keys excluded. This also matches the ecosystem — ts-extras' `ObjectKeys` has the identical name and definition, and type-fest's `KeyAsString` is the same construction.

- **One behavior change: `Record<number, V>` keys are now `` `${number}` `` instead of `string`.** This is more precise — comparing a key of a number-keyed object against a non-numeric string like `key === 'abc'` is now a compile error instead of silently-always-false. There's no ergonomic cost: `` `${number}` `` is assignable to `string`, and callbacks declared as `(value, key: string) => ...` still type-check thanks to contravariance. I updated the corresponding spec expectation to match.

- **Docs.** Added an `ObjectKeys` reference page in all four languages and linked it from the `omitBy`/`pickBy` parameter docs.

Thanks again for the careful work here!

raon0211 (approved):

PR #2048, opened 2026-08-23
fix(omitBy, pickBy): preserve record return types

### LINKED ISSUE #1987: `omitBy` / `pickBy` types

GiHoon1123:
## Summary

Fixes #1987.

`omitBy` and `pickBy` currently wrap every result in `Partial<T>`. For string and number index signatures, that unnecessarily adds `undefined` to the value type and makes the result incompatible with the original record type.

## Changes

- Preserve string and number index signatures in the return type
- Keep `Partial<T>` for objects with a fixed set of keys
- Avoid preserving required properties that can be removed from types that also have an index signature
- Add type coverage for string records, number records, fixed-key objects, and mixed index-signature types

## Test results

- `yarn vitest run src/object/omitBy.spec.ts src/object/pickBy.spec.ts`: 20 tests passed
- `yarn vitest run`: passed
- `yarn typecheck`: passed
- ESLint on changed files: 0 errors

### DIFF
--- a/src/object/omitBy.spec.ts
+++ b/src/object/omitBy.spec.ts
@@ -54,16 +54,35 @@ describe('omitBy', () => {
   it('should type number index signature keys as numeric strings', () => {
     const obj: Record<number, string> = { 1: 'keep', 2: 'omit' };
 
-    omitBy(obj, (_value, key) => {
+    const result = omitBy(obj, (_value, key) => {
       expectTypeOf(key).toEqualTypeOf<`${number}`>();
       return key === '2';
     });
+
+    expectTypeOf(result).toEqualTypeOf<Record<number, string>>();
+  });
+
+  it('should preserve string index signatures', () => {
+    const obj: Record<string, number> = { a: 1, b: 2 };
+    const result = omitBy(obj, value => value === 2);
+
+    expectTypeOf(result).toEqualTypeOf<Record<string, number>>();
+  });
+
+  it('should not preserve required properties alongside an index signature', () => {
+    const obj: Record<string, number> & { required: number } = { required: 1, other: 2 };
+    const result = omitBy(obj, (_value, key) => key === 'required');
+
+    expectTypeOf(result).toEqualTypeOf<Record<string, number>>();
+    expect(result).toEqual({ other: 2 });
   });
 
   it('should preserve string literal key types', () => {
-    omitBy({ a: 1, b: 2 }, (_value, key) => {
+    const result = omitBy({ a: 1, b: 2 }, (_value, key) => {
       expectTypeOf(key).toEqualTypeOf<'a' | 'b'>();
       return false;
     });
+
+    expectTypeOf(result).toEqualTypeOf<Partial<{ a: number; b: number }>>();
   });
 });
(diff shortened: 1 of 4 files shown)


dayongkr:
One thing on the framing: `Record<string, V>` does have undefined at runtime, we just don't see it because `noUncheckedIndexedAccess` is off by default.

Direction is right though. Dropping Partial on the index signature and only making declared keys optional is what I'd want.

Branching on the shape misses a few cases:

```ts
Record<string, number> & { required: number } // required is gone
{ [k: string]: string | number; a: string }   // a: string is gone
{ [k: symbol]: number }                       // still Partial, same bug as the issue
union input                                   // falls back to Partial<T>
```

Splitting T instead of branching on it gets all four:

```ts
type PickIndexSignature<T> = { [K in keyof T as {} extends Record<K, unknown> ? K : never]: T[K] };
type OmitIndexSignature<T> = { [K in keyof T as {} extends Record<K, unknown> ? never : K]: T[K] };
type Simplify<T> = { [K in keyof T]: T[K] } & {}; // display only

type Filtered<T> = Simplify<PickIndexSignature<T> & Partial<OmitIndexSignature<T>>>;
```

`{} extends Record<K, unknown>` is the test for an index signature key: an empty object satisfies `Record<string, unknown>` but not `Record<'a', unknown>`.

So it's the rule with no cases in it. An index signature key was already allowed to be missing, leave it alone. A declared key wasn't, so mark it optional. Nothing to install, the three lines above are the whole thing.

dayongkr:
#2019 fixes the same issue and was opened first, so let's go with that one.
We don't take duplicate PRs. Thanks for the work.

PR #2052, opened 2026-08-24
fix[attempt, attemptAsync]: default error type parameter to unknown

### LINKED ISSUE #2050: Some `attemptAsync` documentation examples do not type-check
### es-toolkit version

1.51.0

### Entry point

es-toolkit

### Reproduction

```typescript
// You can also specify types
interface User {
  id: number;
  name: string;
}

const [error, users] = await attemptAsync<User[]>(async () => {
  const response = await fetch('https://api.example.com/users');
  return response.json();
});
// users is inferred as User[] type

```

### Current:

Image

The reproduction above produces two type errors.
The call that explicitly supplies only `T` produces:
```
TS2558: Expected 2 type arguments, but got 1.
```

These patterns are currently shown on the attemptAsync reference page: https://es-toolkit.dev/reference/util/attemptAsync.html

### expected behavior:
The documented TypeScript examples should type-check against the declarations shipped with the package.

I am not sure whether the intended resolution is to update the documentation examples or change the public type declaration. This report is about the mismatch between the documented usage and the published types.

What is right and what is wrong?

### Validations

raon0211:
## Summary

The documented usage `attempt<string[]>(...)` and `attemptAsync<User[]>(...)` fails to type-check with `TS2558: Expected 2 type arguments, but got 1`, because both functions declare two type parameters (`T`, `E`) and TypeScript does not support partial type argument inference.

fixes #2050

## Changes

- Add a default to the error type parameter (`E = unknown`) in `attempt` and `attemptAsync`, so supplying only the result type argument — as shown in the reference docs — compiles.
- Add tests exercising the single-type-argument usage so a regression fails the build.

`unknown` matches what `E` was already inferred as when no type arguments were given, so existing call sites are unaffected.

### DIFF
--- a/src/util/attempt.spec.ts
+++ b/src/util/attempt.spec.ts
@@ -15,6 +15,12 @@ describe('attempt', () => {
     ).toEqual([new Error('test'), null]);
   });
 
+  it('should accept a single type argument for the result type', () => {
+    const [error, names] = attempt<string[]>(() => ['Alice', 'Bob']);
+    expect(error).toBeNull();
+    expect(names).toEqual(['Alice', 'Bob']);
+  });
+
   it('should return the result of the promise', async () => {
     const [error, result] = attempt(async () => 1);
     expect(error).toBeNull();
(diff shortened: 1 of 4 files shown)


dayongkr:
LGTM!

PR #2054, opened 2026-08-24
fix(plugin): make skills portable when installed individually

### LINKED ISSUE #1932: Agent skills rely on repository-level symlinks and break when installed via `gh skill`
The agent skills under `es-toolkit-plugin/skills` are not fully portable when installed with tools that distribute each skill as a standalone directory. (Using official Github CLI)

Each skill contains a `docs` symbolic link pointing outside the skill directory:

```text
es-toolkit-plugin/skills/guide/docs     -> ../../../docs
es-toolkit-plugin/skills/migrate/docs   -> ../../../docs
es-toolkit-plugin/skills/recommend/docs -> ../../../docs
```

When an installer copies only the selected skill directory, the repository-level `docs` directory is not included. The resulting installed skill therefore contains a dangling symlink or a symlink whose target does not exist.

### Reproduction

Using GitHub CLI 2.96.0:

```bash
gh skill install toss/es-toolkit
```

After installation, inspect one of the installed skills:

```bash
ls -la <installed-skill-directory>
```

The `docs` entry is still a symbolic link to `../../../docs`.

However, the target directory was not installed, so references such as `docs/reference/**/*.md` cannot be accessed by the skill.

zaewc:
## Summary

Fixes #1932.

The `guide`, `recommend`, and `migrate` skills each include a `docs -> ../../../docs` symlink. Installers such as `gh skill` copy one skill directory at a time, which leaves that link dangling.

This takes a narrower approach than #1971. It keeps the existing three-skill setup and removes the repository-level dependency. The skills now inspect the version of es-toolkit installed in the user's project, with a source-tree fallback when they are run from this repository.

## Changes

- Remove the repository-level `docs` symlinks from all three skill
- Resolve package metadata, entry points, and declarations from the installed
  es-toolkit package
- Keep the setup guidance self-contained when es-toolkit is not installed yet
- Add a CI check that rejects symlinks inside standalone skill directories
- Update the plugin readme


.scripts/check-plugin-skills.mjs
```diff
@@ -0,0 +1,50 @@
+import { lstat, readdir } from 'node:fs/promises';
```
dayongkr: Thanks for picking this up, that symlink has been broken for a while.

I'd drop the CI check though. A symlink sneaking back in is cheap to notice and cheap to fix. Skills change fast as models improve, and this directory may not exist in a few months, so guarding it in CI feels heavier than the problem.

dayongkr:
Did you test this with a smaller model like Haiku?

Spelling the steps out this explicitly, down to the lookup command, is mostly what a weaker model needs. A stronger one finds it on its own. I write these skills for the weaker end, so I'm curious what you saw.

.scripts/check-plugin-skills.mjs
zaewc: [5e80fa1](https://github.com/toss/es-toolkit/pull/2054/commits/5e80fa1ef8ea03ac2c7508723ebf3943437becb6) deleted!

zaewc:
@dayongkr 

I hadn't tested with Haiku before opening the PR, but I just ran all three skills with Haiku 4.5.

I tested from a temporary npm project with es-toolkit 1.51.0 installed, not from the source repo

- `guide` found the installed package and read its version/exports.
- `recommend` found `chunk` and checked its declaration before recommending it.
- `migrate` confirmed `get` is only available from `es-toolkit/compat` and checked the declaration before suggesting the migration.

The detailed commands seem useful for Haiku, it followed and adapted them instead of guessing. `migrate` did try to read `dist` as a file once, but recovered and found the right declaration

dayongkr:
One more note, since the symlink fix looks like it brings in a different bug.

Your Haiku run used `chunk` and `get`, both inside the four entry points in the snippet. I tried a Map question instead and it answered that es-toolkit has no function for Maps, suggesting native `Array.from(map.values()).find()`. `findValue` in `es-toolkit/map` is exactly that, but the module isn't in the list and the root doesn't re-export it.

Keeping the change minimal is right and this doesn't have to grow it. Deriving the entries from `Object.keys(pkg.exports)` is about the same size as hardcoding them. Worth testing across a few subpaths too, not just the root ones.

zaewc:
@dayongkr 

Good catch, I replaced the hardcoded list with entries from `pkg.exports`. thanks!!

I also reran the Map question with Haiku.  It found `findValue` from `es-toolkit/map` and checked the installed declaration and I tried a few other subpaths like `server` and `compat` too

dayongkr (approved):
Thanks!!

PR #2056, opened 2026-08-25
fix(add,subtract,multiply,divide): match Lodash on null operands

jaideeppyne:
## The difference (as code)

`es-toolkit/compat`'s math operations disagree with Lodash when an operand is `null`:

```js
add(null, '1')        // lodash: 'null1'  es-toolkit: '1'
add('1', null)        // lodash: '1null'  es-toolkit: '1'
add(null, undefined)  // lodash: null     es-toolkit: undefined
subtract(null, '1')   // lodash: NaN      es-toolkit: -1
multiply(null, '1')   // lodash: NaN      es-toolkit: 0
divide('1', null)     // lodash: NaN      es-toolkit: Infinity
```

## Cause

Two issues shared by all four functions:

1. `value ?? other` treated a present `null` as a missing argument; only `undefined` should be treated as missing (so `add(null, undefined)` should return `null`, not `undefined`).
2. The string branch used the public `toString` (`null` → `''`) instead of Lodash's `baseToString` (`null` → `'null'`), which is why the numeric ops silently coerced `null` to `''` and produced wrong numbers instead of `NaN`.

## Changes

- Select the defined operand with an explicit `undefined` check instead of `??`.
- Coerce a `null` operand to `'null'` in the string branch, matching Lodash's `baseToString`.
- Added regression tests covering `null` operands for all four functions (they fail on `main`).

### DIFF
--- a/benchmarks/bundle-size/add.spec.ts
+++ b/benchmarks/bundle-size/add.spec.ts
@@ -9,6 +9,6 @@ describe('add bundle size', () => {
 
   it('es-toolkit/compat', async () => {
     const bundleSize = await getBundleSize('es-toolkit/compat', 'add');
-    expect(bundleSize).toMatchInlineSnapshot(`495`);
+    expect(bundleSize).toMatchInlineSnapshot(`537`);
   });
 });
--- a/src/compat/math/add.spec.ts
+++ b/src/compat/math/add.spec.ts
@@ -58,6 +58,26 @@ describe('add', () => {
     expect(add(undefined, 4)).toBe(4);
   });
 
+  it(`\`add\` should treat \`null\` operands like Lodash`, () => {
+    // A `null` operand next to a string is stringified as `'null'`, not `''`.
+    // @ts-expect-error - invalid arguments
+    expect(add(null, '1')).toBe('null1');
+    // @ts-expect-error - invalid arguments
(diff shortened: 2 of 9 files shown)


dayongkr:
Verified against lodash 4.18.1; all nine expectations match.

The stringify branch can lean on what is already in the repo. `baseToString` in `src/compat/util/toString.ts` is lodash's `baseToString`, and it already turns `null` into `'null'`. Exporting it collapses the branch and removes the duplicated comment from all four files:

```diff
-import { toString } from '../util/toString.ts';
+import { baseToString } from '../util/toString.ts';

   if (typeof value === 'string' || typeof other === 'string') {
-    // Lodash concatenates using `baseToString`, which stringifies `null` as
-    // `'null'` (unlike `toString`, which returns `''` for nullish values).
-    value = (value === null ? 'null' : toString(value)) as any;
-    other = (other === null ? 'null' : toString(other)) as any;
+    value = baseToString(value) as any;
+    other = baseToString(other) as any;
   }
```

jaideeppyne:
Thanks @dayongkr — good call. Addressed in e38d4e6: `baseToString` is now exported from `src/compat/util/toString.ts`, and `add` / `subtract` / `multiply` / `divide` all use it directly for the string branch instead of the duplicated `null ? 'null' : toString(...)` coercion.

PR #2064, opened 2026-08-28
fix[ceil, floor, round]: coerce arguments like lodash

Archibald1948:
## Summary

`ceil`, `floor` and `round` share `decimalAdjust`, which coerces its arguments differently from Lodash. The precision is read with `Number.parseInt(precision, 10)` where Lodash uses `toInteger(precision)` capped at `292`, and the number is read with `Number(number)` where Lodash uses `toNumber(number)`.

```js
import lodash from 'lodash';

lodash.ceil(4.016, true); // 4.1
compat.ceil(4.016, true); // 5

lodash.ceil(4.016, Infinity); // 4.016
compat.ceil(4.016, Infinity); // 5

lodash.ceil(4.016, [1, 2, 3]); // 5
compat.ceil(4.016, [1, 2, 3]); // 4.1

lodash.ceil(4.016, '1e3'); // 4.016
compat.ceil(4.016, '1e3'); // 4.1

lodash.ceil(Symbol('a')); // NaN
compat.ceil(Symbol('a')); // throws TypeError
```

`parseInt` reads only the leading digits of its input, so `'1e3'` became `1`, `'2px'` became `2` and `[1, 2, 3]` became `1`. In the other direction `true` and `Infinity` parsed to `NaN`, which is falsy, so the precision was dropped entirely and the value was rounded to an integer.

The same mismatch applies to `floor` and `round`, since all three go through `decimalAdjust`.

## Changes

`decimalAdjust`: coerce both arguments the way Lodash's `createRound` does.

```js
number = toNumber(number);
// ...
precision = precision == null ? 0 : Math.min(toInteger(precision), 292);
```

Tests: add coercion cases to `ceil.spec.ts`, `floor.spec.ts` and `round.spec.ts`, covering the values `toInteger` resolves to `0`, the ones it resolves to a usable precision, the `292` cap, and `toNumber`'s handling of symbols.

```js
it('`ceil` should coerce `precision` with `toInteger` like lodash', () => {
  // Values that `toInteger` resolves to `0`, so no decimal adjustment happens.
  expect(ceil(4.016, [1, 2, 3])).toBe(5);
  expect(ceil(4.016, '2px')).toBe(5);

  expect(ceil(4.016, true)).toBe(4.1);
  expect(ceil(4.016, '0x2')).toBe(4.02);

  // Capped at `292`.
  expect(ceil(4.016, '1e3')).toBe(4.016);
  expect(ceil(4.016, Infinity)).toBe(4.016);
});

it('`ceil` should coerce `number` with `toNumber` like lodash', () => {
  expect(ceil(Symbol('a'))).toBe(NaN);
  expect(ceil(4.016, Symbol('a'))).toBe(5);
});
```

## Verification

I compared every combination of 25 numbers and 24 precisions across the three functions against Lodash at runtime. 176 of the 1800 results differed before this change and none differ after. A further 84 hand-picked edge cases — symbols, the sign of `0`, exponential notation, `valueOf`, boxed numbers — went from 12 mismatches to 0.

The existing tests continue to pass unchanged, including the ones covering the sign of `0`, infinite values, and precisions beyond the cap such as `ceil(1.797, 295)`.

## Benchmark results

`decimalAdjust` runs on every call, so I measured it. Medians of five runs of the existing benchmarks, all within noise:

| --- | --- | --- | --- |
| `ceil`, 12 calls | 541,231 hz | 549,898 hz | +1.6% |
| `floor`, 12 calls | 543,934 hz | 548,550 hz | +0.8% |
| `round`, 1 call | 3,936,717 hz | 3,950,223 hz | +0.3% |
| `round`, 9 calls | 665,385 hz | 672,704 hz | +1.1% |

### DIFF
--- a/src/compat/_internal/decimalAdjust.ts
+++ b/src/compat/_internal/decimalAdjust.ts
@@ -12,7 +12,7 @@ export function decimalAdjust(
   }
   // lodash coerces the precision with `toInteger` and caps it at 292, rather
   // than parsing it as a decimal integer literal.
-  precision = precision == null ? 0 : Math.min(toInteger(precision), 292);
+  precision = Math.min(toInteger(precision), 292);
   if (precision && Number.isFinite(Number(number))) {
     const [magnitude, exponent = 0] = number.toString().split('e');
     let adjustedValue: string | number = Math[type](Number(`${magnitude}e${Number(exponent) + precision}`));
--- a/src/compat/_internal/decimalAdjust.ts
+++ b/src/compat/_internal/decimalAdjust.ts
@@ -1,13 +1,18 @@
+import { toInteger } from '../util/toInteger.ts';
+import { toNumber } from '../util/toNumber.ts';
+
 export function decimalAdjust(
   type: 'round' | 'floor' | 'ceil',
   number: number | string,
   precision: number | string = 0
 ): number {
-  number = Number(number);
+  number = toNumber(number);
   if (Object.is(number, -0)) {
     number = '-0';
   }
-  precision = Math.min(Number.parseInt(precision as string, 10), 292);
+  // lodash coerces the precision with `toInteger` and caps it at 292, rather
+  // than parsing it as a decimal integer literal.
+  precision = precision == null ? 0 : Math.min(toInteger(precision), 292);
   if (precision && Number.isFinite(Number(number))) {
     const [magnitude, exponent = 0] = number.toString().split('e');
     let adjustedValue: string | number = Math[type](Number(`${magnitude}e${Number(exponent) + precision}`));
--- a/src/compat/math/ceil.spec.ts
+++ b/src/compat/math/ceil.spec.ts
@@ -85,4 +85,36 @@ describe('ceil', () => {
     expect(ceil(5e-324, 323)).toBe(1e-292);
     expect(ceil(5e-324, -323)).toBe(0);
   });
+
+  it(`\`ceil\` should coerce \`precision\` with \`toInteger\` like lodash`, () => {
+    // Values that `toInteger` resolves to `0`, so no decimal adjustment happens.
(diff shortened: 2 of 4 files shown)


src/compat/_internal/decimalAdjust.ts
```diff
@@ -1,13 +1,18 @@
...
 ): number {
-  number = Number(number);
+  number = toNumber(number);
   if (Object.is(number, -0)) {
     number = '-0';
   }
-  precision = Math.min(Number.parseInt(precision as string, 10), 292);
+  // lodash coerces the precision with `toInteger` and caps it at 292, rather
+  // than parsing it as a decimal integer literal.
+  precision = precision == null ? 0 : Math.min(toInteger(precision), 292);
```
dayongkr: This branch looks unnecessary to me:

```ts
precision = precision == null ? 0 : Math.min(toInteger(precision), 292);
```

`toInteger` already turns `null` and `undefined` into 0, and the signature defaults `precision` to 0. Any reason for keeping it?

Archibald1948: You're right, thanks — I mirrored the shape of Lodash's `createRound`, which writes it as `precision == null ? 0 : nativeMin(toInteger(precision), 292)`, but that guard buys nothing here.

`toInteger` goes through `toFinite`, which starts with `if (!value) { return value === 0 ? value : 0; }`, so both `null` and `undefined` already come out as `0`:

```js
toInteger(null); // 0
toInteger(undefined); // 0
```

I dropped the branch and re-ran the comparison against Lodash over 25 numbers × 29 precisions × 3 functions — including `null`, `undefined`, `-0`, `''` and symbols as the precision — and all 2175 results still match.

Archibald1948: I modified the code and pushed it again.
Please check it for me!

raon0211 (approved):

Issue #1986, opened 2026-08-02
es-toolkit/types: what to add next


dayongkr:
## Context

`es-toolkit/types` shipped in v1.50.0 with six utilities (#1818). This issue tracks what to add next.

I went through a number of well-known OSS projects to see which utility types they keep declaring on their own. The list below is ordered by how often they showed up and how hard they are to get right by hand.

## Priority 1

**`UnionToIntersection<T>`** — `keyof` on a union only keeps the common keys.

```ts
type Keys = keyof ({ a: 1 } | { b: 2 });                     // never
type Keys = keyof UnionToIntersection<{ a: 1 } | { b: 2 }>;  // 'a' | 'b'
```

Also folds a variadic merge result back into one object. Several projects link an external blog post in a comment to explain the contravariant `infer` trick.

**`IsAny<T>`** — `any` satisfies both branches of a conditional, so it has to be caught first.

```ts
type Serialized<T> = T extends Date ? string : T;
type Serialized<any>;  // matches every branch at once, output becomes meaningless

type Safe<T> = IsAny<T> extends true ? any : Serialized<T>;
```

Everything else in this family (`IsUnion`, `IsUnknown`, `IsTuple`) builds on this one.

**`DistributiveOmit<T, K>`** — built-in `Omit` flattens a union into its common keys.

```ts
type Shape = { kind: 'circle'; r: number } | { kind: 'rect'; w: number };

Omit<Shape, 'kind'>;              // {} — r and w are both gone
DistributiveOmit<Shape, 'kind'>;  // { r: number } | { w: number }
```

Every project that has it wrote the same name and the same body.

## Priority 2

**`MaybePromise<T>`** — lets a callback be sync or async.

```ts
function onEvent(handler: () => MaybePromise<void>) {}

onEvent(() => {});        // ok
onEvent(async () => {});  // ok
```

Open question: `Promise` or `PromiseLike`. Note our own `retry`/`withTimeout`/`attemptAsync` currently reject sync callbacks.

**`Primitive`** — recursion base case, and the narrowing type of a runtime guard.

```ts
type Paths<T> = T extends Primitive ? never : /* recurse */;
```

`isPrimitive` already ships this exact union anonymously in its signature, so naming it is the whole change.

**`JsonValue`** (with `JsonObject` / `JsonArray` / `JsonPrimitive`)

```ts
function send(body: JsonValue) {}

send({ a: [1, 'x', null] });  // ok
send({ fn: () => {} });       // error
```

[...]

PR #2074, opened 2026-08-31
fix[DeepPartial, DeepReadonly]: preserve branded primitive types

Tnalxmsk:
## Summary

This PR follows up on the brand-preservation regression item listed under “Worth doing regardless” in [#1986](https://github.com/toss/es-toolkit/issues/1986).

`DeepPartial` and `DeepReadonly` recursively transform object property values. When a plain object contains a branded primitive field, that field is processed through the object branch and is no longer equivalent to the original branded type.

```ts
type UserId = string & { readonly __brand: 'UserId' };

type User = {
  id: UserId;
};

type PartialUser = DeepPartial<User>;
// Expected: { id?: UserId }

type ReadonlyUser = DeepReadonly<User>;
// Expected: { readonly id: UserId }
```

The shared `Primitive` type introduced in [[#2044](https://github.com/toss/es-toolkit/pull/2044)](https://github.com/toss/es-toolkit/pull/2044) provides a common terminal condition for these recursive utilities. Primitive types, including branded primitive types, can therefore pass through without being recursively mapped as objects.

## Changes

- Use the existing `Primitive` type as a terminal condition in `DeepPartial` and `DeepReadonly`.
- Add type-level regression tests covering both direct and nested branded primitive inputs.
- Replace the explicit `any[]` function constraints with `never[]`, preserving parameterized function types without lint warnings.
- Update the related reference documentation in English, Korean, Japanese, and Simplified Chinese.

## Implementation notes

This is a type-only change and does not affect runtime behavior. Reusing the existing `Primitive` type avoids duplicating the primitive union across both utilities.

The `never[]` function constraint still matches arbitrary function signatures in the conditional type while avoiding an explicit `any`.


src/types/Simplify.ts
```diff
@@ -12,4 +15,4 @@
  * type User = Simplify<A & B>;
  * // hover => { name: string; age: number }   (instead of A & B)
  */
-export type Simplify<T> = { [K in keyof T]: T[K] } & {};
+export type Simplify<T> = T extends Primitive ? T : { [K in keyof T]: T[K] } & {};
```
dayongkr: `DeepPartial` and `DeepReadonly` look good. They recurse into every value, so a branded field breaks even when the input is a plain object.

Could you drop the `Simplify` change? It expands one level and leaves field types alone, so a primitive only reaches it when passed at the top level, which is outside what the docs describe.

Tnalxmsk: Thanks for clarifying the intended scope. That makes sense — `Simplify` only expands the top-level shape and already preserves branded types in object properties.

I've removed the `Simplify` implementation, tests, and documentation changes, and narrowed the PR to `DeepPartial` and `DeepReadonly`. I've also updated the PR title and description accordingly.

dayongkr (approved):

PR #2097, opened 2026-09-11
feat(compat): support Node.js 6+ in es-toolkit/compat

raon0211:
## Summary

`es-toolkit/compat` exists to migrate code away from lodash, and that code often still runs in old environments. This makes `es-toolkit/compat` (and every `es-toolkit/compat/*` entrypoint) work on **Node.js 6+ as published**, with no transpilation and no polyfills, and adds the guardrails that keep it that way. The main library keeps its current floor (Node.js 20.12+ / Chrome 98 / Safari 15.4).

Measured on the current `main` with a real Node.js 6.17.1: 132 of 297 `compat/*` entry points fail to even load (optional chaining, `??`, object spread), and after syntax lowering 68 functions still fail at runtime (`Object.hasOwn`, `Object.values`, `Array#flat`, `findLast`, `matchAll`, `\p{…}` regexes, `structuredClone`, …). With this change all 298 entry points load and all 844 generated compat `@example` cases pass on Node.js 6.17.1.

## Changes

- **compat source** (`src/compat/**`): restricted to ES2015 syntax and ES2015 built-ins. Rewrote `?.`, `??`, object spread, optional `catch`, `Object.hasOwn`, `Object.values`, `Array#flat`/`flatMap`/`at`/`findLast`/`findLastIndex`, `String#matchAll`/`trimStart`/`trimEnd`, numeric separators. Behavior is unchanged; the existing specs are the oracle.
- **`words` and the case functions** (`camelCase`, `kebabCase`, `snakeCase`, `startCase`, `lowerCase`, `upperCase`): the Unicode-property regex is now compiled inside `try/catch`, and engines that cannot parse `\p{…}` fall back to lodash's own ES5 word patterns (`src/compat/_internal/legacyWords.ts`), so those engines get exactly lodash's results. Modern engines keep using the existing pattern. A parity spec pins the two implementations to each other.
- **`truncate`**: replaced the named capture group (`(?<result>…)`, Node.js 10+) with a numbered group.
- **`isNative`**: uses the internal `globalThis` shim instead of the bare global (Node.js 12+).
- **`cloneDeepWith`**: compat now has its own implementation instead of delegating to the main library's, which relies on `structuredClone`. This is the only place compat was decoupled from a main implementation; `matches` and `conforms` now use compat's `cloneDeep` too. Two small differences, both toward lodash: objects without a `constructor` (e.g. `Object.create(null)`) become plain objects at every nesting level rather than only at the top level, and `arguments` objects keep their `length` at every level. Error objects are still cloned (previously via `structuredClone`, now by copying the own properties onto an object with the same prototype).
- **Main-library files compat imports** (`groupBy`, `pull`, `remove`, `isEqualWith`, `clone`, `deburr`, `trimStart`, `trimEnd`, `debounce`): one-line changes to drop `Object.hasOwn`, `??`, `?.`, `trimStart`/`trimEnd`, and an unguarded `AggregateError`. `invokeMap` no longer imports the whole `predicate` barrel, which was pulling ~30 unrelated files into compat's dependency graph.
- **ESLint**: a new block applies `eslint-plugin-es-x` (everything newer than ES2015, except `Array#includes`, which Node.js 6 ships) and `eslint-plugin-compat` with a `node 6` target to `src/compat/**`. It is deliberately scoped to compat; the main-library files compat imports are covered by the CI lane below instead.
- **CI**: new `compat-node6` job in `ci.yml` builds the package, then runs it on a real Node.js 6.17.1 (checksum-pinned download): every `compat/*.js` entry point must load, and every generated compat `@example` case must pass. It runs on every PR, since a change to a shared main-library file can break compat on Node.js 6 without touching `src/compat`.
- **browser-compat suite**: the `cases-legacy` variant no longer excludes the word-splitting functions (they now work on ES2015 engines thanks to the fallback), and a `node6` fixture transpiles only the cases, never the dist.
- **Docs**: new "Node.js support" section in `browser-support.md` (en/ko/ja/zh_hans); `CLAUDE.md`/`AGENTS.md` note the compat floor.

`Object.hasOwn` → `hasOwnProperty.call` in the shared main files is not a regression: in a quick microbenchmark on Node.js 24 the `call` form was ~20% faster.


raon0211:
Rebased onto `main` (resolving the `camelCase` bundle-size snapshot conflict from #2102/#2103) and re-measured the snapshots.

Also addressed the two CodeQL findings on `trimEnd`: `str.replace(/\s+$/, '')` is quadratic on a long run of whitespace followed by a non-whitespace character. `trimStart`/`trimEnd` now scan from the edges with a per-character `\s` test (linear, still ES2015), and the compat variants delegate to the main implementations. Added a test covering the full whitespace set and a regression test for the long-whitespace case.

PR #2101, opened 2026-09-15
fix[forEach, forEachRight, forIn, forInRight, forOwn, forOwnRight, reduce, reduceRight]: resolve the iteratee like lodash

Archibald1948:
## The difference (as code)

The iteration functions below call their `iteratee` directly, so anything that is not a function throws, where lodash resolves it the way every other collection method does:

```js
forEach([1, 2, 3], null);                 // lodash: [1, 2, 3]        es-toolkit: TypeError: callback is not a function
forOwn({ a: { x: 1 } }, 'x');             // lodash: { a: { x: 1 } }  es-toolkit: TypeError: iteratee is not a function
reduce([1, 2, 3], null);                  // lodash: 1                es-toolkit: TypeError: iteratee is not a function
reduce([{ a: 1 }, { a: 1 }], { a: 1 });   // lodash: true             es-toolkit: TypeError: iteratee is not a function
```

Affected: `forEach`, `forEachRight`, `forIn`, `forInRight`, `forOwn`, `forOwnRight`, `reduce`, `reduceRight` (and the `each` / `eachRight` aliases).

Shorthands are not just tolerated by lodash, they are honoured. In the `forEach` family a property shorthand that returns `false` stops the iteration, exactly like a callback returning `false`:

```js
let visited = 0;
const array = [{ a: false }, { get a() { visited++; return true; } }];

forEach(array, 'a');
// lodash: stops at the first element, visited === 0
// es-toolkit: TypeError
```

## Cause

In lodash all eight go through `getIteratee`, i.e. `baseIteratee`:

```js
function forEach(collection, iteratee) {
  var func = isArray(collection) ? arrayEach : baseEach;
  return func(collection, getIteratee(iteratee, 3));
}

function forOwn(object, iteratee) {
  return object && baseForOwn(object, getIteratee(iteratee, 3));
}

function reduce(collection, iteratee, accumulator) {
  // ...
  return func(collection, getIteratee(iteratee, 4), accumulator, initAccum, baseEach);
}
```

In compat they only default the parameter to `identity` and then call it, so `null` (not covered by a default parameter) and every shorthand reach a call on a non-function:

```js
export function forOwn(object, iteratee = identity) {
  // ...
  if (iteratee(iterable[key], key, iterable) === false) {
```

`es-toolkit/compat` already has the `baseIteratee` equivalent, `iteratee()`, and `filter` / `partition` already resolve their predicate with it:

```js
predicate = iteratee(predicate);
```

## Changes

Resolve the iteratee once per call, right after the early return, the same way `filter` does:

```js
// forEach, forEachRight (the parameter is named `callback`)
callback = iteratee(callback);

// forIn, forInRight, forOwn, forOwnRight, reduce, reduceRight
// (the parameter is named `iteratee`, so the helper is imported as `createIteratee`, like `groupBy` does)
iteratee = createIteratee(iteratee);
```

`iteratee()` returns a function argument unchanged and maps `null` / `undefined` to `identity`, so existing callers see no difference. Each file gains one import and one line.

## Verification

Focused grid against `lodash@4.18.1`: 10 functions x 6 collections x 8 iteratee values (omitted, `undefined`, `null`, `0`, `'a'`, `{ a: 1 }`, `['a', 1]`, `true`), 480 calls:

| | mismatches |
| --- | --- |
| before | 240 / 480 (every one "lodash returns, es-toolkit throws") |
| after | **0 / 480** |

Broad sweep, every 1- and 2-argument combination of a 35-value pool (1,260 calls per function):

| | `forEach` family (8 incl. aliases) | `reduce` / `reduceRight` |
| --- | --- | --- |
| before | 352 / 1,260 each | 224 / 1,260 each |
| after | **0 / 1,260 each** | 21 / 1,260 each |

No call that matched before mismatches now. The 21 left for `reduce` / `reduceRight` are all `[]`, `Map` or `Set` used as the iteratee. Those used to throw; now they return what compat's shared `iteratee()` produces for those values, which is the same known difference `filter`, `find` and the other shorthand-taking functions already have against lodash — out of scope here. The `forEach` family shows no such residue because it only looks at whether the result is `false`.

Regression tests were added to all eight spec files (14 tests). On `main` they fail with:

```
TypeError: iteratee is not a function
TypeError: callback is not a function
```

Full suite: 714 files, 5053 passed / 2 skipped. `tsc --noEmit` clean, `yarn lint` 0 errors (no new warnings in the touched files), type-tests pass.

## Benchmarks

The change adds one `iteratee()` call per invocation (a `null` check and a `typeof` switch), independent of collection size.

Interleaved A/B (`fix`, then `main`, seven alternating rounds) on the three hottest benchmarks. Before every run the Vite dependency cache was cleared, and after every run I checked the pre-bundled `es-toolkit/compat` for which implementation had actually been measured (14/14 matched the intended side). Each compat row is divided by the lodash row of the same group in the same round to cancel machine drift; the table shows the median of the seven normalized deltas.

| benchmark | normalized delta |
| --- | ---: |
| `forEach` | -0.8% |
| `forEach` — largeArray | -3.8% |
| `forEach` — object | 0.0% |
| `forEach` — string | -2.4% |
| `forOwn` | -0.8% |
| `reduce` | -4.7% |
| `reduce` — large array | -1.1% |

I want to be upfront that these runs were noisy: the machine was under other load at the time (load average around 5 on 8 cores), and individual rounds swung by up to +/-50% even on the lodash controls. The large-array rows are the clearest indicator of the noise floor — a constant per-call cost cannot move a benchmark that spends ~75µs per call by 3.8% — so I read this as no consistent effect within roughly +/-5%, not as a measured slowdown. `filter` and `partition` already resolve their predicate the same way. Happy to re-run on a quiet machine if you'd like tighter numbers.


raon0211 (approved):

Archibald1948:
merge thanks!!

PR #2112, opened 2026-09-18
fix(clone): support Boolean, Number, and String wrapper objects

tushardhawas:
## Summary

`clone` produced a broken result for `Boolean`, `Number`, and `String` **wrapper objects**. The clone had the correct prototype but no internal `[[PrimitiveValue]]` slot, so any method that relies on it threw at runtime:

```ts
import { clone } from 'es-toolkit';

clone(new Number(42)).valueOf();
// ❌ TypeError: Number.prototype.valueOf requires that 'this' be a Number
```

`cloneDeep` already handles these correctly, so `clone` was inconsistent with it:

```ts
import { cloneDeep } from 'es-toolkit';
cloneDeep(new Number(42)).valueOf(); // ✅ 42
```

## Change

Add wrapper-object handling to the shallow `clone`, mirroring `cloneDeep`: reconstruct the wrapper from its primitive value and shallow-copy any extra own enumerable properties, skipping intrinsic non-writable ones (such as a `String` wrapper's indexed characters).

```ts
clone(new Number(42)).valueOf();  // 42
clone(new Boolean(true)).valueOf(); // true
clone(new String('es-toolkit')).valueOf(); // 'es-toolkit'
```

## Tests

Added specs for `Number`, `Boolean`, and `String` wrapper objects (value preserved, correct instance, not the same reference) plus a case covering extra own properties. All `src/object` tests pass (267/267); `tsc --noEmit` and ESLint are clean.

Related to #196.


tushardhawas:
Thanks for the review — addressed both points in 99b4da2f:

- **`valueOf` shadowing:** now read the wrapped primitive from the intrinsic slot (`Number.prototype.valueOf.call(obj)`, and the `Boolean`/`String` equivalents) instead of `obj.valueOf()`, so an own `valueOf` can't cause the wrong value to be cloned.
- **Enumerable symbols:** the extra-property copy now includes enumerable own symbol keys, matching the generic `Object.assign` shallow-clone path.

Added tests for both (a shadowed `valueOf` and an enumerable symbol on a wrapper), and refreshed the bundle-size snapshots.

